Issue #13, opened 2022-06-04
scatter size argument not used


GarrettBlair:
see https://github.com/kushalkolar/fastplotlib/blob/6c0c7ae4cb68929d90a71186617258204a231b44/fastplotlib/graphics.py#L88

kushalkolar:
fixed by #14

PR #14, opened 2022-06-04
Fixed issue #13: size argument not used

GarrettBlair:


kushalkolar:
thanks for finding this!

Issue #23, opened 2022-08-19
vmin vmax should be attributes of Image graphic


clewis7:
useful so that quickminmax doesn't have to be run externally and for other stuff

https://github.com/kushalkolar/fastplotlib/blob/37ad4a816848cb10268f9dda93734e725eabee48/fastplotlib/graphics/graphics.py#L72

PR #24, opened 2022-08-22
changing vmin/vmax to attributes

clewis7:
addresses issue #23 

made vmin/vmax attributes to the class along with getters/setters

now, when looking a z-shift, should be able to change vmin/vmax to clear up images


kushalkolar:
I didn't realize the `clim` property already exists which provides this functionality. Maybe we should rename `vmin` `vmax` to `clim`?

I think at some point we should decide whether `fastplotlib` uses setters and getters or properties. Although I like setters and getters more, `pygfx` uses properties.

Issue #22, opened 2022-06-26
simplification ideas


kushalkolar:
- [x] gridplot args, done in #45 
  - [x] allow passing in nested list of list for `camera` and `controller` kwargs and then make it a numpy array 
  - [x] `controllers` args
    - [x] `"sync"` - everything gets synced

- [x] #36

PR #27, opened 2022-09-23
adding functionality to gridplot to name subplots in gridplot

clewis7:
addressing some of the simplification ideas in #22 , #36 

basically adds the ability to name subplots in a gridplot


fastplotlib/layouts.py
```diff
@@ -7,6 +7,12 @@
 from wgpu.gui.auto import WgpuCanvas
 
 
+class GridPlotIndexer():
+    def __init__(self):
+        self.name: str = None
+        self.index: tuple = None
+
```
kushalkolar, line 14: I think `self.subplot` should be an attribute too, and this would contain the actual subplot.

We should then think about whether to keep the current API of:

```python
gp.subplots[i, j]  # returns subplot at (i, j)

# since this would basically be a duplicate of

gp[i, j].subplot  # returns subplot at (i, j)

but this GridPlotIndexer() interface provides more features like
gp[i, j].name  # returns a user-set name etc.
```

I think we could just make `gp.subplots` a private attribute that stores the subplots as a numpy array, and `gp[i, j].subplot` just calls the appropriate one from private `gp._subplots` which is a numpy array of subplot objects.

fastplotlib/layouts.py
```diff
@@ -49,6 +55,16 @@ def __init__(
...
+        positions = list()
+        for x in range(self.shape[0]):
+            for y in range(self.shape[1]):
+                positions.append((x, y))
+
+        self.indexers = list()
+        for i in range(self.shape[0] * self.shape[1]):
+            self.indexers.append(GridPlotIndexer())
+            self.indexers[i].index = positions[i]
+
```
kushalkolar, line 67: add just `__getitem__`

I think this should be read-only, we don't ones users messing with the subplots manually in the gridplot so `__setitem__` is not necessary (I think).

fastplotlib/layouts.py
```diff
@@ -111,9 +119,17 @@ def __init__(
...
 
+            self.indexers[i, j] = GridPlotIndexer((i,j))
+
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
+    def __getitem__(self, index: tuple):
+        return self.indexers[index[0], index[1]].name
+
+    def __setitem__(self, index: tuple, name: str):
```
kushalkolar, line 130: this is not necessary, the indexer should be read-only. Since `__getitem__()` returns an instance of `GridPlotIndex`, it can be named right there.

example:
```python
indexer = gp[i, j] # returns `GridPlotIndexerInstance`
indexer.name = "bah"

# can also just directly do
gp[i, j].name = "bah"
```

fastplotlib/layouts.py
kushalkolar: @clewis7 you had a question about this today, take a look at it again and if you have questions I will clarify Friday.

kushalkolar: We could also deprecate `GridPlot.subplots` and just make the subplots an attribute of `GridPlotIndexer` instead.

kushalkolar (changes requested):

fastplotlib/layouts.py
```diff
@@ -111,9 +119,17 @@ def __init__(
                 renderer=renderer
             )
 
+            self.indexers[i, j] = GridPlotIndexer((i,j))
+
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
+    def __getitem__(self, index: tuple):
+        return self.indexers[index[0], index[1]].name
```
kushalkolar, line 128: you don't return the name, just the instance of `GridPlotIndexer`

kushalkolar: Also, we want to allow the `index` argument to be either a `tuple[int, int]` or `str` with the name. It then iterates through all `self.indexers` until it finds a match.

fastplotlib/layouts.py
```diff
@@ -7,6 +7,12 @@
 from wgpu.gui.auto import WgpuCanvas
 
 
+class GridPlotIndexer:
+    def __init__(self, position):
+        self.name: str = None
```
kushalkolar, line 12: We need a way to make sure that each `GridPlotIndexer` instance within a `GridPlot` is unique, so we can add the attribute `parent` which is the parent `GridPlot` instance.

We should make `self.name` a property, so when the `name` is being set it checks to make sure it's not used by another `GridPlot` instance

kushalkolar (changes requested):

kushalkolar:
I thought about this a bit more and it makes more sense for `name` to be an attribute of `Subplot`.

`Subplot` already has the attribute `position` which acts similar to `GridPlotIndexer.index` 

more comments in the code

fastplotlib/layouts.py
kushalkolar, line 12: move `name` to `Subplot`, get rid of `GridPlotIndexer` class. 

`position` is already an attribute of `Subplot` so don't need to worry about that.

fastplotlib/layouts.py
```diff
@@ -111,9 +119,19 @@ def __init__(
...
             )
 
+            self.indexers[i, j] = GridPlotIndexer((i,j))
+
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
+    def __getitem__(self, index: Union[tuple[int, int], str]):
+        if type(index) == str:
+            for indexer in self.indexers:
```
kushalkolar, line 129: Going off from the comment in the PR, iterate through `self.subplots` instead of the indexers.

fastplotlib/layouts.py
```diff
@@ -111,9 +119,19 @@ def __init__(
                 renderer=renderer
             )
 
+            self.indexers[i, j] = GridPlotIndexer((i,j))
+
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
+    def __getitem__(self, index: Union[tuple[int, int], str]):
```
kushalkolar, line 127: I think capital T for `Tuple` in the type annotation is required for python < 3.10

kushalkolar:
Add `name` as a `kwarg` to `Subplot` and `names` as a kwarg to `Gridplot` so that the names can be set from the beginning.

clewis7:
naming functionality now works as outlined in #36 

take a look and let me know if I need to make any additional changes, or could improve elegancy!!

kushalkolar:
can you do this without the heatmap branch being merged into here? that branch Is too rudimentary and I want to merge this first

fastplotlib/layouts.py
```diff
@@ -105,32 +103,38 @@ def __init__(
...
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
-    def __getitem__(self, index: Union[tuple[int, int], str]):
+    def __getitem__(self, index: Union[Tuple[int, int], str]):
         if type(index) == str:
-            for indexer in self.indexers:
-                if indexer.name == index:
-                    return indexer
+            for subplot in np.concatenate(self.subplots):
```
kushalkolar, line 133: I would use `self.subplots.ravel()` instead since it provides a view whereas `np.concatenate` returns a copy of the array

kushalkolar (changes requested):

fastplotlib/subplot.py
```diff
@@ -88,6 +96,10 @@ def add_animations(self, funcs: List[callable]):
     def add_graphic(self, graphic, center: bool = True):
         self.scene.add(graphic.world_object)
 
+        if isinstance(graphic, Heatmap):
```
kushalkolar, line 99: remove

fastplotlib/subplot.py
```diff
@@ -1,9 +1,11 @@
 import pygfx
 from pygfx import Scene, OrthographicCamera, PerspectiveCamera, PanZoomController, Viewport, AxesHelper, GridHelper
 from .defaults import camera_types, controller_types
+from .graphics import Heatmap
```
kushalkolar, line 4: remove

kushalkolar (changes requested):
residual heatmap stuff

clewis7:
still need to implement helpful `__repr__` methods for Graphic, GridPlot, and Subplot

fastplotlib/layouts.py
```diff
@@ -102,18 +110,32 @@ def __init__(
             camera = cameras[i, j]
             controller = self._controllers[i, j]
 
-            self.subplots[i, j] = Subplot(
+            if self.names is not None:
+                name = self.names[i][j]
```
kushalkolar, line 114: use regular array indexing, `[i, j]`

fastplotlib/subplot.py
```diff
@@ -86,6 +95,15 @@ def add_animations(self, funcs: List[callable]):
         self._animate_funcs += funcs
 
     def add_graphic(self, graphic, center: bool = True):
+        graphic_names = list()
+        for g in self._graphics:
+            graphic_names.append(g.name)
+
+        if graphic.name in graphic_names:
+            print(graphic_names)
+            raise ValueError("graphics must have unique names")
```
kushalkolar, line 104: this is better:

```python
raise ValueError(f"graphics must have unique names, current graphic names are:\n {graphic_names}")

File: fastplotlib/subplot.py [hunk 13]
```diff
@@ -140,3 +158,18 @@ def set_grid_visibility(self, visible: bool):
...
+
+    def __getitem__(self, name: str):
+        for graphic in self._graphics:
+            if graphic.name == name:
+                return graphic
+        graphic_names = list()
+
+        for g in self._graphics:
+            graphic_names.append(g.name)
+        print("the graphics that currently exist are: ")
```
kushalkolar, line 173: put within the exception message, see previous comment

kushalkolar (changes requested):

kushalkolar:
Can you update the example notebooks according to this PR

fastplotlib/subplot.py
```diff
@@ -140,3 +157,16 @@ def set_grid_visibility(self, visible: bool):
...
+
+    def __getitem__(self, name: str):
+        for graphic in self._graphics:
+            if graphic.name == name:
+                return graphic
+        
+        graphic_names = list()
+        for g in self._graphics:
+            graphic_names.append(g.name)
+        raise ValueError(f"no graphic of given name, the current graphics are:\n {graphic_names}")
```
kushalkolar, line 172: Can you change this from `ValueError` to `IndexError`

fastplotlib/layouts.py
```diff
@@ -102,18 +110,32 @@ def __init__(
...
             )
 
         self._animate_funcs: List[callable] = list()
         self._current_iter = None
 
+    def __getitem__(self, index: Union[Tuple[int, int], str]):
+        if type(index) == str:
+            for subplot in self._subplots.ravel():
+                if subplot.name == index:
+                    return subplot
```
kushalkolar, line 135: After this return can you raise IndexError for "no Subplot with given name" or something

kushalkolar (changes requested):
Minor error handling

fastplotlib/layouts.py
```diff
@@ -133,6 +133,8 @@ def __getitem__(self, index: Union[Tuple[int, int], str]):
             for subplot in self._subplots.ravel():
                 if subplot.name == index:
                     return subplot
+                else:
```
kushalkolar, line 136: nope this will always raise after the first iteration, you need just a raise after the loop

kushalkolar (changes requested):

kushalkolar (approved):

kushalkolar:
great job on your first contribution! :D

Issue #29, opened 2022-10-05
image histogram UI


kushalkolar:
Ideas for creating a histogram UI to adjust vmin-vmax of image graphic similar to `pyqtgraph` `HistogramLUTWidget` 

- [ ] When used, the `viewport` in the `Subplot` needs to be split so that a thin rectangular "sub-viewport" can be used for the histogram LUT widget.
- [ ] Colorbar next to histogram

kushalkolar:
more ideas on the HistogramLUTWidget itself:

1. Disable panning in x-direction, only allow y-axis panning

kushalkolar:
Superseded by a more recent issue

PR #30, opened 2022-10-11
"DockedViewport" - Viewports adjacent to `Subplot` `Viewport`

kushalkolar:
start implementing #29


kushalkolar:
"docked subplots" now work on any side of a subplot, this will allow making a histogram LUT widget similar to pyqtgraph's

image

things to consider:
1. make a "view area" class which is kinda like a `Subplot`, but they can only exist as children of a subplot and are shown within the original Subplot viewport area:

```
PlotArea:
    attrs:
        parent
        position
        renderer
        camera
        controller
    methods:
        get_rect()
        set_viewport_rect()
        render()
        add_graphic()
        center_graphic()
...
```

2. Start a module for widgets like the `HistogramLUTWidget` which are more complex than just single `Graphic` items.

fastplotlib/subplot.py
```diff
@@ -59,21 +60,34 @@ def __init__(
...
+        self.set_viewport_rect()
 
+    def get_rect(self):
+        i, j = self.position
         w, h = self.renderer.logical_size
 
         spacing = 2  # spacing in pixels
 
-        self.viewport.rect = [
+        return np.array([
```
kushalkolar, line 77: I really need to write this better, like:

x_pos =...
y_pos =...
width =...

viewport.rect = (x, y, w, h)

Issue #36, opened 2022-11-01
Naming subplots and graphics


kushalkolar:
- [x] `__getitem__` for `GridPlot`, index the subplots using tuple of `int` or `str` by giving subplots names. `Subplot` has `name` attribute. #27 
  ```python
  gridplot[0, 0] # returns Subplot at [0, 0]
  gridplot[0, 0].name = "bah"
  gridplot["bah"] # returns Subplot at 0, 0
  ```
- [x] each `Graphic` can also be given a `name` attribute for referencing it. `Subplot.__getitem__` can take `str` to reference the `Graphic` items.

- [x] implement a way to get `Graphic` instances from a subplot. 

- [x] We can also create a useful `__repr__` for `Gridplot`, `Subplot` and `Graphic`.

kushalkolar:
for the 2nd point, to implement for `Graphic` items, basically use the same methods that you implement for `Gridplot` but in `Subplot`. 

In the end, this should work for retrieving things:

```python3
gp["subplot-name"]["graphic-name"]
```

PR #43, opened 2022-11-09
#36 task 4, repr implementation 

clewis7:
just started with `__repr__` implementations for the each of the graphics, could not think of anything else to add but:
- graphic type
- name if not `None`
- hex location 
- data

if there is something else you want me to add to the graphic `__repr__` methods, just lmk

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -62,3 +62,12 @@ class Graphic:
     def update_data(self, data: Any):
         pass
 
+    def __repr__(self):
+        if self.name is not None:
+            print("Graphic Name: " + self.name + "\n")
+            print("Graphic Location: " + hex(id(self)) + "\n")
+        else:
+            print("Graphic Location: " + hex(id(self)) + "\n")
+        print(self.data)
+        return ""
+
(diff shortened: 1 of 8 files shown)


fastplotlib/graphics/_base.py
```diff
@@ -61,3 +61,13 @@ def children(self) -> pygfx.WorldObject:
 
     def update_data(self, data: Any):
         pass
+
+    def __repr__(self):
+        title = "Fastplotlib " + self.__class__.__name__ + "\n"
```
kushalkolar, line 66: returning "fastplotlib" is kinda redundant, how about something like this which is shorter:

```
"user-given-name" <__class__.__name__> Graphic @ <hex-id>
```

fastplotlib/subplot.py
```diff
@@ -170,3 +170,15 @@ def __getitem__(self, name: str):
         for g in self._graphics:
             graphic_names.append(g.name)
         raise IndexError(f"no graphic of given name, the current graphics are:\n {graphic_names}")
+
+    def __repr__(self):
+        title = "Fastplotlib " + self.__class__.__name__ + "\n"
```
kushalkolar, line 175: Redundant, can just be `__class__.__name__`

rest LGTM

kushalkolar (changes requested):

kushalkolar (approved):
LGTM, we can worry about gridplot and other stuff later.

Issue #39, opened 2022-11-05
very high level interactivity API


kushalkolar:
Just a vague idea for now. Currently basic interactivity is implemented with the `pygfx` events system for the [`Heatmap`](https://github.com/kushalkolar/fastplotlib/pull/32). We can think of a high level interactivity API where certain features across Graphic can linked together by a selection of built-in interactivity. 

Usage example:

```python
hm_data = np.random.rand(100, 20)
data_indices = np.arange(20)

contours = # get 20 contours

hm_graphic = plot.heatmap(hm_data, indices=data_indices)

# plot the other graphic in the same or some other scene

contour_graphic = plot.multi_line(contours, indices=data_indices)

hm_options = dict(
    feature="row",
    update_color=[1, 1, 1, 0.5],
    update_data=None,  # can change the data if desired
    event="double_click"
)

contour_options = dict(
    feature="line",
    update_color=[1, 1, 1, 1],
    update_data=None,  # if desired
    event="click"
)

# bi-directional linking
# heatmap -> contours
linker = hm_graphic.link(source_options=hm_options, target=contour_graphic, target_options=contour_options)

# contours -> heatmap
linker2 = contour_graphic.link(source_options=contour_options, target=hm_graphic, target_options=hm_options)

# to unlink
hm_graphic.unlink(linker)
```

Also allow custom indices mapper to make the data from graphics where the indices don't directly map onto the indices of another graphic. For examples Image -> Contour 

```python
# continue from above, so the previous heatmap is also present somewhere in the renderer/canvas

img_data = np.random.rand(512, 512)

image_graphic = plot.image(img_data)

def get_closest_contour(indices) -> np.array:
    # convert source graphic indices to target graphic indices
    # stuff to map image indices -> contour indices
    return contour_indices

image_options = dict(
    feature="pixel",
    update_data=None, # we don't want to change anything in the image
    event="click",
    indices_mapper=get_closest_contour,
)

contour_options_img = dict(
    feature="line",
    update_color=[1, 1, 1, 1],
    event=None,
    chain_event=True, # this graphic will pass on the source's event to its target if a target exists
)

linker_img = image_graphic.link(source_options=image_options, target=contour_graphic, target_options=contour_options_img)
```

When `image_graphic` gets a `click` event it triggers the `contour_graphic` to change its "line" feature, but since we set `chain_event=True` for `contour_options_img`, the `contour_graphic` will pass along the `click` event to anything that it is linked to downstream.

[...]

PR #61, opened 2022-12-12
beginning base logic for interactivity impl

clewis7:
starting #39


kushalkolar:
maybe an even simpler way to implement this:

`Interaction.link` is just a very simple method that sets the "feature" at the given indices with the new data for **any** `Graphic` that is passed, not just self. 

```python3

def link(
  self,
  event: str, 
  target: Graphic, 
  feature: str, 
  new_data: Any, 
  indices_mapper: callable = None
):
    # event picks subset of self.indices -> indices_mapper() -> indices_subset -> target._set_feature(new_data, indices_subset)
```

fastplotlib/graphics/line.py
```diff
@@ -53,3 +56,32 @@ def update_colors(self, colors: np.ndarray):
...
+        valid_events = ["click"]
+        if event in valid_events:
+            self.world_object.add_event_handler(self.event_handler, event)
+            self.events[event] = list()
+            self.events[event].append((target, feature, new_data))
+        else:
+            raise ValueError("event not possible")
+
+    def event_handler(self, event):
+        for event in self.events[event]:
```
kushalkolar, line 86: You are setting the variable during each loop iteration as the dict key

fastplotlib/graphics/line.py
```diff
@@ -53,3 +56,32 @@ def update_colors(self, colors: np.ndarray):
...
+        elif name == "data":
+            self.update_data(new_data)
+        else:
+            raise ValueError("name arg is not a valid feature")
+
+    def link(self, event: str, target: Graphic, feature: str, new_data: Any, indices_mapper: callable = None):
+        valid_events = ["click"]
+        if event in valid_events:
+            self.world_object.add_event_handler(self.event_handler, event)
+            self.events[event] = list()
```
kushalkolar, line 80: This should be an instance variable

kushalkolar: Should be set in the init of the graphic, not here it'll reset every time it's called

fastplotlib/graphics/line.py
```diff
@@ -53,3 +56,32 @@ def update_colors(self, colors: np.ndarray):
...
+    @property
+    def indices(self) -> Any:
+        return self.indices
+
+    @property
+    def features(self) -> List[str]:
+        return self.features
+
+    def _set_feature(self, name: str, new_data: Any, indices: Any):
+        if name == "color":
```
kushalkolar, line 69: Use getattr instead of if statements

kushalkolar (changes requested):

fastplotlib/graphics/line.py
kushalkolar: This works:

```python3
        if event.type in self.events.keys():
            for target_info in self.events[event.type]:
                target_info[0]._set_feature(name=target_info[1], new_data=target_info[2], indices=None)
```

fastplotlib/graphics/_base.py
```diff
@@ -1,11 +1,12 @@
-from typing import Any
+from typing import Any, List
```
kushalkolar, line 1: just import *

fastplotlib/graphics/_base.py
```diff
@@ -101,4 +102,14 @@ def link(self, event: str, feature: Any, feature_data: Any, target: Graphic, tar
         # indice mapper takes in source features and maps to target features
         pass
 
+    @abstractmethod
+    def event_handler(self, event):
```
kushalkolar, line 106: let's implement event_handler as a normal method in the base class

fastplotlib/graphics/line.py
```diff
@@ -24,7 +25,7 @@ def __init__(self, data: np.ndarray, zlevel: float = None, size: float = 2.0, co
             material=material(thickness=size, vertex_colors=True)
         )
 
-        self.events = {}
+        self.registered_callbacks = {}
```
kushalkolar, line 28: set this in the base `Graphic`, and use dict() to create a new dict (there are caveats to initializing dicts and lists with expressions like this that can lead to weird bugs)

kushalkolar (changes requested):

fastplotlib/graphics/line.py
clewis7: I don't understand this one

fastplotlib/graphics/_base.py
```diff
@@ -70,6 +71,12 @@ def __repr__(self):
         else:
             return f"fastplotlib.{self.__class__.__name__} @ {hex(id(self))}"
 
+    def event_handler(self, event):
```
kushalkolar, line 74: put this in the `Interaction` class

fastplotlib/graphics/_base.py
```diff
@@ -102,14 +109,11 @@ def link(self, event: str, target: Graphic, feature: str, new_data: Any, indices
...
 
-    @abstractmethod
-    def event_handler(self, event):
-        pass
-
 @dataclass
 class EventData:
     """Class for keeping track of the info necessary for interactivity after event occurs."""
-    def __init__(self, target: Graphic, feature: str, new_data: Any):
+    def __init__(self, target: Graphic, feature: str, new_data: Any, indices: Any):
```
kushalkolar, line 115: this doesn't keep indices, the indices are unique to every event

fastplotlib/graphics/line.py
```diff
@@ -81,12 +79,7 @@ def link(self, event: str, target: Graphic, feature: str, new_data: Any, indices
             raise ValueError("event not possible")
 
         if event in self.registered_callbacks.keys():
-            self.registered_callbacks[event].append(EventData(target=target, feature=feature, new_data=new_data))
+            self.registered_callbacks[event].append(EventData(target=target, feature=feature, new_data=new_data, indices=None))
```
kushalkolar, line 82: indices are obtained when the event happens, not here

kushalkolar: if you rename `event` to `event_type` here in `link()` I think it'll help become clearer

kushalkolar (changes requested):

fastplotlib/graphics/line.py
kushalkolar: https://docs.python.org/3/library/functions.html#getattr

```python3
update_func = gettattr(self, f"update_{name}")
update_func(new_data, indices=indices)
```

kushalkolar: you will need to implement a way for the update functions to use the indices

kushalkolar (changes requested):
almost there :D

clewis7:
Made the changes that you requested, I'm sure there are further changes needed to be made

going to move on to linecollection and try to get that up and running!

fastplotlib/graphics/_base.py
```diff
@@ -68,3 +71,39 @@ def __repr__(self):
...
+
+    @property
+    @abstractmethod
+    def features(self) -> List[str]:
+        pass
+
+    @abstractmethod
+    def _set_feature(self, name: str, new_data: Any, indices: Any):
+        pass
+
```
kushalkolar, line 93: you probably need a method `_reset_feature()`

fastplotlib/graphics/_base.py
```diff
@@ -68,3 +71,39 @@ def __repr__(self):
...
+    def link(self, event: str, target: Graphic, feature: str, new_data: Any, indices_mapper: callable = None):
+        pass
+
+    def event_handler(self, event):
+        if event.type in self.registered_callbacks.keys():
+            for target_info in self.registered_callbacks[event.type]:
+                target_info.target._set_feature(feature=target_info.feature, new_data=target_info.new_data)
+
+@dataclass
+class EventData:
```
kushalkolar, line 104: this is really more of callback data than event data, because the event data comes from pygfx handling and `target`, `feature` and `new_data` are callback information.

Also this isn't how you construct a dataclass, you don't define `__init__` 
https://docs.python.org/3/library/dataclasses.html

fastplotlib/graphics/line.py
```diff
@@ -50,3 +50,31 @@ def update_colors(self, colors: np.ndarray):
...
+        return None
+
+    def _set_feature(self, feature: str, new_data: Any, indices: Any = None):
+        if feature in ["colors", "data"]:
+            update_func = getattr(self, f"update_{feature}")
+            update_func(new_data)
+        else:
+            raise ValueError("name arg is not a valid feature")
+
+    def link(self, event_type: str, target: Graphic, feature: str, new_data: Any, indices_mapper: callable = None):
```
kushalkolar, line 69: I think this can also just be moved to the base Interaction, what do you think?

fastplotlib/graphics/_base.py
clewis7: for resetting the feature, would you want to store the "old data", for example the previous color, somewhere? and if so, how would you recommend doing that?

my thought would be:
- store the old data/color somewhere
- when event handler is called would compare the data (if equal to old then set_feature is called and if it is already the new data then you would call reset_feature)...but then what is the difference between just calling set_feature and passing the old_data as a param

not sure, what are your thoughts?

kushalkolar: yes we can store the old data and the indices they come from. 

`_reset_feature` is called every time the `event_handler` handles an event in order to set everything back to how it was originally before changing the feature with new data.

fastplotlib/graphics/line.py
```diff
@@ -1,13 +1,15 @@
...
+from ._base import Graphic, CallbackData, Interaction
 
-
-class LineGraphic(Graphic):
-    def __init__(self, data: np.ndarray, z_position: float = 0.0, size: float = 2.0, colors: np.ndarray = None, cmap: str = None, *args, **kwargs):
+class LineGraphic(Graphic, Interaction):
+    def __init__(self, data: np.ndarray, zlevel: float = None, size: float = 2.0, colors: np.ndarray = None, cmap: str = None, *args, **kwargs):
         super(LineGraphic, self).__init__(data, colors=colors, cmap=cmap, *args, **kwargs)
 
+        self.zlevel = zlevel
```
kushalkolar, line 11: what happened here, I renamed this to `z_position`

fastplotlib/graphics/linecollection.py
```diff
@@ -1,16 +1,20 @@
...
 
-class LineCollection():
-    def __init__(self, data: List[np.ndarray], z_position: Union[List[float], float] = None, size: Union[float, List[float]] = 2.0, colors: Union[List[np.ndarray], np.ndarray] = None,
+class LineCollection(Interaction):
+    def __init__(self, data: List[np.ndarray], zlevel: Union[List[float], float] = None, size: Union[float, List[float]] = 2.0, colors: Union[List[np.ndarray], np.ndarray] = None,
                  cmap: Union[List[str], str] = None, *args, **kwargs):
 
-        if not isinstance(z_position, float) and z_position is not None:
-            if not len(data) == len(z_position):
+        if not isinstance(zlevel, float) and zlevel is not None:
```
kushalkolar, line 16: same as above, I renamed this to `z_position`

clewis7: oops, was trying to fix stupid merge conflicts

should be fixed now

clewis7:
sorry for all of the commits, git was not my friend for a moment but everything should be fine now I hope

fastplotlib/graphics/image.py
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 from typing import Tuple, Any, List
+=======
+from typing import *
```
kushalkolar, line 4: you have a merge conflict that slipped through somehow, just keep this line and delete the ones around it

clewis7:
So, I got it to work with line collection

things are not perfect by any means but got a nice dopamine rush...hopefully indicative of things moving in the right direction

fastplotlib/graphics/_base.py
```diff
@@ -93,34 +90,53 @@ def _set_feature(self, feature: str, new_data: Any, indices: Any):
...
+            old_data = list()
+            for line in target.data:
+                old_data.append(getattr(line, feature).copy())
+        else:
+            old_data = getattr(target, feature).copy()
+
         if event_type in self.registered_callbacks.keys():
             self.registered_callbacks[event_type].append(
-                CallbackData(target=target, feature=feature, new_data=new_data, old_data=getattr(target, feature).copy()))
+                CallbackData(target=target, feature=feature, new_data=new_data, old_data=old_data, indices_mapper=indices_mapper))
```
kushalkolar, line 112: `old_data` shouldn't be in `CallbackData`, `old_data` should be managed by the target graphic event time it handles an event. When the target graphic handles an event, it should have a copy of the feature data at the indices that are being modified and reset the feature at those indices, and then set the feature w.r.t. to the new `CallbackData`.

To elaborate:
1. The target graphic keeps a private dict of `(features, orig_vals)` that were modified at certain indices from the previous callback, something in the form `{"feature": (indices, orig_vals_at_indices)}`, you could call it `self._previous_modified`. And with the new `GraphicFeature` class from #78 we won't even need a dict for it, we can just add `_modified_indices` and `_orig_vals_at_modified_indices` as private attributes to `GraphicFeature` :smile: .

2. In `_set_feature()` it first uses the information from above to reset the feature at those indices to their original values

fastplotlib/graphics/_base.py
```diff
@@ -93,34 +90,53 @@ def _set_feature(self, feature: str, new_data: Any, indices: Any):
...
         if event.type in self.registered_callbacks.keys():
             for target_info in self.registered_callbacks[event.type]:
-                target_info.target._reset_feature(feature=target_info.feature, old_data=target_info.old_data, indices=None)
-                target_info.target._set_feature(feature=target_info.feature, new_data=target_info.new_data, indices=None)
+                # need to map the indices to the target using indices_mapper
+                if target_info.indices_mapper is not None:
+                    indices = target_info.indices_mapper(target=target_info.target, indices=click_info)
+                else:
+                    indices = None
+                # reset feature of target using stored old data
```
kushalkolar, line 130: ah I see you've done it like this, but it's better for the target graphic to handle old data because we want the "feature resetting at previously modified indices" to happen regardless of the event type. For example, if we've registered different types of callbacks to keyboard events, and different mouse events, we want the same reset to happen regardless of what previous `event.type` triggered the previous feature data to change.

fastplotlib/graphics/image.py
```diff
@@ -76,18 +72,14 @@ def __init__(
...
         )
 
-    @property
-    def indices(self) -> Any:
-        pass
-
     @property
     def features(self) -> List[str]:
-        pass
+        return ["cmap", "data"]
```
kushalkolar, line 77: This won't be necessary after the new #78  :smile: (keep it in for now for testing w.r.t. current `master` branch)

fastplotlib/graphics/linecollection.py
```diff
@@ -2,10 +2,9 @@
 import pygfx
 from typing import Union, List
 
-from .line import LineGraphic
+from fastplotlib.graphics.line import LineGraphic
```
kushalkolar, line 5: always use relative imports from within a package, I'm guessing this was pycharm's fault

fastplotlib/graphics/linecollection.py
```diff
@@ -2,10 +2,9 @@
...
 from typing import *
-from ._base import Interaction
-
+from fastplotlib.graphics._base import Interaction
```
kushalkolar, line 7: same as above, always use relative imports from within a package, I'm guessing this was pycharm's fault

fastplotlib/graphics/linecollection.py
kushalkolar: it likes to do this when doing the automated refactoring :weary:

fastplotlib/graphics/linecollection.py
```diff
@@ -49,33 +48,30 @@ def __init__(self, data: List[np.ndarray], z_position: Union[List[float], float]
...
     def features(self) -> List[str]:
-        pass
+        return ["colors", "data"]
 
     def _set_feature(self, feature: str, new_data: Any, indices: Any):
-        if feature in ["colors", "data"]:
-            update_func = getattr(self, f"update_{feature}")
-            self.collection[indices].update_func(new_data)
+        if feature in self.features:
+            update_func = getattr(self.data[indices], f"update_{feature}")
```
kushalkolar, line 59: the way it's written it'll only work with single integer indices, we can chat about it

kushalkolar: once we implement this it'll be much easier https://github.com/kushalkolar/fastplotlib/issues/76#issuecomment-1358903126 

it would become `getattr(self, feature)[indices] = new_data`

fastplotlib/graphics/linecollection.py
```diff
@@ -49,33 +48,30 @@ def __init__(self, data: List[np.ndarray], z_position: Union[List[float], float]
...
-    def indices(self, indices: Any):
-        pass
+            self.data.append(LineGraphic(d, _z, _size, _colors, _cmap))
 
     @property
     def features(self) -> List[str]:
-        pass
+        return ["colors", "data"]
 
     def _set_feature(self, feature: str, new_data: Any, indices: Any):
```
kushalkolar, line 57: I think that for the `_set_feature` method in each `Graphic` or `GraphicCollection` we can replace `Any` for the indices to the indices that are appropriate for that graphic type.

kushalkolar (commented):
great work! you can continue playing with it, once #78 is ready (very soon!) it'll be much easier to do the rest

fastplotlib/graphics/_base.py
kushalkolar: with #78 it might be possible for us to just pass the GraphicFeature instance _directly_ instead of the string name when calling `link`, for example:

`graphic.link(event_type, target=some_graphic, feature=some_graphic.color...`

fastplotlib/graphics/_base.py
```diff
@@ -68,3 +71,72 @@ def __repr__(self):
...
+    def event_handler(self, event):
+        if event.type == "click":
+            # storing click information for each click in self.indices
+            #self.indices(np.array(event.pick_info["index"]))
+            click_info = np.array(event.pick_info["index"])
+        if event.type in self.registered_callbacks.keys():
+            for target_info in self.registered_callbacks[event.type]:
+                # need to map the indices to the target using indices_mapper
+                if target_info.indices_mapper is not None:
+                    indices = target_info.indices_mapper(target=target_info.target, indices=click_info)
```
kushalkolar, line 127: an `indices_mapper` function should be given source and target, something like this:

```
indices_mapper(source=self, target=target_info.target, event_indices=click_info
```

And the user can write their function to take in all 3 and do whatever they want.

kushalkolar:
Now that #78 brings events to `GraphicFeatures` we want to make `link()` work seamlessly for both `pygfx` events and `GraphicFeature` events. Something like this should work:

```python
def link(self, event_type: str, target: Any, ...):
  pygfx_events = ["click", ...]
  graphic_feature_events = ["colors", "data", "present", ...]

  if event_type in pygfx_events:
    self.world_object.add_event_handler(...)

  elif event_type in graphic_feature_events:
    feature = getattr(self, event_type)
    feature.add_event_handler(...)
```

kushalkolar:
should probably also generalize the `indices_mapper` to just be an event mapper that get access to the source, target, and all event info. For example if we want click events on a contour to toggle the `present` feature of some other graphic it's diff from indices mapping.

fastplotlib/graphics/_base.py
clewis7: If info for resetting feature is moved to ```_set_feature()``` graphic, then we will no longer need method for ```_reset_feature()```?

fastplotlib/graphics/_base.py
clewis7: I agree! Just wasn't sure at the time the best way to implement, but now that graphic features exist it should be a lot easier to store old data!

fastplotlib/graphics/_base.py
clewis7: yes, I assumed that some general form would need to be established

fastplotlib/graphics/_base.py
kushalkolar: we should still have a `_reset_feature()` to separate functionality

Issue #64, opened 2022-12-13
better organization of Graphics


kushalkolar:
1. We will soon have `LineCollection` so we should probably add a class attribute to each `Graphic` subclass which is used by `Plot` when attaching all the graphic names as methods, since we'd want `plot.line_collection()` rather than `plot.linecollection()` and it maintains pep8 standards
2. It would be useful for `Subplot.get_graphics()` to return a dict instead of a list, where the keys are the object class.
example:

```python
from fastlotlib.graphics import ImageGraphic

graphics_in_subplot = gp[0, 0].get_graphics() 

graphics_in_subplot[ImageGraphic] # list of all `ImageGraphic` graphic instances within the subplot
```

@clewis7 thoughts?

clewis7:
I like this idea! I can work on making ```Subplot.get_graphics()``` return a dict instead of a list

kushalkolar:
I think it'd be useful to use the __init_subclass__ to parse Graphic class names automatically and set it as a class attribute that can be used for this purpose, but also for attaching the various add_graphic-name methods, so we can have this parsing code in just one place.

This would also make it so that the dict keys for this PR are str, I think str is better since then the user doesn't need another set of import statements for graphic objects.

kushalkolar:
printing it out like a table or just with nice tabspacing would be nice:

```yaml
images:
  "image-name-a" <mem location etc...>
line_collection:
  "lc-A" <n-items>, ...
```

PR #65, opened 2022-12-13
storing graphics in dict

clewis7:
addresses #64

moved to store graphics in dict with keys as the graphic class

keys currently need to passed in as a string, need to work on implementation to allow for 

```python 
gp[0,0].get_graphics()[ImageGraphic]
```
as opposed to 
```python
gp[0,0].get_graphics()["ImageGraphic"]

### DIFF
--- a/fastplotlib/layouts/_base.py
+++ b/fastplotlib/layouts/_base.py
@@ -39,7 +39,9 @@ class PlotArea:
 
         self.renderer.add_event_handler(self.set_viewport_rect, "resize")
 
-        self._graphics: List[Graphic] = list()
+        self._graphics = {}
+        # \
+        #     List[Graphic] = list()
 
         self.name = name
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by clewis7** [comment]
right now, I have it creating the keys to the dict by 
```python 
self._graphics[graphic.__class__.__name__] = list()
self._graphics[graphic.__class__.__name__].append(graphic)
```
but ```__name__``` returns a string, any suggestions on how to get graphic class name not as a string when just ```__class__``` gives ```fastplotlib.graphics.image.ImageGraphic```

kushalkolar:
yea not sure whether the best approach is to have it as a `str` or the class type itself, I think using `graphic.__class__` should work, you can check if: 

```python3
graphic.__class__ is ImageGraphic
```

kushalkolar:
after this is done modify the `__repr__` for `PlotArea` such that if there are > 5 graphics of any type, show only the first 5 and last 5 because I know there will be _someone_ at the workshop who crashes their jupyter because they ended up calling the repr while having 10,000 graphics in a gridplot :upside_down_face:

kushalkolar:
Now that #78 brings events to `GraphicFeatures` we want to make `link()` work seamlessly for both `pygfx` events and `GraphicFeature` events. Something like this should work:

```python
def link(self, event_type: str, target: Any, ...):
  pygfx_events = ["click", ...]
  graphic_feature_events = ["colors", "data", "present", ...]

  if event_type in pygfx_events:
    self.world_object.add_event_handler(...)

  elif event_type in graphic_feature_events:
    feature = getattr(self, event_type)
    feature.add_event_handler(...)
```

kushalkolar:
after #88 there is now a class attribute that stores the name of the graphic as a str in lower case, see: https://github.com/kushalkolar/fastplotlib/pull/88/files#diff-a57994ecff1cde4e05c8175265219f0b5200da144ce8733b9eb560c09bcea7ddR9

kushalkolar:
rebase on mater or just start a new PR, this is pretty straight forward.

kushalkolar:
I think that since `PlotAreas` are indexable we don't really need this. We can revisit if it becomes necessary in the future.

Issue #76, opened 2022-12-18
LineStack


kushalkolar:
Just realized what's probably the best way to do this:

Every Graphic sets `world_object.position` after init based on a position kwarg (probably expose position as a user settable attribute, maybe start making worldobject itself private).
In LineStack, as we iterate through adding each Line, we set position.y as the previous position.y + data[:, 1].max() + separation_height

kushalkolar:
we should make collections indexable, something like this would be nice:

```python
line_collection[:50].cmap = "jet"
line_collection[50:].colors = "r"

# and it gives this insane level of fine grained control
line_collection[50::2].colors[:10] = "b"

# removes anything with visible = False from the scene
# and then re-stacks only the visible lines and auto-scales
line_stack[10:].visible = False
```

This should be possible if `Collection.__getitem__` returns an object with the same interface as the corresponding Graphic, and just iterates after parsing indices from the slice.

other features for `LineStack`

- line_stack.reorder(new_index_order), reorders the stacking from bottom to top

kushalkolar:
or actually it would be better to have it more similar to the indexing in non-collection Graphics, need to think of how to do this. We have a base `BaseCollectionFeature` which has slice parsing and then each `CollectionColor` for example uses the indices to assign the feature to each graphic within the collection.

kushalkolar:
starting point:

```python3
plot = Plot()

y_zero = 0
y_sep = 10
for i, c in enumerate(data[:]):
    if i > 0:
        y_zero += data[i - 1].max()
        c = c + y_zero + y_sep
    
    color = np.random.rand(4)
    
    plot.add_line(np.dstack([np.arange(c.size), c])[0], colors=color)

plot.show()
```

kushalkolar:
This will make it very easy to make plots where we can combine colors within individuals lines and across lines in a collection to representing neural state space across both trials AND time simultaneously.

For example:

first 50 Early trials are bluish, last 50 trials are reddish:

```python3
# start with white
line_collection.colors = "w"

gradient = np.linspace(0, 1, 50)

# first 50 trials bluish
line_collection[:50].colors[:, 2] = gradient

# indicate time with increasing green
line_collection[:50].colors[:, 1] = gradient

# last 50 trials reddish
line_collection[50:].colors[:, 0] = gradient

# indicate time with increasing green
line_collection[50:].colors[:, 1] = gradient
```

PR #90, opened 2022-12-27
graphic collections are indexable and so are their features

kushalkolar:
implements ideas from #76 for basic line collection, still need to do LineStack

@clewis7 The `add_event_handler()` (see below) just adds it to each individual feature instance, the `pick_info` gives out a `collection_index` key. I think this is fine for implementing `link()` on the `LineCollection` level. We can think of a better way later. 

All this is possible, tested and works:

```python
lines[10:15].colors = "g"
lines[10:15].colors[30:60] = "cyan"
lines[[1, 2, 5]].colors = "w"
lines[[16, 19]].colors[::2] = "w"
lines[2:6].colors.add_event_handler(lambda x: print(x))
lines[3].colors = "magenta"

    returns:
    pick_info: {'index': range(0, 1000), 'collection-index': 3, 'world_object': <pygfx.Line at 0x7f00de72f8e0>, 'new_data': [array]}

lines[:].present.add_event_handler(plot.auto_scale)
lines[10:].present = False
lines[:5].present = False
lines[:].present = True
```


kushalkolar:
adding or removing a graphic from a collection needs to reset the indices.

In feature event pick_info we could also emit the parent graphic name. Or this could be handled in link, @clewis7 any idea on which is better

PR #93, opened 2022-12-28
Interactivity rebase

clewis7:
git hates me, feeling is mutual 

implements #39 

hopefully the last interactivity branch before it can be merged


clewis7:
okay, functionality is back to where it was before...can click on single contour and change color

should now be ready to finish interactivity

fastplotlib/graphics/_base.py
```diff
@@ -79,3 +80,222 @@ def __repr__(self):
...
+
+    @abstractmethod
+    def _reset_feature(self, feature: str):
+        pass
+
+    def link(self, event_type: str, target: Any, feature: str, new_data: Any, callback_function: callable = None):
+        if event_type in self.pygfx_events:
+            self.world_object.add_event_handler(self.event_handler, event_type)
+        elif event_type in self.feature_events:
+            feature = getattr(self, event_type)
```
kushalkolar, line 98: aren't the even types for features named like "colors-changed" and not "color". Maybe you can just change the event type to be like "color" instead of "color-changed"? For example see: 

https://github.com/kushalkolar/fastplotlib/blob/master/fastplotlib/graphics/features/_colors.py#L192

fastplotlib/graphics/_base.py
```diff
@@ -79,3 +80,222 @@ def __repr__(self):
...
+            for target_info in self.registered_callbacks[event.type]:
+                if target_info.callback_function is not None:
+                    # if callback_function is not None, then callback function should handle the entire event
+                    target_info.callback_function(source=self, target=target_info.target, event=event, new_data=target_info.new_data)
+                elif isinstance(self, GraphicCollection):
+                    indices = event.pick_info["collection_index"]
+                    target_info.target._set_feature(feature=target_info.feature, new_data=target_info.new_data, indices=indices)
+                else:
+                    target_info.target._set_feature(feature=target_info.feature, new_data=target_info.new_data,
+                                                    indices=None)
```
kushalkolar, line 122: for single Graphic the indices should be from `event.pick_info["index"]`

fastplotlib/graphics/line.py
```diff
@@ -83,3 +83,35 @@ def __init__(
 
         if z_position is not None:
             self.world_object.position.z = z_position
+
+    def _set_feature(self, feature: str, new_data: Any, indices: Any = None):
+        if not hasattr(self, "_previous_data"):
+            self._previous_data = {}
```
kushalkolar, line 89: better to create a dict or any iterable types explicitly like `dict()`, the `{}` can be buggy in a few very strange situations

`{}` is technically faster than `dict()`, but the difference is in nanoseconds

fastplotlib/graphics/line.py
```diff
@@ -83,3 +83,35 @@ def __init__(
...
+        elif hasattr(self, "_previous_data"):
+            self._reset_feature(feature)
+        # if feature in self._feature_events:
+        feature_instance = getattr(self, feature)
+        if indices is not None:
+            previous = feature_instance[indices].copy()
+            feature_instance[indices] = new_data
+        else:
+            previous = feature_instance[:].copy()
```
kushalkolar, line 98: this should be:

```python
previous = feature_instance().copy()
```

not all features are indexable (like present)

fastplotlib/graphics/line.py
```diff
@@ -83,3 +83,35 @@ def __init__(
...
+            feature_instance[:] = new_data
```
kushalkolar, line 99: similar, this should be like this which works for both indexable and non-indexable features. 

```python
feature_instance._set(new_data)
```

fastplotlib/graphics/line.py
```diff
@@ -83,3 +83,35 @@ def __init__(
...
+        else:
+            self._previous_data[feature] = PreviouslyModifiedData(previous_data=previous, previous_indices=indices)
+        # else:
+        #     raise ValueError("name arg is not a valid feature")
+
+    def _reset_feature(self, feature: str):
+        if feature not in self._previous_data.keys():
+            raise ValueError("no previous data registered for this feature")
+        else:
+            feature_instance = getattr(self, feature)
```
kushalkolar, line 112: more readable if you do:

```python
prev_ixs = self._previous_data[feature].previous_indices
if prev_ixs is not None:
    feature_instance[prev_ixs] = ...
```

fastplotlib/graphics/line.py
```diff
@@ -83,3 +83,35 @@ def __init__(
...
+            if self._previous_data[feature].previous_indices is not None:
+                feature_instance[self._previous_data[feature].previous_indices] = self._previous_data[
+                    feature].previous_data
+            else:
+                feature_instance[:] = self._previous_data[feature].previous_data
```
kushalkolar, line 117: similar to previous comment, use this which works for both indexable and non-indexable:

```python
feature_instance._set(...)
```

fastplotlib/graphics/linecollection.py
```diff
@@ -113,3 +117,35 @@ def __init__(
...
+
+    def _set_feature(self, feature: str, new_data: Any, indices: Union[int, range]):
+        if not hasattr(self, "_previous_data"):
+            self._previous_data = {}
+        elif hasattr(self, "_previous_data"):
+            self._reset_feature(feature)
+        #if feature in # need a way to check if feature is in
+        feature_instance = getattr(self, feature)
+        if indices is not None:
+            previous = feature_instance[indices].copy()
```
kushalkolar, line 129: so for collection it's going to be a bit diff, for keeping previous data just store the new data (which becomes previous data on the next event). This will become clear in my next comments

```python
# put this import at the top of the module
from copy import deepcopy

previous = deepcopy(new_data)
```

fastplotlib/graphics/linecollection.py
```diff
@@ -113,3 +117,35 @@ def __init__(
...
 
             self.add_graphic(lg, reset_index=False)
+
+    def _set_feature(self, feature: str, new_data: Any, indices: Union[int, range]):
+        if not hasattr(self, "_previous_data"):
+            self._previous_data = {}
+        elif hasattr(self, "_previous_data"):
+            self._reset_feature(feature)
+        #if feature in # need a way to check if feature is in
+        feature_instance = getattr(self, feature)
```
kushalkolar, line 127: this will not work for collections, see next comments

fastplotlib/graphics/linecollection.py
```diff
@@ -113,3 +117,35 @@ def __init__(
...
+        if indices is not None:
+            previous = feature_instance[indices].copy()
+            feature_instance[indices] = new_data
```
kushalkolar, line 130: for graphic collections the indices will refer to graphic indices, not vertex indices like for simple graphics.:

try:

```python
coll_feature = getattr(self[indices], feature)
coll_feature._set(new_data)
```

fastplotlib/graphics/linecollection.py
```diff
@@ -113,3 +117,35 @@ def __init__(
...
+        else:
+            previous = feature_instance[:].copy()
```
kushalkolar, line 132: see above comments, modify rest of the method accordingly

fastplotlib/graphics/linecollection.py
```diff
@@ -113,3 +117,35 @@ def __init__(
...
+        else:
+            self._previous_data[feature] = PreviouslyModifiedData(previous_data=previous, previous_indices=indices)
+        # else:
+        #     raise ValueError("name arg is not a valid feature")
+
+    def _reset_feature(self, feature: str):
+        if feature not in self._previous_data.keys():
+            raise ValueError("no previous data registered for this feature")
+        else:
+            feature_instance = getattr(self, feature)
```
kushalkolar, line 146: see comments from `_set_feature`

fastplotlib/graphics/_base.py
clewis7: wait, can you clarify? you want the event_type for feature events to be "colors" or "colors-changed"?

clewis7: for now I changed it to "colors", "data", etc. in order to continue getattr() but can change if you want

kushalkolar: sounds good, simpler and minimize the parsing is better.

clewis7:
update: about to start working on line collection but made the changes you suggested to base and line graphics

line graphic is single graphic so it normally wouldn't have indices, but I made sure that it would work if slice, int, etc. is passed into set_feature just in the event that it would happen

fastplotlib/graphics/line.py
```diff
@@ -8,6 +8,11 @@
 
 
 class LineGraphic(Graphic, Interaction):
+    feature_events = [
+        "data",
+        "colors",
+        "cmap",
```
kushalkolar, line 14: there is technically no `"cmap"` event, the CmapFeature is just a wrapper around ColorsFeature (i.e. it subclasses ColorFeature) and just parses cmap strings into an n_points, RGBA colors array based on the user-provided indexing.

PR #103, opened 2023-01-03
Plot streaming 1D/2D data in notebook

ArjunPutcha:
add demo nb to plot streaming data. One case of full replacement of static buffer, another case of partial replacement of static buffer with standard indexing. Next will be partial replacement with np.roll and collections.deque


kushalkolar:
Would be useful to see how this works with multiprocessing or shared memory: https://docs.python.org/3/library/multiprocessing.shared_memory.html

The shared memory is especially interesting for online analysis, can have a completely different process that is dedicated to data acquisition and/or analysis, and fastplotlib is running in the process that only does visualization. 

... I Just realized I'm gonna have to figure out how to get this working on windows because my microscope computer is windows :sob:

ArjunPutcha:
Would be very useful for my purposes, having a separate thread to handle acquisition. Will look into it! Just pulled onto windows, gonna test fpl on there 😅

Arjun
________________________________

ArjunPutcha:
For now, doesn't appear .add_animations() function works within process (processes don't reinstatiate automatically, placing add_animations() function within while loop doesn't work and is inherently inefficient). Need to look more into API to see if there's a way to run within process.

Multiprocessing can still be done for data processing, used SharedMemory object so add_animations simply access the shared_memory object and updates data

kushalkolar:
Should we close this and open a new one? I guess we can have something like a "use cases" dir in examples?

PR #104, opened 2023-01-07
reorganization of color utils

kushalkolar:
1. support qualitative colormaps
2. docstrings
3. cleanup
4. get_colors() -> make_colors()
5. get_colors_dict() -> make_colors_dict()


kushalkolar:
@EricThomson

In case you were using `get_colors()` (unlikely I assume) this is a breaking change, probably last one for the month :smile:

PR #114, opened 2023-01-12
better example nbs

kushalkolar:


kushalkolar:
I really don't think we need this.

PR #150, opened 2023-03-09
better buffer handling

kushalkolar:
- all feature buffers are now public
- `ImageGraphic` and `HeatmapGraphic` allow setting the input array buffer directly as the texture data buffer if `isolated_buffer=False`. The default is `True`, so by default the texture data buffer is a new array that is initialized using the shape and type of the input array, and then the isolated texture data buffer is set using the input data array values.

allows this type of behavior:
image

### DIFF
--- a/fastplotlib/graphics/image.py
+++ b/fastplotlib/graphics/image.py
@@ -2,11 +2,13 @@ from typing import *
 from math import ceil
 from itertools import product
 
+import numpy as np
 import pygfx
 from pygfx.utils import unpack_bitfield
 
 from ._base import Graphic, Interaction, PreviouslyModifiedData
 from .features import ImageCmapFeature, ImageDataFeature, HeatmapDataFeature, HeatmapCmapFeature
+from .features._base import to_gpu_supported_dtype
 from ..utils import quick_min_max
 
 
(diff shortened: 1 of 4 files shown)


fastplotlib/graphics/image.py
```diff
@@ -65,20 +72,29 @@ def __init__(
...
 
-        self.data = ImageDataFeature(self, data)
+        data = to_gpu_supported_dtype(data)
+
+        # TODO: we need to organize and do this better
+        if isolated_buffer:
+            # initialize a buffer with the same shape as the input data
+            # we do not directly use the input data array as the buffer
+            # because if the input array is a read-only type, such as
+            # numpy memmaps, we would not be able to change the image data
```
clewis7, line 82: so, the point of having the option for using an isolated buffer is to be able to change data in the event that the type of the data is a memmap instead of an array?

fastplotlib/graphics/image.py
```diff
@@ -65,20 +72,29 @@ def __init__(
...
+            buffer_init = np.zeros(shape=data.shape, dtype=data.dtype)
+        else:
+            buffer_init = data
 
         if (vmin is None) or (vmax is None):
             vmin, vmax = quick_min_max(data)
 
-        texture_view = pygfx.Texture(self.data(), dim=2).get_view(filter=filter)
+        texture_view = pygfx.Texture(buffer_init, dim=2).get_view(filter=filter)
```
clewis7, line 90: before, `Texture` and `Geometry` were all under the hood correct?

Issue #152, opened 2023-03-24
tests


clewis7:
- [ ] Graphics
    - [x] `Image`
        - [x] 2D 512x512 rand image which is pre-saved with cmaps
            - [x] changing cmap
        - [x] 2D RGB array, imageio example images
        - [x] changing vmin vmax
    - [x] `Line`
      - [x] sine, cosine, sinc from examples
          - [x] slicing colors, data
          - [x] present followed by auto_scale()
    - [x] `Scatter`
    - [x] `GridPlot`
    - [x] `ImageWidget`
    - [x] upload screenshot test images and diffs if tests failed: see https://github.com/pygfx/pygfx/blob/main/.github/workflows/ci.yml#L132 and https://github.com/actions/upload-artifact

kushalkolar:
You can basically go through our demos and edit the list here to add the remaining list of tests

PR #158, opened 2023-03-31
fpl tests

clewis7:
refactored so there are no large binary files

did reorg for names

#152


examples/gridplot/gridplot_cmap.py
kushalkolar, line 1: I guess we can just keep it to make sure accessing graphics within a gridplot works etc.

examples/imagewidget/iw_simple.py
```diff
@@ -0,0 +1,43 @@
...
+    data=data,
+    slider_dims=["t"],
+    vmin_vmax_sliders=True,
+    cmap="gnuplot2"
+)
+
+data = np.load("../data/random.npy")
+
+iw.show()
+
```
kushalkolar, line 33: can you test before and after setting timepoint?

set timepoint using the slider  as well as `@current_index`

examples/imagewidget/iw_simple.py
```diff
@@ -0,0 +1,43 @@
...
+iw["t"] = 250
```
kushalkolar, line 34: I don't think this is possible.

examples/imagewidget/iw_simple.py
```diff
@@ -0,0 +1,43 @@
...
+
+iw.plot.center_scene()
+
```
kushalkolar, line 37: also check that window functions and frame_apply works

examples/scatter/scatter.py
```diff
@@ -0,0 +1,49 @@
...
+
+# create a random distribution of 10,000 xyz coordinates
+n_points = 10_000
+
+# if you have a good GPU go for 1.5 million points :D
+# this is multiplied by 3
+n_points = 500_000
+
+# dimensions always have to be [n_points, xyz]
+dims = (n_points, 3)
```
kushalkolar, line 29: you don't need all these since you're loading existing data

examples/line/present.py
```diff
@@ -0,0 +1,48 @@
...
+from pygfx import WgpuRenderer
+
+canvas = WgpuCanvas()
+renderer = WgpuRenderer(canvas)
+
+plot = Plot(canvas=canvas, renderer=renderer)
+
+sine = np.load("../data/sine.npy")
+cosine = np.load("../data/cosine.npy")
+sinc = np.load("../data/sinc.npy")
```
kushalkolar, line 23: just make these, don't save them since it's not randomly generated

examples/line/line.py
```diff
@@ -0,0 +1,44 @@
...
+from pygfx import WgpuRenderer
+
+canvas = WgpuCanvas()
+renderer = WgpuRenderer(canvas)
+
+plot = Plot(canvas=canvas, renderer=renderer)
+
+sine = np.load("../data/sine.npy")
+cosine = np.load("../data/cosine.npy")
+sinc = np.load("../data/sinc.npy")
```
kushalkolar, line 23: just make these, don't save them since it's not randomly generated

examples/line/dataslice.py
```diff
@@ -0,0 +1,48 @@
...
+from pygfx import WgpuRenderer
+
+canvas = WgpuCanvas()
+renderer = WgpuRenderer(canvas)
+
+plot = Plot(canvas=canvas, renderer=renderer)
+
+sine = np.load("../data/sine.npy")
+cosine = np.load("../data/cosine.npy")
+sinc = np.load("../data/sinc.npy")
```
kushalkolar, line 23: just make these, don't save them since it's not randomly generated

examples/line/colorslice.py
```diff
@@ -0,0 +1,56 @@
...
+from pygfx import WgpuRenderer
+
+canvas = WgpuCanvas()
+renderer = WgpuRenderer(canvas)
+
+plot = Plot(canvas=canvas, renderer=renderer)
+
+sine = np.load("../data/sine.npy")
+cosine = np.load("../data/cosine.npy")
+sinc = np.load("../data/sinc.npy")
```
kushalkolar, line 23: just make these, don't save them since it's not randomly generated

examples/line/present.py
kushalkolar, line 1: this tests both present and scaling

kushalkolar: so you can just add that to the comment and filename

kushalkolar (changes requested):

kushalkolar:
it's not working just through pytest because:

1. you need an `__init__.py` in `examples/tests`
2. You need relative import for testutils, `from .testutils import` instead of `from testutils import`
3. you need relative paths to data files, for example:
```python
data_path = Path(__file__).parent.parent.joinpath("data", "random.npy")
data = np.load(data_path)
```

kushalkolar:
Seems like zenodo is still the best place to store these, ah well. The file will be small at least so download time should be very fast.

examples/line/present.py
clewis7: when I didn't save them was when I kept getting errors locally about the data changing...idk what the precision is like when you call `np.sin()` or `np.cos()` on a linspace but I think this is what was causing the problem before and that is why I saved them

kushalkolar: what if you force `float32` when making the linspace?

clewis7: will try

examples/imagewidget/iw_simple.py
clewis7: need to think some more about testing image widget, somehow need renderer and canvas to be offscreen as in the other examples

kushalkolar:
superseeded by #179

Issue #159, opened 2023-04-02
improve garbage collection


kushalkolar:
Currently garbage collection doesn't seem to be working properly, if a world object is removed from the scene and deleted the GPU VRAM is not freed. However it seems like if we remove objects from a scene and delete them just through `pygfx` the GPU VRAM is freed. Should probably use weakreferences in our Graphic classes.

Run this, and then after the image is removed from the scene if you click or move the mouse on the canvas it will free up the GPU VRAM.
```python
import numpy as np
from wgpu.gui.auto import WgpuCanvas, run
import pygfx as gfx
from weakref import
import subprocess

canvas = WgpuCanvas()
renderer = gfx.WgpuRenderer(canvas)
scene = gfx.Scene()
camera = gfx.OrthographicCamera(5000, 5000)
camera.position.x = 2048
camera.position.y = 2048


def make_image():
    data = np.random.rand(4096, 4096).astype(np.float32)

    return gfx.Image(
        gfx.Geometry(grid=gfx.Texture(data, dim=2)),
        gfx.ImageBasicMaterial(clim=(0, 1)),
    )

def draw():
    renderer.render(scene, camera)
    canvas.request_draw()


def print_nvidia(msg):
    print(msg)
    print(
        subprocess.check_output(["nvidia-smi", "--format=csv", "--query-gpu=memory.used"]).decode().split("\n")[1]
    )
    print()


def add_img(*args):
    print_nvidia("Before creating image")
    img = make_image()
    print_nvidia("After creating image")
    scene.add(img)
    img.add_event_handler(remove_img, "click")
    draw()
    print_nvidia("After add image to scene")


def remove_img(*args):
    img = scene.children[0]
    scene.remove(img)
    draw()
    print_nvidia("After remove image from scene")
    del img
    draw()
    canvas.ev
    print_nvidia("After del image")


renderer.add_event_handler(add_img, "double_click")

draw()
run()
```

kushalkolar:
done in #159

PR #160, opened 2023-04-03
garbage collection of WorldObjects 

kushalkolar:
Follows up from #159 

This implements:

- A global `WORLD_OBJECTS` dict in `fastplotlib.graphics._base` used internally by `fastplotlib` to keep track of all pygfx `WorldObject` instances in a session. `Graphic` instances only use a weak reference proxy to access their `WorldObject` using their hex id. All references to world objects outside of this dict are via proxies.
- A global `GRAPHICS` dict in `fastplotlib.layouts._base` to keep track of all `Graphic` objects, functions similar to the global world objects dict. All references to Graphic objects outside of this dict are via proxies. The various `plot.add_<graphic>` methods only return proxies.

Adds a `PlotArea.delete_graphic()` method which is the only way that a `Graphic` should be deleted. This also frees up GPU VRAM.

GPU VRAM freed and tested for:
- [x] Image
- [x] Line
- [x] LineCollection
- [x] Scatter
- ~[ ] Text~ will do this later once text is more stabilized
- [x] Heatmap (new one)

- [x] Modify how `LineSlider is added in `PlotArea` but don't need to check garbage collection since their memory footprint is very small. UPDATE: This isn't necessary, `LineSlider` continues to work.

RAM freed and tested for:
- [x] Image
- [x] Line
- [x] LineCollection
- [x] Scatter
- [x] Heatmap (new one)

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -50,7 +50,8 @@ class BaseGraphic:
 
 class Graphic(BaseGraphic):
     def __init__(
-            self, name: str = None):
+            self, name: str = None
+    ):
         """
 
         Parameters
@@ -64,15 +65,16 @@ class Graphic(BaseGraphic):
         self.registered_callbacks = dict()
         self.present = PresentFeature(parent=self)
 
-        self._world_objects = WORLD_OBJECTS
+        # store hex id str of Graphic instance mem location
+        self.loc: str = hex(id(self))
 
     @property
     def world_object(self) -> WorldObject:
         """Associated pygfx WorldObject. Always returns a proxy, real object cannot be accessed directly."""
-        return weakref.proxy(self._world_objects[hex(id(self))])
+        return weakref.proxy(WORLD_OBJECTS[hex(id(self))])
 
     def _set_world_object(self, wo: WorldObject):
-        self._world_objects[hex(id(self))] = wo
+        WORLD_OBJECTS[hex(id(self))] = wo
 
     @property
     def position(self) -> Vector3:
@@ -112,7 +114,7 @@ class Graphic(BaseGraphic):
             return rval
 
     def __del__(self):
-        del self._world_objects[hex(id(self))]
+        del WORLD_OBJECTS[self.loc]
 
 
 class Interaction(ABC):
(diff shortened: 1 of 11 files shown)


kushalkolar:
GPU VRAM is freed by deleting the `WorldObject`, we probably need to make weak references to the various feature data to allow RAM to be freed when those are deleted. Maybe by using a similar dict to keep data, `COLOR_FEATURE`, `DATA_FEATURE`.

kushalkolar:
GPU VRAM get freed perfectly.

System RAM gets cleared to an extent, it doesn't keep growing which is the most important thing. There must be a lingering reference because we create data array in jupyter, and this is the data that we feed to `add_<graphic>`. But since it doesn't keep growing I think it's fine for now.

fastplotlib/graphics/_base.py
```diff
@@ -264,22 +285,21 @@ class PreviouslyModifiedData:
...
-        return self._world_object
+        self._graphics: List[str] = list()
 
     @property
     def graphics(self) -> Tuple[Graphic]:
-        """returns the Graphics within this collection"""
-        return tuple(self._graphics)
+        """The Graphics within this collection. Always returns a proxy to the Graphics."""
+        proxies = [weakref.proxy(COLLECTION_GRAPHICS[loc]) for loc in self._graphics]
+        return tuple(proxies)
```
clewis7, line 302: `self._graphics` houses all of the locs for each graphic in a collection, so then when a user accesses `self.graphics` it will return a weak proxy to each of the graphics that are being indexed for?

fastplotlib/graphics/image.py
```diff
@@ -272,7 +274,8 @@ def __init__(
         start_ixs = [list(map(lambda c: c * chunk_size, chunk)) for chunk in chunks]
         stop_ixs = [list(map(lambda c: c + chunk_size, chunk)) for chunk in start_ixs]
 
-        self._world_object = pygfx.Group()
+        world_object = pygfx.Group()
```
clewis7, line 277: why do we want to expose the `world_object` now when it was always private before?

fastplotlib/graphics/_base.py
```diff
@@ -100,6 +113,9 @@ def __repr__(self):
         else:
             return rval
 
+    def __del__(self):
+        del WORLD_OBJECTS[self.loc]
```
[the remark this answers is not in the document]
kushalkolar: yup

fastplotlib/graphics/_base.py
kushalkolar: yup!

kushalkolar: would probably be easier to read by unfurling the list comprehension

PR #163, opened 2023-04-14
catch up with pygfx refactor of of cameras, controllers, and Texture

kushalkolar:
This implements latest changes from pygfx for the cameras, controllers, and Texture refactor:

- https://github.com/pygfx/pygfx/pull/457
- https://github.com/pygfx/pygfx/pull/490
- https://github.com/pygfx/pygfx/pull/493

This simplified a lot of things in fastplotlib: 
1. `PlotArea._refresh_cameras()` is gone
2. `PlotArea` `center_graphic()`, `center_scene()`, `auto_scale()`, simplified
3. Use `Texture` directly instead of `Texture.get_view()` in `ImageGraphic` and `HeatmapGraphic`

The only thing that is a regression is that it's not possible to use the same controller on two cameras with different states, so the more complex Scatter and Line examples where the cameras are flipped on some subplots doesn't work anymore, using the controllers for these plots will make all the cameras have the same state as the first camera added to the controller. Minor feature regression but I don't think anyone was using cameras in this way anyways.

Other than the above regression all examples work, our big epic visualization with calcium & behavior data also works so this is good to go!


kushalkolar:
will need to update `setup.py` to require latest pygfx when they do a new release

ivoflipse:
Can you wrap multiple Controllers and then pass the wrappers events to the separate Controllers?

kushalkolar:
> Can you wrap multiple Controllers and then pass the wrappers events to the separate Controllers?

You mean one controller for multiple viewports (that's how it is currently), and then pass events separately to the cameras?

ivoflipse:
You said:

> The only thing that is a regression is that it's not possible to use the same controller on two cameras with different states,

I think we intend the same thing, but I'm suggesting a 1-1 relation between Controller and Camera. But having 1 Controller control 2 other Controller. I think you're suggesting to have 1 Controller control 2 or more Cameras

Either way, it was more philosophical, I'm sure you'll find a way to adapt 😉

kushalkolar:
Ah, so it's now written that a controller can control a controller?

Anyways this is a very minor regression and we probably won't fix it until
we get a use case.

Thanks fit the comments!

ivoflipse:
Well I was suggesting that you wrap it with a Controller that controls Controllers 😉

But I think you're wise to wait until you actually have a need for it

PR #178, opened 2023-04-19
basic recording for Plot

### LINKED ISSUE #171: record plot to video file
maybe a simple `PlotArea.record_start()` and `PlotArea.record_stop()`. Saves frames during `PlotArea.render()`

kushalkolar:
closes #171 

usage:

```python
# make plot and stuff

plot.record_start("./video.mp4")

# ... do stuff, interact, etc

plot.record_stop()
```

codec options:

```python3
plot_movie.record_start("./test_265.mp4", codec="libx264", options={"crf": "20"})
```

Uses a separate `Process` for writing frames using [`PyAV`](https://github.com/PyAV-Org/PyAV), snapshot of canvas is sent using `Queue`. It should be fast enough for anything that's not too crazy. Right now the framerate is variable since we just grab frames in an animation function.

Tried OpenCV initially but it was acting really weird. It would write a video file just fine, but then when you started writing the next video file it made the previous one unreadable. `VideoCapture.relese()` was called, but there was something weird with multiprocessing I think. PyAV doesn't have this issue and seems to be better anyways with codecs etc.


fastplotlib/layouts/_record_mixin.py
```diff
@@ -0,0 +1,238 @@
...
+            if frame is None:
+                self.container.close()
+                break
+
+            frame = av.VideoFrame.from_ndarray(
+                frame[:self.stream.height, :self.stream.width],  # trim if necessary because of x264
+                format="rgb24"
+            )
+
+            for packet in self.stream.encode(frame):
```
clewis7, line 63: what are these lines doing? what is a packet?

clewis7: also `.mux()` function?

fastplotlib/layouts/_record_mixin.py
```diff
@@ -0,0 +1,238 @@
...
+            return
+
+        # reset timer
+        self._record_timer = t
+
+        if self._video_writer is not None:
+            ss = self.canvas.snapshot()
+            # exclude alpha channel
+            self._video_writer_queue.put(ss.data[..., :-1])
+
```
clewis7, line 105: what indexing does this do on the snapshot? 

`ss.data[..., :-1]`

clewis7 (commented):
lgtm, I see what you mean with the mixin class now

clewis7:
just out of curiosity, what happens when you change the size during the recording?

fastplotlib/layouts/_record_mixin.py
kushalkolar: I don't know exactly, I copied from the pyav docs example for writing video to file. It works so.

fastplotlib/layouts/_record_mixin.py
kushalkolar: `...` is to skip dimensions, so this way we only index the last dimension (RGBA) and we then only take the RGB, discard alpha channel.

Equivalent to `[:, :, - 1]`

kushalkolar:
The video width and height are fixed, so if you change the canvas size the width and height will change. I think it then would get trimmed.

Issue #149, opened 2023-03-09
Toolbar


kushalkolar:
similar to bokeh's tool buttons, automatically append this to the bottom of any plot, unless `tool_buttons=False` or something. Could use icons for it, can get free icons here: thenounproject.com

For all plot areas:
- [x] auto scale
- [x] center scene
- [x] checkable button for maintain aspect
- [x] enable/disable pan zoom controller, once this is merged: https://github.com/pygfx/pygfx/pull/457
- [x] flip `camera.scale.y`, https://fontawesome.com/icons/rotate?f=classic&s=solid
- [x] record start & stop once this is merged: #178 
  - just save in the nb local path for now, use timestamp for filename
- ~add selectors (once they are ready)~ adding selectors with buttons doesn't make much sense, can wait until we have right-click menus for those since those are really specific to a parent graphic.
- [x]  tooltip for all buttons 

ImageWidget specific:
- [x] reset vmin vmax - for `ImageWidget`, can use a contrast icon
- [x] Play button, try out the ipywidgets play, stop, pause etc. stuff again
- [x] step size, `BoundedIntText` widget

kushalkolar:
There's a boolean attribute or property `enabled` for controllers

kushalkolar:
What do you think about buttons for adding selectors. For example, click a " add `LinearSelector`" button to toggle it, then click on the graphic (line, line collection, or heatmap) to add the selector to, the selector gets added and button gets un-toggled.

kushalkolar:
done in #195 , #207

PR #195, opened 2023-05-17
working on toolbar

clewis7:
#149 

started working on creating a toolbar similar to bokeh's 

used thenounproject.com to get free icons for buttons

have buttons in place in ipywidget HBox and then put plot and toolbar in VBox 

just playing around in notebooks for now, need to figure out event system before incorporating into plots


clewis7:
current thought is to just use default bool for tool_buttons and when plot.show is called if True will render plot in VBox with toolbar

kushalkolar:
Do these images appear as buttons? If not is there a way to set an image inside a ipywidget button?

clewis7:
I tried to make the buttons as images and was unsuccessful but then I found a library called ipyevents that allows you to make the images clickable but was going to keep playing around with making the button widget as images

kushalkolar:
Seems like you can choose icons from the "FontAwesome" collection and set them on ipywidget buttons with the `icon=` arg!

https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20List.html#togglebutton
https://fontawesome.com/icons

kushalkolar:
I think the best places to implement would be `Plot.show()` and `GridPlot.show()` . And make and return them only if it's detected to be running within jupyter. 

One caveat with GridPlot is that we often want to apply these options (like not maintain aspect) to only certain subplots. Maybe we can have a dropdown menu for `GridPlot` which is filled with the subplot positions or names (if exists). Clicking within a subplot's scene also activates it in the dropdown menu.

Ideally it would be nice to show these buttons within the canvas, but that is more involved: https://github.com/pygfx/pygfx/issues/519 

what do you think about this for center-scene? https://fontawesome.com/icons/arrows-to-eye?f=classic&s=solid

clewis7:
unfortunately ipywidgets is only current with fontawesome buttons from version 5.15.4, so the arrow-to-eye icon is not available

clewis7:
I think for now I will just use VBox and have the toolbar below the plot or gridplot and then when everything is functional I can look into doing canvas UI components

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
 
-        return self.canvas
+        if self.toolbar:
+            tools = ToolBar(self).toolbar
+            return VBox([self.canvas, tools])
+        else:
+            return self.canvas
+
+
+class ToolBar:
```
kushalkolar, line 117: Make a separate module for this class, something like "_toolbar.py". You probably want a separate class for the Plot and Gridplot ones.

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+                self.plot.controller.enabled = False
+            else:
+                self.plot.controller.enabled = True
+
+        panzoom_tool.observe(panzoom_control, 'value')
+        auto_tool.on_click(auto_scale)
+        center_tool.on_click(center_scene)
+
+    @property
+    def toolbar(self):
```
kushalkolar, line 159: This isn't a good way to do it since it's creating and HBox every time it's called. Make the HBox, called it "widget" or something.

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+
+        Parameters
+        ----------
+        plot:
+        """
+        self.plot = plot
+
+        self._tools = list()
+
+        auto_tool = Button(value=False, disabled=False, icon='expand-arrows-alt', layout=Layout(width='auto'))
```
kushalkolar, line 131: Make these attributes, and more informative names, like "auto scale button"

kushalkolar, line 131: Set tooltips for all buttons

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
         if autoscale:
             self.auto_scale(maintain_aspect=True, zoom=0.95)
 
-        return self.canvas
+        if self.toolbar:
```
kushalkolar, line 110: Just make this a kwarg in show() instead of the constructor. Create a new toolbar widget instance in show() it it's not instantiated, else use existing instance.

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+        plot:
+        """
+        self.plot = plot
+
+        self._tools = list()
+
+        auto_tool = Button(value=False, disabled=False, icon='expand-arrows-alt', layout=Layout(width='auto'))
+        center_tool = Button(value=False, disabled=False, icon='compress-arrows-alt', layout=Layout(width='auto'))
+        panzoom_tool = ToggleButton(value=False, disabled=False, icon='hand-pointer', layout=Layout(width='auto'))
+        maintain_aspect_tool = ToggleButton(value=False, disabled=False, description="1:1", layout=Layout(width='auto'))
```
kushalkolar, line 134: Default value should be True not False

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+        ----------
+        plot:
+        """
+        self.plot = plot
+
+        self._tools = list()
+
+        auto_tool = Button(value=False, disabled=False, icon='expand-arrows-alt', layout=Layout(width='auto'))
+        center_tool = Button(value=False, disabled=False, icon='compress-arrows-alt', layout=Layout(width='auto'))
+        panzoom_tool = ToggleButton(value=False, disabled=False, icon='hand-pointer', layout=Layout(width='auto'))
```
kushalkolar, line 133: Default value should be True

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+        maintain_aspect_tool = ToggleButton(value=False, disabled=False, description="1:1", layout=Layout(width='auto'))
+        maintain_aspect_tool.style.font_weight = "bold"
+        self._tools.extend([auto_tool, center_tool, panzoom_tool, maintain_aspect_tool])
```
kushalkolar, line 136: Not necessary, just make the HBox instance here and call it widget

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+                self.plot.auto_scale(maintain_aspect=True)
+            else:
+                self.plot.auto_scale()
+
+        def center_scene(obj):
+            self.plot.center_scene()
+
+        def panzoom_control(obj):
+            if panzoom_tool.value:
+                # toggle pan zoom controller
```
kushalkolar, line 149: I think you cal just get the bool value and set it directly, this if else control flow is unnecessary

fastplotlib/plot.py
```diff
@@ -103,4 +107,54 @@ def show(self, autoscale: bool = True):
...
+                self.plot.controller.enabled = False
+            else:
+                self.plot.controller.enabled = True
+
+        panzoom_tool.observe(panzoom_control, 'value')
+        auto_tool.on_click(auto_scale)
+        center_tool.on_click(center_scene)
+
```
kushalkolar, line 157: Missing maintain aspect

kushalkolar (changes requested):

clewis7:
Running into small issue, want to get your thoughts...

` You can specify the enumeration of selectable options by passing a list (options are either (label, value) pairs, or simply values for which the labels are derived by calling str).`

So gp tool bar is working when subplots have names or when there is combination of names/no names for subplots but I am running into an issue when all the subplots have no names due to above

When I try to input the options for the dropdown widget as tuples corresponding to plot position that can later be used as indexing for toolbar events, I think the widget thinks I am passing (label, value) pairs and therefore only the label is being put into the option instead of the entire subplot position :(

But I am also confused because it does fine when I am having a mix of subplots with names and no names

fastplotlib/plot.py
clewis7: I had it like that initially but kept running into circular import issues because `ToolBar` will encapsulate a `Plot` instance but then `Plot.show()` will need `ToolBar` instance

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,3 +288,61 @@ def __next__(self) -> Subplot:
 
     def __repr__(self):
         return f"fastplotlib.{self.__class__.__name__} @ {hex(id(self))}\n"
+
+
+class GridPlotToolBar:
```
kushalkolar, line 293: Separate module and see my comments on Plot

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,3 +288,61 @@ def __next__(self) -> Subplot:
...
+        self._tools = list()
+
+        auto_tool = Button(value=False, disabled=False, icon='expand-arrows-alt', layout=Layout(width='auto'))
+        center_tool = Button(value=False, disabled=False, icon='compress-arrows-alt', layout=Layout(width='auto'))
+        panzoom_tool = ToggleButton(value=False, disabled=False, icon='hand-pointer', layout=Layout(width='auto'))
+        maintain_aspect_tool = ToggleButton(value=False, disabled=False, description="1:1", layout=Layout(width='auto'))
+        maintain_aspect_tool.style.font_weight = "bold"
+        self._tools.extend([auto_tool, center_tool, panzoom_tool, maintain_aspect_tool])
+
+        positions = [[i, j] for i, j in self.plot._get_iterator()]
```
kushalkolar, line 314: Use the shape attribute of Gridplot and itertools.product instead.

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,3 +288,61 @@ def __next__(self) -> Subplot:
...
+        values = list()
+        for pos in positions:
+            if self.plot[pos].name is not None:
+                values.append(self.plot[pos].name)
+            else:
+                values.append(tuple(pos))
```
kushalkolar, line 320: Change this into a string, like "(0, 0)" for numerical positions

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,3 +288,61 @@ def __next__(self) -> Subplot:
...
+        self._dropdown = Dropdown(options=values, disabled=False, description='Plots:')
```
kushalkolar, line 321: Subplots for description

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,3 +288,61 @@ def __next__(self) -> Subplot:
...
+
+        def auto_scale(obj):
+            current = self._dropdown.value
```
kushalkolar, line 324: Make a property that parses the Subplot and determines if its str or numerical. If it starts with "(" we can assume its numeric for now. This property returns the Subplot instance.

kushalkolar (changes requested):

fastplotlib/plot.py
kushalkolar: Ah right nevermind

fastplotlib/plot.py
clewis7: maintain aspect is only for when `plot.auto_scale(maintain_aspect=True)` is called, so I just have a check happening when the `autoscale_button` is being clicked

kushalkolar: it's a property of camera that can be set.

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
         if autoscale:
             self.auto_scale(maintain_aspect=True, zoom=0.95)
 
-        return self.canvas
+        if toolbar:
+            tools = ToolBar(self).toolbar
```
kushalkolar, line 108: Make the toolbar an attribute of Plot so its only instantiated once. This would create new toolbars every time show is called which gets messy

fastplotlib/plot.py
kushalkolar: No this property is unnecessary, just use widget as a public attribute

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
         if autoscale:
             self.auto_scale(maintain_aspect=True, zoom=0.95)
 
-        return self.canvas
+        if toolbar:
```
kushalkolar, line 107: Also verify its running in jupyter

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
...
+                                                   tooltip='Maintain camera aspect ratio for all dims')
+        self.maintain_aspect_button.style.font_weight = "bold"
+
+        self._widget = HBox([self.autoscale_button,
+                             self.center_scene_button,
+                             self.panzoom_controller_button,
+                             self.maintain_aspect_button])
+
+        def auto_scale(obj):
+            if self.maintain_aspect_button.value:
```
kushalkolar, line 143: Use the camera's value, not the button

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
...
+                                                   layout=Layout(width='auto'),
+                                                   tooltip='Maintain camera aspect ratio for all dims')
+        self.maintain_aspect_button.style.font_weight = "bold"
+
+        self._widget = HBox([self.autoscale_button,
+                             self.center_scene_button,
+                             self.panzoom_controller_button,
+                             self.maintain_aspect_button])
+
+        def auto_scale(obj):
```
kushalkolar, line 142: Make all of these regular methods. I don't like nesting functions unless it's necessary

kushalkolar (changes requested):

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
...
+
+        Parameters
+        ----------
+        plot: encapsulated plot instance that will be manipulated using the toolbar buttons
+        """
+        self.plot = plot
+
+        self.autoscale_button = Button(value=False, disabled=False, icon='expand-arrows-alt',
+                                       layout=Layout(width='auto'), tooltip='Auto-scale the camera w.r.t to the scene')
+        self.center_scene_button = Button(value=False, disabled=False, icon='compress-arrows-alt',
```
kushalkolar, line 128: what about this for center instead? https://fontawesome.com/icons/align-center?f=classic&s=solid

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
...
+        Basic toolbar for a Plot instance.
+
+        Parameters
+        ----------
+        plot: encapsulated plot instance that will be manipulated using the toolbar buttons
+        """
+        self.plot = plot
+
+        self.autoscale_button = Button(value=False, disabled=False, icon='expand-arrows-alt',
+                                       layout=Layout(width='auto'), tooltip='Auto-scale the camera w.r.t to the scene')
```
kushalkolar, line 127: simpler tool tips, just "auto-scale scene" is enough, people don't need to be thinking about cameras

fastplotlib/plot.py
```diff
@@ -103,4 +104,57 @@ def show(self, autoscale: bool = True):
...
+        self.center_scene_button = Button(value=False, disabled=False, icon='compress-arrows-alt',
+                                          layout=Layout(width='auto'), tooltip='Auto-center the scene, does not scale')
+        self.panzoom_controller_button = ToggleButton(value=True, disabled=False, icon='hand-pointer',
+                                                      layout=Layout(width='auto'), tooltip='Toggle panzoom controller')
+        self.maintain_aspect_button = ToggleButton(value=True, disabled=False, description="1:1",
+                                                   layout=Layout(width='auto'),
+                                                   tooltip='Maintain camera aspect ratio for all dims')
```
kushalkolar, line 134: simpler tooltip, "maintain aspect"

kushalkolar (changes requested):

fastplotlib/plot.py
kushalkolar: you also don't need the if else control flow here

kushalkolar (changes requested):

fastplotlib/layouts/_gridplot.py
```diff
@@ -294,3 +308,94 @@ def __next__(self) -> Subplot:
...
+    def click_handler(self, ev):
+        for subplot in self.plot:
+            pos = subplot.map_screen_to_world((ev.x, ev.y))
+            if pos is not None:
+                # update self.dropdown
+                if subplot.name is None:
+                    self.dropdown.value = str(subplot.position)
+                else:
+                    self.dropdown.value = subplot.name
+                subplot.controller.enabled = self.panzoom_controller_button.value
```
kushalkolar, line 399: You want it the other way around, click a Subplot sets the buttons from the Subplot values

fastplotlib/layouts/_gridplot.py
```diff
@@ -294,3 +308,94 @@ def __next__(self) -> Subplot:
...
+
+    def maintain_aspect(self, obj):
+        current = self.parser
+        current.camera.maintain_aspect = self.maintain_aspect_button.value
+
+    def flip_camera(self, obj):
+        current = self.parser
+        current.camera.scale.y = -1 * current.camera.scale.y
+
+    def click_handler(self, ev):
```
kushalkolar, line 390: Better relevant name

fastplotlib/layouts/_gridplot.py
```diff
@@ -294,3 +308,94 @@ def __next__(self) -> Subplot:
...
+        self.panzoom_controller_button.observe(self.panzoom_control, 'value')
+        self.autoscale_button.on_click(self.auto_scale)
+        self.center_scene_button.on_click(self.center_scene)
+        self.maintain_aspect_button.observe(self.maintain_aspect, 'value')
+        self.flip_camera_button.on_click(self.flip_camera)
+
+        self.plot.renderer.add_event_handler(self.click_handler, "click")
+
+    @property
+    def parser(self) -> Subplot:
```
kushalkolar, line 362: Better name, this bacially gives current Subplot

kushalkolar (changes requested):

kushalkolar (approved):

PR #199, opened 2023-05-21
more selector stuff

### LINKED ISSUE #186: Need garbage collection of selectors
Otherwise they will pollute `PlotArea._selectors`, it's best to get rid of them. This is especially the case with `mesmerize-viz` for the cnmf viz, it's better to add a new `LineSelector` each time.

This sorta depends on #172 so they can easily be managed by `PlotArea`

### LINKED ISSUE #201: Drag transparent selectors
If alpha of fill is less than 1, add and event handler on the renderer and check if the point_down event was within the bbox of the fill mesh.

kushalkolar:
closes #186, closes #201

### DIFF
--- a/fastplotlib/graphics/image.py
+++ b/fastplotlib/graphics/image.py
@@ -14,16 +14,16 @@ from ..utils import quick_min_max
 
 
 class _ImageHeatmapSelectorsMixin:
-    def add_linear_selector(self, selection: int = None, padding: float = 50, **kwargs) -> LinearSelector:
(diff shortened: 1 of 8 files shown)


kushalkolar:
Probably possible to simplify the way `bounds` are used by just using `world_object.get_world_bounding_box()` instead of calculating it ourself using offsets etc. Just set origin to be `(0, 0)` or `parent.position` as default. Then use `get_world_bounding_box()` to get x & y limits and then get all geometry data within these bounds.

kushalkolar:
I'm going to merge this now so I can implement pylinalg refactor, will continue in another PR

PR #203, opened 2023-05-22
catchup with pygfx linalg refactor

kushalkolar:


kushalkolar:
everything works except rectangle region selector, will do that later

PR #205, opened 2023-05-24
fast line collection kinda works with color feature

kushalkolar:
https://github.com/kushalkolar/fastplotlib/assets/9403332/de2679b3-8a9d-4ad4-b5b8-3dc3349699ec


kushalkolar:
This is no longer necessary after #214 and #215, but it might be useful in the future where a `Graphic` can have access to a slice of a buffer instead of an entire buffer.

PR #207, opened 2023-05-24
finish buttons for toolbar

clewis7:
finishes #149 

added record button, need to make image widget toolbar

PR #209, opened 2023-05-25
selector garbage collection, use `selection` feature name, cleanup event handlers

clewis7:


kushalkolar:
let's also throw in #211

fastplotlib/graphics/features/_base.py
```diff
@@ -254,6 +259,9 @@ def cleanup_array_slice(key: np.ndarray, upper_bound) -> np.ndarray:
     if key.dtype == bool:
         key = np.nonzero(key)[0]
 
+    if key.size < 1:
```
kushalkolar, line 262: I forgot what this fixed :smile:

fastplotlib/graphics/selectors/_linear_region.py
```diff
@@ -11,6 +11,9 @@
 
 
 class LinearBoundsFeature(GraphicFeature):
+    feature_events = (
+        "data",
```
kushalkolar, line 15: this is supposed to be bounds, but as for #211 let's rename it to selection

fastplotlib/graphics/features/_base.py
kushalkolar: @clewis7 can you jog my memory? Is this when you do something like `line_graphic.colors[xs > 0] = "w"`, but `xs > 0` is an empty array?

clewis7: yes, bc we kept getting an error when trying to set the colors and a behavior had not a occurred so the array was all zeroes, would never give True

kushalkolar:
this got included in #207

Issue #214, opened 2023-06-06
probably get rid of `CollectionIndexer` and `CollectionFeature` in favor of numpy arrays for graphic collections


kushalkolar:
Right now `CollectionIndexer` and `CollectionFeature` are used for fancy indexing of Graphics in a `GraphicCollection`, such as `LineCollection`. However I have found that this indexing, since we're using lists and for loops a lot, becomes very slow with thousands of world objects.

```python
# very slow, 16 seconds to change 1,000
 for g in contours[ixs_visible].graphics:
     if not g.visible:
         g.visible = True
        
 for g in contours[ixs_hide].graphics:
     if g.visible:
         g.visible = False
```

However if we just use a numpy array of graphics instead of a tuple then it is MUCH faster:

```python
# very fast, 10 ms to change 1,000 on an RX 470
for g in contours.graphics[ixs_visible]:
    if not g.visible:
        g.visible = True
    
for g in contours.graphics[ixs_hide]:
    if g.visible:
        g.visible = False
```

We can set the array as read-only by setting the writeable flag: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html#numpy.ndarray.flags

Could create a new class that uses numpy array of Graphics to parse feature changes etc., basically have `CollectionIndexer` use the numpy array of graphics.

kushalkolar:
So it seems like keeping `CollectionIndexer` is fine if we use a numpy array for graphics

Fast:
```python
t1 = time()
line_stack[::2].colors = "magenta"
print(time() - t1)
```

But then this is still slow:
```python
t1 = time()
line_stack[::2].colors[::2] = "cyan"
print(time() - t1)
```

PR #215, opened 2023-06-06
use numpy array instead of tuple for indexing collections

kushalkolar:
starts #214


kushalkolar:
merging for now, will continue optimizing later

Issue #190, opened 2023-05-10
Add set_data() to ImageWidget 


kushalkolar:
This would change the current data arrays in the image widget. For example if input and mcorr movies are shown for a batch item, "set_data([new_input, new_mcorr])" would change the data arrays shown in the viewer. Useful with DataGrid to display DataFrames and select rows. 

Allow this to work as long as:

1. The number of data arrays is the same as the current number of data arrays, this corresponds to the Gridplot shape.
2. The ndim of the new data arrays matches

Allow set data to take kwargs such as:
- `reset_frame_index: bool` , which sets `@current_frame` index back to `0` if `True`
- `reset_vmin_vmax`, `True` by default. 

Also force setting the graphic data for each subplot to make sure they update the current frame.

kushalkolar:
done in #216

PR #216, opened 2023-06-07
set_data()

clewis7:
#190 

add public method to ImageWidget to change the data

### DIFF
--- a/fastplotlib/widgets/image.py
+++ b/fastplotlib/widgets/image.py
@@ -1,3 +1,4 @@
+import itertools
 import traceback
 from datetime import datetime
 from itertools import product
@@ -854,6 +855,30 @@ class ImageWidget:
             }
 
             self.vmin_vmax_sliders[i].set_state(state)
+
+    def set_data(self, new_data: Union[np.ndarray, List[np.ndarray]], reset_vmin_vmax: bool = True, reset_current_index: bool = True):
+        """Change data of widget"""
+        if reset_current_index:
+            self.current_index["t"] = 0
+            self.sliders["t"].value = 0
+
+        if isinstance(new_data, np.ndarray):
+            if new_data.shape != self._data[0].shape:
+                raise ValueError(f"new data shape {new_data.shape} does not equal current data shape {self._data[0].shape}")
+            self._data[0] = new_data
+            self.plot.graphics[0].data = self._data[0][self.current_index["t"]]
+        else:
+            if len(self._data) != len(new_data):
+                raise ValueError(f"number of new data arrays {len(new_data)} must match current number of data arrays {len(self._data)}")
+            positions = list(product(range(self.plot.shape[0]), range(self.plot.shape[1])))
+            for i, (new_darray, darray) in enumerate(zip(new_data, self._data)):
+                if new_darray.shape != darray.shape:
+                    raise ValueError(f"new data shape {new_darray.shape} does not equal current data shape {darray.shape}")
+                self._data[i] = new_data[i]
+                self.plot[positions[i][0], positions[i][1]].graphics[0].data = self._data[i][self.current_index["t"]]
+
+        if reset_vmin_vmax and len(self.vmin_vmax_sliders) != 0:
+            self.reset_vmin_vmax()
 
     def show(self, toolbar: bool = True):
         """
--- a/fastplotlib/widgets/image.py
+++ b/fastplotlib/widgets/image.py
@@ -862,7 +862,7 @@ class ImageWidget:
             self.current_index["t"] = 0
             self.sliders["t"].value = 0
 
-        if isinstance(new_data, np.ndarray):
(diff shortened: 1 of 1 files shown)


fastplotlib/widgets/image.py
```diff
@@ -855,6 +856,30 @@ def reset_vmin_vmax(self):
 
             self.vmin_vmax_sliders[i].set_state(state)
 
+    def set_data(self, new_data: Union[np.ndarray, List[np.ndarray]], reset_vmin_vmax: bool = True, reset_current_index: bool = True):
+        """Change data of widget"""
+        if reset_current_index:
```
kushalkolar, line 861: change to `reset_indices`, which resets indices for all dimensions

fastplotlib/widgets/image.py
```diff
@@ -855,6 +856,30 @@ def reset_vmin_vmax(self):
...
+
+        if isinstance(new_data, np.ndarray) and isinstance(self.plot, Plot):
+            if new_data.shape != self._data[0].shape:
+                raise ValueError(f"new data shape {new_data.shape} does not equal current data shape {self._data[0].shape}")
+            self._data[0] = new_data
+            self.plot.graphics[0].data = self._data[0][self.current_index["t"]]
+        else:
+            if len(self._data) != len(new_data):
+                raise ValueError(f"number of new data arrays {len(new_data)} must match current number of data arrays {len(self._data)}")
+            positions = list(product(range(self.plot.shape[0]), range(self.plot.shape[1])))
```
kushalkolar, line 873: this is unecessary, `GridPlot` is iterable :smile:

fastplotlib/widgets/image.py
```diff
@@ -855,6 +856,30 @@ def reset_vmin_vmax(self):
...
+            for i, (new_darray, darray) in enumerate(zip(new_data, self._data)):
```
kushalkolar, line 874: just add the image widget's gridplot instance to the `zip`

fastplotlib/widgets/image.py
kushalkolar:
```python
for subplot in gridplot:
  subplot#
```

fastplotlib/widgets/image.py
```diff
@@ -855,6 +856,30 @@ def reset_vmin_vmax(self):
...
+                if new_darray.shape != darray.shape:
+                    raise ValueError(f"new data shape {new_darray.shape} does not equal current data shape {darray.shape}")
+                self._data[i] = new_data[i]
+                self.plot[positions[i][0], positions[i][1]].graphics[0].data = self._data[i][self.current_index["t"]]
```
kushalkolar, line 878: add a comment that we assume `graphics[0]` is always the "image widget manged data"

fastplotlib/widgets/image.py
```diff
@@ -855,6 +856,30 @@ def reset_vmin_vmax(self):
...
+
+        if reset_vmin_vmax and len(self.vmin_vmax_sliders) != 0:
```
kushalkolar, line 880: this should happen regardless of vmin vmax sliders existing, only based on the kwarg `reset_vmin_vmax`

kushalkolar (changes requested):

clewis7:
running into a small issue for setting the data when the ndims of the data increase beyond "txy"

setting the overall data:
`self._data[0] = new_data`

however, if the dims are higher than 3 and I am trying to update the current displayed image, I get broadcasting errors because the shapes become mismatched
`self.plot.graphics[0].data = self._data[0][self.current_index["t"]]`

would need to do something like 
`self.plot.graphics[0].data = self._data[0][self.current_index["t"]]self.current_index["z"]` etc.
for whatever dims you are capable of sliding through, but I'm not sure how to make this work if you have an indeterminate amount of dims you could be sliding through

fastplotlib/widgets/image.py
kushalkolar: Modify reset_vmin_vmax() so that it works when no sliders are available.

kushalkolar:
Just do the 0th index for all dims upto the last 2. 

Also I just realized we shouldn't directly set the graphic data, if windows functions or something are being used etc. So I would just set the current_index property. But it doesn't do anything if the current index is the same! So you could make a private method that has the logic for the current index setter with a "force" argument to force the update.

clewis7:
the examples I have to set_data are working locally, I think it is sufficient to just force a call to the current index setter in order to get the graphics to update

kushalkolar:
Maybe it's true that a simple call to the current_index setter is sufficient to set the image graphic data, haven't seen that code in a while. If it works then it works! :D

kushalkolar:
Maybe it's true that a simple call to the current_index setter is sufficient to set the image graphic data, haven't seen that code in a while. If it works then it works! :D

kushalkolar (approved):

Issue #222, opened 2023-06-12
should really just make `ImageWidget` use `GridPlot` even when using just 1 data array


kushalkolar:
gets rid of conditionals and repetitive code

@clewis7 thoughts?

clewis7:
I thought the same thing, would be a simple refactor I think and would handle the issues I was having in `set_data()`

clewis7:
can assign to me if you want

kushalkolar:
good luck with this :spaghetti:

kushalkolar:
done in #230

Issue #166, opened 2023-04-16
Histogram vmin vmax widget 


kushalkolar:
pyqtgraph-like histograms that can be tied to an Image. We can make this very nice looking. In addition to a vertical line graphic to display the histogram, we can have a thick vertical on the side with colors set to the current cmap!

![image](https://github.com/kushalkolar/fastplotlib/assets/9403332/ff159200-b472-44a7-a0ae-9b999d89a44d)

clewis7:
first step...fix histogram graphic lol

kushalkolar:
No not necessary, just use a line

kushalkolar:
Look at the pyqtgraph widget

kushalkolar:
So now after #244 we can probably just use bidirectional link() between LinearRegionSelector of the HistogramLUTWidget and the image cmap vmin, vmax

PR #226, opened 2023-06-12
begin refactor of image widget

clewis7:
#222 

refactor iw so that it only uses GridPlot

simplest way to enact is to just force np.ndarray data to be in shape (1,1)

works for now, need to do some more strenuous testing to make sure nothing is broken

### DIFF
--- a/fastplotlib/widgets/image.py
+++ b/fastplotlib/widgets/image.py
@@ -79,11 +79,11 @@ class _WindowFunctions:
 
 class ImageWidget:
     @property
-    def plot(self) -> GridPlot:
+    def gridplot(self) -> GridPlot:
         """
-        The plotter used by the ImageWidget. Will always be a ``GridPlot``.
+        ``GridPlot`` instance within the `ImageWidget`.
         """
-        return self._plot
+        return self._gridplot
 
     @property
     def data(self) -> List[np.ndarray]:
@@ -467,10 +467,10 @@ class ImageWidget:
         if grid_plot_kwargs is None:
             grid_plot_kwargs = {"controllers": "sync"}
 
-        self._plot: GridPlot = GridPlot(shape=grid_shape, **grid_plot_kwargs)
+        self._gridplot: GridPlot = GridPlot(shape=grid_shape, **grid_plot_kwargs)
 
         self.image_graphics = list()
-        for data_ix, (d, subplot) in enumerate(zip(self.data, self.plot)):
+        for data_ix, (d, subplot) in enumerate(zip(self.data, self.gridplot)):
             minmax = quick_min_max(self.data[data_ix])
 
             if self._names is not None:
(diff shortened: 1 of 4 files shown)


fastplotlib/widgets/image.py
```diff
@@ -79,9 +79,9 @@ def __repr__(self):
 
 class ImageWidget:
     @property
-    def plot(self) -> Union[Plot, GridPlot]:
+    def plot(self) -> GridPlot:
         """
-        The plotter used by the ImageWidget. Either a simple ``Plot`` or ``GridPlot``.
+        The plotter used by the ImageWidget. Will always be a ``GridPlot``.
```
kushalkolar, line 84:
```suggestion
The `GridPlot` instance within the `ImageWidget`
```

fastplotlib/widgets/image.py
```diff
@@ -79,9 +79,9 @@ def __repr__(self):
 
 class ImageWidget:
     @property
-    def plot(self) -> Union[Plot, GridPlot]:
+    def plot(self) -> GridPlot:
```
kushalkolar, line 82: let's just rename this to `gridplot` and then I will have fun refactoring elsewhere in mesmerize etc. :smile:

fastplotlib/widgets/image.py
```diff
@@ -466,91 +464,56 @@ def __init__(
...
-                data_range_40p = np.ptp(minmax) * 0.3
+                data_range_40p = np.ptp(minmax) * 0.4
 
                 minmax_slider = FloatRangeSlider(
                     value=minmax,
                     min=minmax[0] - data_range_40p,
                     max=minmax[1] + data_range_40p,
                     step=data_range / 150,
-                    description=f"min-max",
+                    description=f"mm: {name_slider}",
```
kushalkolar, line 492: I initially thought of doing this but the issue is that ipywidget descriptions are very short, if you can find a way to have longer descriptions that would be ideal.

Anyways this will become irrelevent once we make #166

fastplotlib/widgets/image.py
```diff
@@ -466,91 +464,56 @@ def __init__(
...
-                    _kwargs = kwargs
-
-                frame = self._process_indices(d, slice_indices=self._current_index)
-                frame = self._process_frame_apply(frame, data_ix)
-                ig = ImageGraphic(frame, name="image", **_kwargs)
-                subplot.add_graphic(ig)
-                subplot.name = name
-                subplot.set_title(name)
-                self.image_graphics.append(ig)
+                _kwargs = deepcopy(kwargs)
```
kushalkolar, line 505: maybe we just ignore this, the user can set it after if they want. This class is getting too polluted with unnecessary functionality

kushalkolar (changes requested):

clewis7:
image_widget notebook functioning, added small example for setting the data as well

PR #234, opened 2023-06-16
more examples

kushalkolar:
- [ ] line collection


kushalkolar:
We don't need this, line collection and line stack examples are in: https://github.com/fastplotlib/fastplotlib/tree/master/examples/desktop/line_collection

also in the selector example nbs

PR #235, opened 2023-06-16
add canvas kwarg that takes str

kushalkolar:
allows forcing a canvas with just a str kwarg:

```python
import fastplotlib as fpl
plot = fpl.Plot(canvas="glfw")

image_graphic = plot.add_image(data=data, name="random-image")

plot.show()
```

@EricThomson


EricThomson:
This is very cool: it looks like it can take options `jupyter`, `glfw`, and `qt` ? That's great!

kushalkolar:
Yup! So you don't have to instantiate the canvas and renderer yourself just to use a different framework.

PR #238, opened 2023-06-17
basic svg export

kushalkolar:
WIP, pygfx SVG renderer is giving weird outputs, not rendering Image objects, camera view is wrong, lines don't have nice endings.

contours example:
image


almarklein:
FWIW I don't expect the SVG renderer in pygfx to be getting any better for the foreseeable future. Unless someone else is interested in moving it forward.

kushalkolar:
It's really not a priority for us either. I can work on simpler png export some time later, but also not a high priority now.

PR #241, opened 2023-06-17
more cmap handling, add cmap_values

### LINKED ISSUE #191: More ColorsFeature handling
Take list or 1d array of numbers and automatically map these to colors based on the cmap argument. Useful for line collection for ROIs and line stacks, but also for single Line such as y_colorbar

- [ ] modify cmaps for line and scatter (position based graphics) after creating them, probably implement in `CmapFeature`
- [ ] modify cmap for colletions

- [x] Use an array of values for mapping the colors, for examples if we have a cosine wave allow passing in a 1D array of values that represent a variable for mapping from a colormap, for example set the cmap based on sine values for a cosine line. Or with caiman set the contour colors based on `r_values`
- [ ] there needs to be a way to modify these values after being set, maybe in the `CmapFeature` for `Line` and `Scatter`
- [x] modify the `cmap_values` after being set for collections

kushalkolar:
closes #191

### DIFF
--- a/examples/line/line.py
+++ b/examples/line/line.py
@@ -9,13 +9,8 @@ Example showing cosine, sine, sinc lines.
 from fastplotlib import Plot
 import numpy as np
 
-from wgpu.gui.offscreen import WgpuCanvas
-from pygfx import WgpuRenderer
 
-canvas = WgpuCanvas()
-renderer = WgpuRenderer(canvas)
-
-plot = Plot(canvas=canvas, renderer=renderer)
+plot = Plot()
 
 xs = np.linspace(-10, 10, 100)
 # sine wave
--- a/examples/line/line_cmap.py
+++ b/examples/line/line_cmap.py
@@ -1,0 +1,38 @@
+"""
+Line Plot
+============
+Example showing cosine, sine, sinc lines.
+"""
+
+# test_example = true
+
+from fastplotlib import Plot, run
+import numpy as np
+
+
+plot = Plot()
+
+xs = np.linspace(-10, 10, 100)
+# sine wave
+ys = np.sin(xs)
+sine = np.dstack([xs, ys])[0]
+
+# cosine wave
+ys = np.cos(xs) + 5
+cosine = np.dstack([xs, ys])[0]
+
+# sinc function
+a = 0.5
+ys = np.sinc(xs) * 3 + 8
+sinc = np.dstack([xs, ys])[0]
+
+# cmap_values from an array, so the colors on the sine line will be based on the sine y-values
+sine_graphic = plot.add_line(data=sine, thickness=10, cmap="plasma", cmap_values=sine[:, 1])
+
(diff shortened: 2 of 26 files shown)


kushalkolar:
@EricThomson @clewis7 can I get your thoughts on this.

The use case for this in `LineCollection` is to make it really easy to color contours using a cmap based on the various quality metrics! For example let's say you used the "jet" cmap and wanted to visualize the cnn classifier predictions, things which are probably neurons would have red contours, and low classified things would be blue.

It also allows for using qualitative colormaps and giving it arrays of ints, like cluster labels. So you could see the spatial distribution of different clusters for example.

(meanwhile in matplotlib, it's really hard :upside_down_face: https://stackoverflow.com/questions/8500700/how-to-plot-a-gradient-color-line)

usage is really simple:

```python
 cmap_values from an array, so the colors on the sine line will be based on the sine y-values
sine_graphic = plot.add_line(
    data=sine, 
    thickness=10, 
    cmap="plasma", 
    cmap_values=sine[:, 1]
)

# qualitative colormaps, useful for cluster labels for example
cmap_values = [0] * 25 + [5] * 25 + [1] * 25 + [2] * 25
cosine_graphic = plot.add_line(
    data=cosine, 
    thickness=10, 
    cmap="tab10", 
    cmap_values=cmap_values
)
```

image

For collections:

```python
plot.add_line_collection(
    circles, 
    cmap="bwr", 
    cmap_values=cmap_values, 
    thickness=10
)
```

image

```python
# this makes 16 circles, so we can create 16 cmap values, so it will use these values to set the
# color of the line based by using the cmap as a LUT with the corresponding cmap_value

# qualitative colormap used for mapping 16 cmap values for each line
# for example, these could be cluster labels
cmap_values = [
    0, 1, 1, 2,
    0, 0, 1, 1,
    2, 2, 3, 3,
    1, 1, 1, 5
]

plot = fpl.Plot()

plot.add_line_collection(
    circles, 
    cmap="tab10", 
    cmap_values=cmap_values, 
    thickness=10
)

plot.show()

plot.canvas.set_logical_size(800, 800)
```

image

kushalkolar:
update scatter to use this as well

```python
from fastplotlib import Plot, run
import numpy as np
from pathlib import Path
from sklearn.cluster import AgglomerativeClustering


plot = Plot()

data_path = Path(__file__).parent.parent.joinpath("data", "iris.npy")
data = np.load(data_path)


agg = AgglomerativeClustering(n_clusters=3)

agg.fit_predict(data)


scatter_graphic = plot.add_scatter(
    data=data[:, :-1],
    sizes=15,
    alpha=0.7,
    cmap="Set1",
    cmap_values=agg.labels_
)

plot.show()

plot.canvas.set_logical_size(800, 800)

plot.auto_scale()

scatter_graphic.cmap = "tab10"

if __name__ == "__main__":
    run()
```

scatter_cmap

PR #247, opened 2023-06-20
allowing selectors to be indexed by name

### LINKED ISSUE #246: indexing selectors by name
currently can only index plot graphics by name, for example:

```
import fastplotlib as fpl
import numpy as np

plot  = fpl.Plot()
plot.add_image(data=data, name='image')

plot['image']
```

however, if you have named selectors, you cannot index them in the same manor

I think this would be a nice feature to have, would help me with my behavior stuff

@kushalkolar thoughts?

clewis7:
closes #246

### DIFF
--- a/fastplotlib/layouts/_base.py
+++ b/fastplotlib/layouts/_base.py
@@ -320,6 +320,9 @@ class PlotArea:
         for g in self.graphics:
             graphic_names.append(g.name)
 
+        for s in self.selectors:
+            graphic_names.append(s.name)
+
         if name in graphic_names:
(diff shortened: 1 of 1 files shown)


clewis7:
was curious to see if opening a PR as a draft would keep CI from running, does not appear so

kushalkolar:
LGTM, in the future we can think about maybe having dictionary lookup for graphic or selector names instead of the current for loop, it will be much faster. However I'm not sure of a usecase for using names when you have thousands of graphics, those types of situations are usually numerically indexed, so just a maybe for now.

PR #248, opened 2023-06-20
auto mixin class for generating add_<graphic> methods

### LINKED ISSUE #157: auto-generate mixin class for `Subplot` to hold all the `add_<graphic>` methods
This will solve our issues with doc generation for all these methods like `Plot.add_image()`.

clewis7:
closes #157


clewis7:
automixin class generated looks good, but then when I go to use the methods I get errors with the `Graphic` classes not being defined even though I am importing them in the file

fastplotlib/layouts/_base.py
```diff
@@ -12,14 +12,16 @@
 from ..graphics._base import Graphic, GraphicCollection
 from ..graphics.selectors._base_selector import BaseSelector
 
+from ..utils import GraphicMethods
```
kushalkolar, line 15: let's make a new module specifically for this, like `_add_graphic_mixin`

fastplotlib/layouts/_base.py
```diff
@@ -12,14 +12,16 @@
...
+
 # dict to store Graphic instances
 # this is the only place where the real references to Graphics are stored in a Python session
 # {hex id str: Graphic}
 GRAPHICS: Dict[str, Graphic] = dict()
 SELECTORS: Dict[str, BaseSelector] = dict()
 
 
-class PlotArea:
+class PlotArea(GraphicMethods):
```
kushalkolar, line 24: mixin to subplot, not here, we want to keep `PlotArea` as a very bare base class

fastplotlib/utils/add_graphics.py
```diff
@@ -0,0 +1,537 @@
+from typing import *
+
+import numpy
+
+from fastplotlib.graphics import *
```
kushalkolar, line 5: use relative import

fastplotlib/utils/generate_add_methods.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+
+modules = dict(inspect.getmembers(fastplotlib.graphics))['__all__']
+
+
+def generate_add_graphics_methods():
+    # clear file and regenerate from scratch
+    open('add_graphics.py', 'w').close()
+
+    sys.stdout = open('add_graphics.py', 'w')
```
kushalkolar, line 13: just use regular file open and writing instead of stdout and print, I didn't know this is even possible :laughing:

fastplotlib/utils/add_graphics.py
```diff
@@ -0,0 +1,537 @@
...
+            data = np.random.rand(512, 512)
+            # plot the image data
+            plot.add_image(data=data)
+            # show the plot
+            plot.show()
+        
+		"""
+        g = ImageGraphic(*args, **kwargs)
+        self.add_graphic(g)
+
```
kushalkolar, line 71: also need to return a proxy to the graphic

fastplotlib/utils/generate_add_methods.py
```diff
@@ -0,0 +1,38 @@
...
+
+    print("\nclass GraphicMethods:")
+    print("\tdef __init__(self):")
+    print("\t\tpass\n")
+
+    for m in modules:
+        class_name = getattr(sys.modules[__name__], m)
+        method_name = class_name.type
+
+        print(f"\tdef add_{method_name}{inspect.signature(class_name.__init__)}:")
```
kushalkolar, line 27: add the return type annotation as well

kushalkolar (changes requested):

fastplotlib/layouts/_base.py
clewis7: renamed the file to above, but I just put the new module in utils, can move somewhere else if you prefer

fastplotlib/utils/add_graphics.py
clewis7: when I use relative import I get this error:

`ImportError: attempted relative import with no known parent package`

clewis7: Even with relative import, I get errors that the graphics are not being imported

so then when I have the return type defined, I am getting import errors that the graphic names are note defined which will also impact the instantiation in each of the methods of a given graphic

clewis7:
seem to be having some importing issues, not sure how to fix but the generated class should be looking better!

fastplotlib/layouts/_base.py
kushalkolar: Make the new module here in layouts

fastplotlib/utils/add_graphics.py
kushalkolar: You probably need to go up two levels, from ..graphics

kushalkolar: Paste the errors

clewis7:
```
  File "/home/clewis7/repos/fastplotlib/fastplotlib/layouts/generate_add_methods.py", line 3, in <module>
    from ..graphics import *
ImportError: attempted relative import with no known parent package
```

clewis7:
```
Cell In[7], line 1
----> 1 plot.add_image(data=data)

File ~/repos/fastplotlib/fastplotlib/layouts/_add_graphic_mixin.py:168, in GraphicMethodsMixin.add_image(self, data, vmin, vmax, cmap, filter, isolated_buffer, *args, **kwargs)
    111 	def add_image(self, data: Any, vmin: int = None, vmax: int = None, cmap: str = 'plasma', filter: str = 'nearest', isolated_buffer: bool = True, *args, **kwargs) -> weakref.proxy(ImageGraphic):
    112 		"""
    113 	
    114         Create an Image Graphic
   (...)
    166         
    167 		"""
--> 168 		g = ImageGraphic(*args, **kwargs)
    169 		self.add_graphic(g)
    171 		return weakref.proxy(g)

TypeError: ImageGraphic.__init__() missing 1 required positional argument: 'data'
```

fastplotlib/layouts/_subplot.py
```diff
@@ -124,16 +126,16 @@ def __init__(
...
-        for graphic_cls_name in graphics.__all__:
-            cls = getattr(graphics, graphic_cls_name)
-
-            pfunc = partial(self._create_graphic, cls)
-            pfunc.__signature__ = signature(cls)
-            pfunc.__doc__ = cls.__init__.__doc__
-
-            # cls.type is defined in Graphic.__init_subclass__
-            setattr(self, f"add_{cls.type}", pfunc)
+        # # attach all the add_<graphic_name> methods
```
kushalkolar, line 129: Delete

fastplotlib/utils/generate_add_methods.py
```diff
@@ -0,0 +1,45 @@
...
+
+    f = open('../layouts/_add_graphic_mixin.py', 'w')
+
+    f.write('from typing import *\n')
+    f.write('import numpy\n')
+    f.write('from ..graphics import *\n')
+    f.write('import weakref\n\n')
+
+    f.write("\nclass GraphicMethodsMixin:\n")
+    f.write("\tdef __init__(self):\n")
```
kushalkolar, line 23: Spaces instead of tabs because that's what we use on the rest of our codebase.

fastplotlib/utils/generate_add_methods.py
```diff
@@ -0,0 +1,45 @@
...
+#from ..graphics import *
+modules = list()
+
+for name, obj in inspect.getmembers(sys.modules[__name__]):
+    if inspect.isclass(obj):
+        modules.append(obj)
+
+def generate_add_graphics_methods():
+    # clear file and regenerate from scratch
+    open('../layouts/_add_graphic_mixin.py', 'w').close()
```
kushalkolar, line 13: Use `__file__` to get the path to the current module, lookup `__file__`

fastplotlib/layouts/_subplot.py
```diff
@@ -38,12 +38,12 @@
...
 from ._utils import make_canvas_and_renderer
 from ._base import PlotArea
-from .. import graphics
 from ..graphics import TextGraphic
 from ._defaults import create_camera, create_controller
+from ._add_graphic_mixin import GraphicMethodsMixin
 
 
-class Subplot(PlotArea):
+class Subplot(PlotArea, GraphicMethodsMixin):
```
kushalkolar, line 46: Give the module and class the same name

fastplotlib/layouts/_subplot.py
```diff
@@ -38,12 +38,12 @@
 
 from ._utils import make_canvas_and_renderer
 from ._base import PlotArea
-from .. import graphics
```
kushalkolar, line 41: ?

clewis7:
The only way I could get the generated `add_<graphic>` methods to work was to include the non-default arguments (mainly `data=data`) in the instantiation of the given graphic for a method

all of the graphics have a non-default `data` arg except for text, which for now I just manually changed, will need to think of a long-term solution as more graphics are added

I think it has to do with the nature of a mixin class, both the mixin class and the graphic class expect the kwarg arg `data` and if it is not specified in the graphic instantiation in the mixin method then it is not getting the `data` arg at all

clewis7:
I think maybe my screenshots will be different because of the things that you changed in recent PRs?

fastplotlib/layouts/graphic_methods_mixin.py
```diff
@@ -0,0 +1,558 @@
...
+        if 'name' in kwargs.keys():
+            self._check_graphic_name_exists(kwargs['name'])
+
+        graphic = graphic_class(*args, **kwargs)
+        self.add_graphic(graphic, center=center)
+
+        # only return a proxy to the real graphic
+        return weakref.proxy(graphic)
+
+    def add_heatmap(self, data: Any, vmin: int = None, vmax: int = None, cmap: str = 'plasma', filter: str = 'nearest', chunk_size: int = 8192, isolated_buffer: bool = True, *args, **kwargs) -> weakref.proxy(HeatmapGraphic):
```
kushalkolar, line 28: can you change the type annotations to use `[]` instead of `()`:

```python
-> weakref.proxy[Graphic]
```

fastplotlib/utils/generate_add_methods.py
```diff
@@ -0,0 +1,67 @@
...
+        modules.append(obj)
+
+def generate_add_graphics_methods():
+    # clear file and regenerate from scratch
+    current_module = pathlib.Path(__file__).parent.parent.resolve()
+
+    open(current_module.joinpath('layouts/graphic_methods_mixin.py'), 'w').close()
+
+    f = open(current_module.joinpath('layouts/graphic_methods_mixin.py'), 'w')
+
```
kushalkolar, line 21: Add a disclaimer comment to the very top of the `graphic_methods_mixin.py` file to state that this module is auto-generated and should not be modified directly

kushalkolar (changes requested):

fastplotlib/layouts/graphic_methods_mixin.py
clewis7: if I change the type annotation to this, it thinks that I am trying to index `weakref.proxy`

kushalkolar:
Actually let's just set the type annotation as the `Graphic` that is returned and put in the docstring that it always returns a proxy.  I think that makes inspection tools and tab completion etc. to work properly, rather than putting them as `proxy[Graphic]`

kushalkolar:
last question, docstrings and tab completion work in jupyterlab?

clewis7:
yeah, works for me!

kushalkolar (approved):

PR #249, opened 2023-06-21
fix CI to only run when not draft PR

### LINKED ISSUE #232: running CI only when not a draft
want to make sure that CI does not run until PR is complete so that we do not waste action minutes 
- need to investigate if CI runs for drafts
- how to automatically make a PR a draft when first created

clewis7:
closes #232

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,12 +7,18 @@ on:
   pull_request:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


clewis7:
Seems to work, CI got skipped!

kushalkolar:
Mark as ready for review and see if it gets triggered?

PR #254, opened 2023-06-22
remove JupyterWgpuCanvas from imports, allow minimal install to work

kushalkolar:


kushalkolar:
tested and work with just glfw, docs also build

PR #267, opened 2023-06-29
Sync Subplots by Name

ArjunPutcha:
Currently only done upon initialization. Will work on making this property mutable post-initialization soon.


fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
                 self.shape
             )
 
+        if "names" in kwargs.keys():
+            self.names = to_array(kwargs["names"])
+            print(self.names)
```
clewis7, line 97:
```suggestion
            print(self.names)
```

examples/notebooks/gridplot_simple.ipynb
kushalkolar, line 1: clear outputs before comitting notebooks

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
+            if self.names.shape != self.shape:
+                raise ValueError
```
clewis7, line 99:
```suggestion
                raise ValueError(f"subplot names: {self.names} must be in gridplot shape: {self.shape}")
```

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
                     self.shape[0] * self.shape[1], dtype=int
                 ).reshape(self.shape)
 
+        if (self.names is not None) & all(isinstance(item, str) for item in controllers[0]):
+            idx_async_cont = np.shape(controllers)[0]
+            c = to_array(controllers)
+            controllers = np.zeros(
+                self.shape[0] * self.shape[1], dtype=int
+            ).reshape(self.shape)
+            for i in range(self.shape[0]):
```
clewis7, line 115: can just use `itertools.product` 

```
positions = list(product(range(self.shape[0]), range(self.shape[1])))
```

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
+        else:
+            self.names = None
+
         if isinstance(controllers, str):
             if controllers == "sync":
                 controllers = np.zeros(
                     self.shape[0] * self.shape[1], dtype=int
                 ).reshape(self.shape)
 
+        if (self.names is not None) & all(isinstance(item, str) for item in controllers[0]):
```
clewis7, line 109: will you add comments here?

clewis7 (changes requested):

fastplotlib/layouts/_gridplot.py
kushalkolar: I agree, this is quite nested.

btw, `list` is not necessary when using product, can just do `for positions in product(...)`

ArjunPutcha: merci beaucoup, changed.

examples/notebooks/gridplot_simple.ipynb
kushalkolar, line 1: I see that in the example you've given a mix of subplot names and an integer for controllers. I think that if subplots are to be synced using plot names, then _all_ plot names should be provided. 

So basically we allow setting controllers in the following ways:
1. `None`, i.e. not passed, every subplot has its own controller
2. `"sync"`, - syncs all controllers
3. provide integers 
4. provide list of list of names (this PR), each sublist gets the same controllers

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
             if controllers == "sync":
                 controllers = np.zeros(
                     self.shape[0] * self.shape[1], dtype=int
                 ).reshape(self.shape)
 
+        # Check if 'names' and 'controllers' both have values; check if values in controller array are strings
+        if (self.names is not None) and (controllers is not None):
+            c = to_array(controllers)
+            if isinstance(c[0, 0], str):
+                idx_async_controllers = np.shape(c)[0]
```
kushalkolar, line 112: This is messy, use `enumerate` instead of an incrementer:

```python

positions_iter = product(range(self.shape[0]), range(self.shape[1]))
for i, position in enumerate(positions_iter):
  ...
```

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
         if isinstance(controllers, str):
             if controllers == "sync":
                 controllers = np.zeros(
                     self.shape[0] * self.shape[1], dtype=int
                 ).reshape(self.shape)
 
+        # Check if 'names' and 'controllers' both have values; check if values in controller array are strings
+        if (self.names is not None) and (controllers is not None):
+            c = to_array(controllers)
+            if isinstance(c[0, 0], str):
```
kushalkolar, line 111: do a check to make sure all are str

fastplotlib/layouts/_gridplot.py
```diff
@@ -34,14 +34,14 @@ def to_array(a) -> np.ndarray:
...
-        size: Tuple[int, int] = (500, 300),
-        **kwargs,
+            self,
+            shape: Tuple[int, int],
+            cameras: Union[np.ndarray, str] = "2d",
+            controllers: Union[np.ndarray, str] = None,
+            canvas: Union[str, WgpuCanvas, pygfx.Texture] = None,
+            renderer: pygfx.WgpuRenderer = None,
+            size: Tuple[int, int] = (500, 300),
+            **kwargs,
```
kushalkolar, line 44: don't change the spacing, we're trying to figure out a linter to use but meanwhile try to keep it the same

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
                 self.shape
             )
 
+        if "names" in kwargs.keys():
+            self.names = to_array(kwargs["names"])
+            if self.names.shape != self.shape:
+                raise ValueError(f"subplot names: {self.names} must be in gridplot shape: {self.shape}")
+        else:
+            self.names = None
```
kushalkolar, line 100: we should probably just add `names` as a kwarg with default value of `None`. Allow it to be either list of list of `str` or array of `str`

kushalkolar: And then add a sanity check to make sure if `names` is provided that it conforms to the types

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
+        # Check if 'names' and 'controllers' both have values; check if values in controller array are strings
+        if (self.names is not None) and (controllers is not None):
+            c = to_array(controllers)
+            if isinstance(c[0, 0], str):
+                idx_async_controllers = np.shape(c)[0]
+                controllers = np.zeros(
+                    self.shape[0] * self.shape[1], dtype=int
+                ).reshape(self.shape)
+                for positions in product(range(self.shape[0]), range(self.shape[1])):
+                    item = np.argwhere(c == self.names[positions])
```
kushalkolar, line 117: I don't understand what `item` is, can you write this clearer with a more informative name?

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,12 +92,35 @@ def __init__(
...
+                    if len(item) != 0:
```
kushalkolar, line 118: ??

fastplotlib/layouts/_gridplot.py
```diff
@@ -116,8 +139,8 @@ def __init__(
         # create controllers if the arguments were integers
         if np.issubdtype(controllers.dtype, np.integer):
             if not np.all(
-                np.sort(np.unique(controllers))
-                == np.arange(np.unique(controllers).size)
+                    np.sort(np.unique(controllers))
+                    == np.arange(np.unique(controllers).size)
```
kushalkolar, line 143: don't change the spacing

fastplotlib/layouts/_gridplot.py
```diff
@@ -226,10 +242,10 @@ def _call_animate_functions(self, funcs: Iterable[callable]):
...
 
     def add_animations(
-        self,
-        *funcs: Iterable[callable],
-        pre_render: bool = True,
-        post_render: bool = False,
+            self,
+            *funcs: Iterable[callable],
+            pre_render: bool = True,
+            post_render: bool = False,
```
kushalkolar, line 248: spacing

fastplotlib/layouts/_gridplot.py
```diff
@@ -281,7 +297,7 @@ def remove_animation(self, func):
             self._animate_funcs_post.remove(func)
 
     def show(
-        self, autoscale: bool = True, maintain_aspect: bool = None, toolbar: bool = True
+            self, autoscale: bool = True, maintain_aspect: bool = None, toolbar: bool = True
```
kushalkolar, line 300: spacing

kushalkolar (changes requested):

fastplotlib/layouts/_gridplot.py
ivoflipse: Ruff!

kushalkolar: Does it have an option to set code lines length but allow longer docstring length? Large docstrings are useful for tables!

fastplotlib/layouts/_gridplot.py
ArjunPutcha: should to_array function set dtype = "o" to avoid typecasting all elements to same type then? Right now it calls np.array(a) with no dtype set, so if even 1 str is in the list, all list elements are converted to str

examples/notebooks/gridplot_simple.ipynb
ArjunPutcha: Got it. Right now default is if the name isn't provided it is defaulted to unique controllers. You want to explicitly require if a names kwarg is provided, it must be the same shape as provided shape. Correct?

fastplotlib/layouts/_gridplot.py
ArjunPutcha: Perfect

fastplotlib/layouts/_gridplot.py
ArjunPutcha: enumerate doesn't fulfill same role in this case, but since we're requiring names kwarg to have same shape now we don't need this anymore. Will rewrite

fastplotlib/layouts/_gridplot.py
ArjunPutcha: dtype=object *

fastplotlib/layouts/_gridplot.py
ivoflipse: No clue, but ruff is so fast that fastplotlib must use it 😉

clewis7 (commented):
otherwise LGTM, maybe just add some code comments here and there so later we know what is happening

fastplotlib/layouts/_gridplot.py
```diff
@@ -92,11 +93,41 @@ def __init__(
...
+        if controllers is not None:
+            if isinstance(controllers, str):
+                if controllers == "sync":
+                    controllers = np.zeros(
+                        self.shape[0] * self.shape[1], dtype=int
+                    ).reshape(self.shape)
+            else:
+                c = to_array(controllers)
+                if (c.shape[0]*c.shape[1]) != (self.shape[0]*self.shape[1]):
+                    raise ValueError(f"number of controllers: {controllers} must be the same as number of elements "
```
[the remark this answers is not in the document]
ArjunPutcha: Good catch, just modified

kushalkolar:
Now that #280 exists, we're going to need to figure out how to consolidate this :smile:

ArjunPutcha:
Anything you want me to do on my end? Or will you handle it?

kushalkolar:
We'll have to brainstorm a bit, probably at scipy sprint?

kushalkolar:
superseeded by #382

PR #271, opened 2023-06-30
min version pins for numpy, jupyter-rfb, ipywidgets

kushalkolar:
- numpy < 1.23 causes issues with type annotations in the form: `-> np.ndarray[Graphic]`
- use new ipywidgets and jupyter rfb, older ones in combinations can be a mess

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ extras_require = {
         "sphinx",
         "pydata-sphinx-theme<0.10.0",
         "glfw",
-        "jupyter-rfb>=0.4.1"  # required so ImageWidget docs show up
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,13 +13,15 @@
         "sphinx",
         "pydata-sphinx-theme<0.10.0",
         "glfw",
-        "jupyter_rfb"  # required so ImageWidget docs show up
+        "jupyter-rfb>=0.4.1"  # required so ImageWidget docs show up
```
kushalkolar, line 16:
```suggestion
        "jupyter-rfb>=0.4.1",  # required so ImageWidget docs show up
```

kushalkolar (commented):
forgot a comma

PR #277, opened 2023-07-06
notebook screenshot tests

kushalkolar:


kushalkolar:
So it seems like non-git LFS files need to be changed and commited and pushed to the branch for the other new LFS files to be picked up by github actions :man_shrugging:

PR #279, opened 2023-07-06
using sphinx gallery to add examples to docs automatically

clewis7:
huge WIP

PR #281, opened 2023-07-07
removing outdated code block examples

clewis7:
all examples are in docs now, will eventually have user guide and examples rendering in docs, see #279


kushalkolar:
You'll also have to regenerate the add_graphic Mixin after.

kushalkolar:
@clewis7 can you just check quick that this is everything, regenerate mixin, and then mark for review so CI runs?

Issue #297, opened 2023-08-23
autoconnect existing ipywidgets to selections 


kushalkolar:
Came up while helping Minel on her use cases, instead of manually connecting an ipywidget's `observe`, we could make it simpler by providing a `add_ipywidget_handler(<ipywidget_instance>)` that links the ipywidget and `Selector`. 

Thoughts:
- should this be bidirectional? I think yes, any caveats (recurusion etc.?)
- `LinearSelector` is compatible with `IntSlider` and `FloatSlider`, `FloatLogSlider`
- `LinearRegionSelector` is compatible with `IntRangeSlider` and `FloatRangeSlider` 
- I don't think it makes sense for `RectangleRegionSelector` to be connectable to any corresponding `ipywidget`?

@clewis7 thoughts?

clewis7:
I think this sounds fine, perhaps there would be issues with recursion if bidirectional when handling events

Just to clarify you are saying if you added a linear selector to a plot and then wanted to connect to a ipywidget to the selector to have a general method for connecting the linear selector to the ipywidget? 

I am struggling to think of the use case, what was Minel trying to do?

minel-arinel:
I have an ImageWidget with a calcium imaging recording and a line collection of contours plotted on top. I also have a separate Plot that shows the line graph of temporal responses of a clicked cell on the ImageWidget. On the Plot, I want to add a LinearSelector to mark as a time point of the ImageWidget. So basically just being able to slide that LinearSelector left-right on the temporal Plot as the ImageWidget 't' slider is changing.

kushalkolar:
@minel-arinel after #298 you can just do:

```python
selector.add_ipywidget_handler(iw.sliders["t"])
```

kushalkolar:
@minel-arinel #298 is now merged

PR #298, opened 2023-09-07
auto connect a ipywidget slider to a linear selector

kushalkolar:
implements #297 for int and float sliders

@minel-arinel 

Usage:
```python
# make ipywidget
ipywidget_slider = IntSlider(min=0, max=100, description="slider")

# connect to selector
selector.add_ipywidget_handler(ipywidget_slider)
```

### DIFF
--- a/fastplotlib/graphics/selectors/_linear.py
+++ b/fastplotlib/graphics/selectors/_linear.py
@@ -25,6 +25,8 @@ class LinearSelector(Graphic, BaseSelector):
 
     @limits.setter
     def limits(self, values: Tuple[float, float]):
+        # check that `values` is an iterable of two real numbers
+        # using `Real` here allows it to work with builtin `int` and `float` types, and numpy scaler types
         if len(values) != 2 or not all(map(lambda v: isinstance(v, Real), values)):
             raise TypeError(
                 "limits must be an iterable of two numeric values"
--- a/fastplotlib/graphics/selectors/_linear.py
+++ b/fastplotlib/graphics/selectors/_linear.py
@@ -196,7 +196,7 @@ class LinearSelector(Graphic, BaseSelector):
 
     def _ipywidget_callback(self, change):
         # update the LinearSelector if the ipywidget value changes
-        if self._block_ipywidget_call:
+        if self._block_ipywidget_call or self._moving:
             return
 
         self.selection = change["new"]
--- a/fastplotlib/graphics/selectors/_linear.py
+++ b/fastplotlib/graphics/selectors/_linear.py
@@ -27,7 +27,6 @@ class LinearSelector(Graphic, BaseSelector):
         parent: Graphic = None,
         end_points: Tuple[int, int] = None,
         arrow_keys_modifier: str = "Shift",
-        ipywidget_slider=None,
         thickness: float = 2.5,
         color: Any = "w",
         name: str = None,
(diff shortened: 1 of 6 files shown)


kushalkolar:
ipywidget -> selector is working well, selector -> ipywidget is not working well if the step size is not 1, it's wonky and jumpy

kushalkolar:
ok this works now, made an attr `_moving` for `BaseSelector` :D

kushalkolar:
Implemented for `LinearRegionSelector` as well, and made `limits` a settable proprety for both `LinearRegionSelector` and `LinearSelector`. 

@clewis7 will merge after your review

https://github.com/fastplotlib/fastplotlib/assets/9403332/449bbebc-097f-4a42-86cf-51fc98c59947

https://github.com/fastplotlib/fastplotlib/assets/9403332/b2a841cd-e51d-4325-a01f-05db6d2113d4

fastplotlib/graphics/selectors/_linear.py
```diff
@@ -18,6 +19,19 @@
 
 
 class LinearSelector(Graphic, BaseSelector):
+    @property
+    def limits(self) -> Tuple[float, float]:
+        return self._limits
+
+    @limits.setter
+    def limits(self, values: Tuple[float, float]):
+        if len(values) != 2 or not all(map(lambda v: isinstance(v, Real), values)):
```
clewis7, line 28: add code comment here for `not all(map(lambda v: isinstance(v, Real), values))`

pretty ambiguous as is, comment will be helpful later to remember what is happening

fastplotlib/graphics/selectors/_linear.py
```diff
@@ -18,6 +19,19 @@
...
+            raise TypeError(
+                "limits must be an iterable of two numeric values"
+            )
+        self._limits = tuple(map(round, values))  # if values are close to zero things get weird so round them
```
clewis7, line 32: where does `round` come from? I don't see any definitions or imports

fastplotlib/graphics/selectors/_linear.py
```diff
@@ -84,7 +94,8 @@ def __init__(
         if len(limits) != 2:
             raise ValueError("limits must be a tuple of 2 integers, i.e. (int, int)")
 
-        limits = tuple(map(round, limits))
+        self._limits = tuple(map(round, limits))
```
clewis7, line 97: see above

fastplotlib/graphics/selectors/_linear_region.py
```diff
@@ -9,6 +18,19 @@
 
 
 class LinearRegionSelector(Graphic, BaseSelector):
+    @property
+    def limits(self) -> Tuple[float, float]:
+        return self._limits
+
+    @limits.setter
+    def limits(self, values: Tuple[float, float]):
+        if len(values) != 2 or not all(map(lambda v: isinstance(v, Real), values)):
```
clewis7, line 27: same as above

clewis7 (commented):
mostly looks good to me, I think just some more code comments would be helpful for understanding what is going on

From what I see, and correct me if I am wrong...

A user will create an ipywidget slider and then can add this to a linear selector using the "add" method; internally you set up the widget to have bidirectional linkage with the associated linear selector

A given selector looks like it can have more than one associated ipywidget, which are managed by a private list

fastplotlib/graphics/selectors/_linear.py
kushalkolar: checks that `values` is an iterable of two real numbers

kushalkolar:
> A user will create an ipywidget slider and then can add this to a linear selector using the "add" method; internally you set up the widget to have bidirectional linkage with the associated linear selector

Yes, they add a handler. 

> A given selector looks like it can have more than one associated ipywidget, which are managed by a private list

yes

PR #300, opened 2023-09-07
add jupyter-sidecar as a dependency, update simple notebook

### LINKED ISSUE #299: make `jupyterlab-sidecar` a nb dependency?
@clewis7 they are now on jupyterlab v4, thoughts? It will make the demo nbs nicer.

https://github.com/jupyter-widgets/jupyterlab-sidecar

clewis7:
closes #299 

Only updated `simple.ipynb` because the other ones are pretty short


kushalkolar:
idea: what if automate the sidecar part with a kwarg to `show()`, such as:

```python
plot.show(sidecar=True)  # default is True

# if there should be some kwargs passed to sidecar, such as the title for it etc. 
plot.show(sidecar=True, sidecar_kwargs={...})
```

In `ImageWidget` this would have to be implemented such that the sliders are slow in the sidecar, basically the full `VBox` get sidecar'd

kushalkolar:
What happens when you run the tests? Does sidecar interfere with creating the screenshots?

clewis7:
tests passing locally for me, just needed to add `sidecar` dependency to "tests" in `setup.py`

clewis7:
> idea: what if automate the sidecar part with a kwarg to `show()`, such as:
> 
> ```python
> plot.show(sidecar=True)  # default is True
> 
> # if there should be some kwargs passed to sidecar, such as the title for it etc. 
> plot.show(sidecar=True, sidecar_kwargs={...})
> ```
> 
> In `ImageWidget` this would have to be implemented such that the sliders are slow in the sidecar, basically the full `VBox` get sidecar'd

I think this would be cool bc it alleviates having to have:
```
sc = Sidecar()
with sc:
    display(plot.show())
```

also would be nice to have `plot.close()` also close the sidecar if there is one

Maybe each plot instance would have a `self.sidecar` attribute that is initially `None` and if `sidecar=True` in `plot.show()` it gets initialized?

kushalkolar:
You will need to figure out how sidecar works with mulitple `with sc: display(plot.show())` calls, and then we will need to determine what type of behavior we want.

Thoughts to get you started:
- Do repeated `with sc: display(plot.show())` calls create new sidecars? Or does it do nothing?
- If the user closes the `sidecar` but not the plot, what do repeated `with sc: display(plot.show())` calls do? 
  - I think that if closing the `sidecar` allows us to re-display it, we can have the following type of behavior: 
    - `plot.close()` closes the plot and sidecar.
    - if the user closes the sidecar by pressing the "**x**", the plot can be re-displayed in the `sidecar` using `plot.show()`. So users must explicitly close plots using `plot.close()`, closing the sidecar keeps them in GPU RAM.

clewis7:
Interesting find today, multiple calls of
```
with sc:
    display(plot.show())
```
across different running notebooks will open the different plots in the same sidecar with different tabs

clewis7:
image

Subsequent calls to 
``` python
with sc:
    display(plot.show())
```
opens the plot a second time in the same tab of the existing sidecar side by side

clewis7:
closing the sidecar does **NOT** close the plot, can then be re-rendered either in another sidecar or in a cell

clewis7:
if you close the sidecar with **"x"**, then the sidecar instance has been closed and subsequent calls to

``` python
with sc:
     display(plot.show()))
```
does nothing

clewis7:
With all of the above comments in mind:

I think it is most relevant to have:

``` python
plot.show(sidecar=True, sidecar_kwargs={...})
```
and should function as follows:

- default for all plots is for it to be displayed in sidecar, `sidecar=True`
- `plot.close()` will close the sidecar and plot instance 
- clicking the **"x"** will close the sidecar but keep the plot in memory and redisplaying the plot via `plot.show()` will open sidecar with plot again (I think if we keep track of sidecar instance for a plot with `self.sidecar`, then can continue to use the instance to redisplay plot even if closed with **"x"**...will have to check if this works)

clewis7:
@kushalkolar added sidecar for plot to start, two issues:
1) if you close a sidecar with **"x"**, calling `plot.show()` again will do nothing
2) I think that this breaks the notebook tests that you have in place, see error message below 

``` python
Cell In[17], line 2
      1 # testing cell, ignore
----> 2 plot_test("astronaut", plot)

File ~/repos/fastplotlib/examples/notebooks/nb_test_utils.py:32, in plot_test(name, plot)
     29         regenerate_screenshot(name, snapshot.data)
     31 try:
---> 32     assert_screenshot_equal(name, snapshot.data)
     33 except AssertionError:
     34     FAILURES.append(name)

File ~/repos/fastplotlib/examples/notebooks/nb_test_utils.py:44, in assert_screenshot_equal(name, data)
     41 def assert_screenshot_equal(name, data):
     42     ground_truth = iio.imread(SCREENSHOTS_DIR.joinpath(f"nb-{name}.png"))
---> 44     is_similar = np.allclose(data, ground_truth)
     46     update_diffs(name, is_similar, data, ground_truth)
     48     assert is_similar, (
     49         f"notebook snapshot for {name} has changed"
     50     )
...
File ~/venvs/mescore/lib/python3.11/site-packages/numpy/core/numeric.py:2356, in isclose.<locals>.within_tol(x, y, atol, rtol)
   2354 def within_tol(x, y, atol, rtol):
   2355     with errstate(invalid='ignore'):
-> 2356         return less_equal(abs(x-y), atol + rtol * abs(y))

ValueError: operands could not be broadcast together with shapes (269,449,4) (300,500,4) 
```

must have something to do with canvas size vs the snapshot, perhaps snapshots just need to be regenerated?

clewis7:
nonetheless, sidecar is functional for a single plot as so:

``` python
# create a `Plot` instance
plot = Plot()

# get a grayscale image
data = iio.imread("imageio:camera.png")

# plot the image data
image_graphic = plot.add_image(data=data, name="sample-image")

# show the plot
plot.show(sidecar_kwargs={"title": "sample image", "layout": {'width': '800px'}})
```

image

wanted to get your thoughts before implementing in gridplot and imagewidget

examples/notebooks/simple.ipynb
kushalkolar, line 1: I think let's keep the simple.ipynb simple, without kwargs for sidecar. We can explain it in the docstring

kushalkolar, line 1: why do we have `plot.sidecar.close()` calls, instead of just `plot.close()` calls for the demo nb?

fastplotlib/layouts/_plot.py
```diff
@@ -4,10 +4,12 @@
 import os
 
 import pygfx
+from IPython.display import display
```
kushalkolar, line 7: this should be under `if is_jupyter():`

fastplotlib/layouts/_plot.py
```diff
@@ -117,7 +132,18 @@ def show(
             self.toolbar = ToolBar(self)
             self.toolbar.maintain_aspect_button.value = maintain_aspect
 
-        return VBox([self.canvas, self.toolbar.widget])
+        if not sidecar:
+            return VBox([self.canvas, self.toolbar.widget])
+
+        if self.sidecar is None:
+            # sidecar doesn't like unpacking when kwargs are `None`
```
kushalkolar, line 139: unpacking `None` has nothing to do with sidecar. You cannot unpack a `None` type, `**` is only for unpacking dictionaries :D

fastplotlib/layouts/_plot.py
```diff
@@ -88,6 +96,13 @@ def show(
         toolbar: bool, default True
             show toolbar
 
+        sidecar: bool, default True
+            display the plot in a ``jupyterlab-sidecar``
+
+        sidecar: dict, default None
```
kushalkolar, line 102: sidecar_kwargs

fastplotlib/layouts/_plot.py
```diff
@@ -117,7 +132,18 @@ def show(
...
+
+        if self.sidecar is None:
+            # sidecar doesn't like unpacking when kwargs are `None`
+            if sidecar_kwargs is not None:
+                self.sidecar = Sidecar(**sidecar_kwargs)
+            else:
+                self.sidecar = Sidecar()
+
+        with self.sidecar:
+            return display(VBox([self.canvas, self.toolbar.widget]))
```
kushalkolar, line 146: I'm curious, what happens if you just do `display(...)` without the return?

kushalkolar (changes requested):

kushalkolar:
> With all of the above comments in mind:
> 
> I think it is most relevant to have:
> 
> ```python
> plot.show(sidecar=True, sidecar_kwargs={...})
> ```
> 
> and should function as follows:
> 
> * default for all plots is for it to be displayed in sidecar, `sidecar=True`
> * `plot.close()` will close the sidecar and plot instance
> * clicking the **"x"** will close the sidecar but keep the plot in memory and redisplaying the plot via `plot.show()` will open sidecar with plot again

This sounds good to me! 

Is the 3rd point functional in the current implementation?

examples/notebooks/simple.ipynb
clewis7: because you have a function call to visualize all of the plots together at the end of the notebook...if you close the plot you won't be able to do so

fastplotlib/layouts/_plot.py
clewis7: opens a sidecar, does not put plot in sidecar lol but just displays below cell

clewis7:
still getting error that I mentioned above ab mismatched sizes between canvas and screenshot, even though I regenerated them and ran the tests locally and they passed...but they fail in those cells you have in the notebook

clewis7:
current implementation for `Plot` and `GridPlot` assumes that you will not display a plot in sidecar without the toolbar...

@kushalkolar I think this makes sense? To me, you would only want to not have the toolbar if you are not rendering in Jupyter. I could not imagine wanting to use sidecar and not have the toolbar...thoughts? Can update the logic, but not sure it is necessary. 

**Other:** Should we implement `Sidecar` for `ImageWidget`? Primary concern is there is currently no `close()` method for `ImageWidget` so closing the underlying `GridPlot` keeps the sliders which would prevent closing `Sidecar` properly. Could implement `close()` method and then `ImageWidget` could function more like `Plot` and `GridPlot` do with `Sidecar`...

fastplotlib/layouts/_gridplot.py
```diff
@@ -13,12 +13,16 @@
...
     from ipywidgets import HBox, Layout, Button, ToggleButton, VBox, Dropdown
+    from sidecar import Sidecar
+    from IPython.display import display
 
 from ._utils import make_canvas_and_renderer
 from ._defaults import create_controller
 from ._subplot import Subplot
 from ._record_mixin import RecordMixin
 
+PLOT_OPEN = False
```
kushalkolar, line 24: This cannot be a global because it would then be shared across plot instances. It must be an attribute of the plot instance.

fastplotlib/layouts/_gridplot.py
```diff
@@ -343,7 +362,25 @@ def show(
                 0, 0
             ].camera.maintain_aspect
 
-        return VBox([self.canvas, self.toolbar.widget])
+        if not sidecar:
+            return VBox([self.canvas, self.toolbar.widget])
+
+        # used when plot.show() is being called again but sidecar has been closed via "x" button
+        # need to force new sidecar instance
+        # couldn't figure out how to get access to "close" button in order to add observe method on click
```
kushalkolar, line 370: how do we know if the user has closed the sidecar with the **x** button?

fastplotlib/layouts/_plot.py
```diff
@@ -8,11 +8,15 @@
...
 if is_jupyter():
     from ipywidgets import HBox, Layout, Button, ToggleButton, VBox
+    from sidecar import Sidecar
+    from IPython.display import display
 
 from ._subplot import Subplot
 from ._record_mixin import RecordMixin
 from ..graphics.selectors import PolygonSelector
 
+PLOT_OPEN = False
```
kushalkolar, line 18: cannot be global, see previous

fastplotlib/layouts/_plot.py
kushalkolar: O_O ok lol

kushalkolar (changes requested):

kushalkolar:
> current implementation for `Plot` and `GridPlot` assumes that you will not display a plot in sidecar without the toolbar...
> 
> @kushalkolar I think this makes sense? To me, you would only want to not have the toolbar if you are not rendering in Jupyter. I could not imagine wanting to use sidecar and not have the toolbar...thoughts? Can update the logic, but not sure it is necessary.

Yes this makes sense, I think that if someone wants sidecar but no toolbar that's such a bizarre cornercase. If they don't want toolbars in jupyter in general that's a weird cornercase and they're probably trying to make their own thing, in which case they can customize it themselves

> **Other:** Should we implement `Sidecar` for `ImageWidget`? Primary concern is there is currently no `close()` method for `ImageWidget` so closing the underlying `GridPlot` keeps the sliders which would prevent closing `Sidecar` properly. Could implement `close()` method and then `ImageWidget` could function more like `Plot` and `GridPlot` do with `Sidecar`...

Yes sounds good :D

fastplotlib/layouts/_gridplot.py
clewis7: this would indicate that the **x** but was used because you know that the plot instance is still open but `plot.show()` has been called again which would imply that the sidecar has been closed with the **x** or with `plot.sidecar.close()` and not with `plot.close()`

if the sidecar is still open and you are calling `plot.show()` again (which I don't know why you would) but this would open the plot in a new tab of the existing sidecar as to above where I showed that multiple calls will put the plots side-by-side in the same sidecar which looks weird

kushalkolar: > if the sidecar is still open and you are calling `plot.show()` again (which I don't know why you would) but this would open the plot in a new tab of the existing sidecar as to above where I showed that multiple calls will put the plots side-by-side in the same sidecar which looks weird

Users will always try bizarre things :smile: . I say we go with it, and see what the quirks are when they arise.

examples/notebooks/linear_region_selector.ipynb
kushalkolar, line 1: why `sidecar=False` here?

examples/notebooks/linear_selector.ipynb
kushalkolar, line 1: I think we can make another keyword arg, `vbox=[<list of widgets>]` which if provided it will put a `VBox` that contains the plot canvas and the other widgets passed by the user in the sidecar! So the desired usage is like:

```python
plot.show(vbox=[ipywidget_slider, other_ipywidgets, ...])
```

And it will put the plot in the sidecar, and the other widgets in the sidecar as well in a vbox!

What do you think?

kushalkolar (changes requested):

examples/notebooks/linear_selector.ipynb
kushalkolar: This makes me think, `show()` is then doing two functions: (1) starts the render loop, (2) shows the plot. I think we should now split `show()` into two functions, one that just starts the render loop and another which only shows the plot. `show()` should call something like `start_render()` if the render has not yet started.

clewis7: Hmmmm, also what if you add a widget after rendering and the plot is in the sidecar? 

``` python
plot.add_linear_selector()
```

I think we would need to modify the "add" functions for those widgets to check if `plot.sidecar is not None` and then will know to render those in the sidecar

clewis7: > I think we can make another keyword arg, `vbox=[<list of widgets>]` which if provided it will put a `VBox` that contains the plot canvas and the other widgets passed by the user in the sidecar! So the desired usage is like:
> 
> ```python
> plot.show(vbox=[ipywidget_slider, other_ipywidgets, ...])
> ```
> 
> And it will put the plot in the sidecar, and the other widgets in the sidecar as well in a vbox!
> 
> What do you think?

I think this makes sense, because you might have additional widgets that you have already created and want them rendered with the plot instance in the sidecar. This case works when you know those widgets you want ahead of time...see what I said above for if you don't...

examples/notebooks/linear_region_selector.ipynb
clewis7: was getting a weird testing error, but I mostly think it was because the versions of the notebook that were being tested when I initially pushed were not what I have locally (git does some weird stuff with the notebooks when I switch between branches)

examples/notebooks/linear_selector.ipynb
clewis7: > This makes me think, `show()` is then doing two functions: (1) starts the render loop, (2) shows the plot. I think we should now split `show()` into two functions, one that just starts the render loop and another which only shows the plot. `show()` should call something like `start_render()` if the render has not yet started.

for the line that you left this comment on in the notebook...the reason I had to do `sidecar=False` is because you can't VBox a sidecar...gets wonky lol 

but I think what you said after that about having VBox kwarg is good if you want to group things in sidecar, but if `sidecar=False`...would we still want to use the VBox kwarg to group things in the general cell outputs?

kushalkolar: > Hmmmm, also what if you add a widget after rendering and the plot is in the sidecar?

Anything they want to add has to be added by the `show()` call, that's how it works in regular jupyter outputs anyways :) 
 
> ```python
> plot.add_linear_selector()
> ```

?? Do you mean `line_graphic.add_linear_selector`, this doesn't add a widget anyways. 

> I think we would need to modify the "add" functions for those widgets to check if `plot.sidecar is not None` and then will know to render those in the sidecar

This is not a thing, it doesn't add widgets.



> > This makes me think, `show()` is then doing two functions: (1) starts the render loop, (2) shows the plot. I think we should now split `show()` into two functions, one that just starts the render loop and another which only shows the plot. `show()` should call something like `start_render()` if the render has not yet started.
> 
> for the line that you left this comment on in the notebook...the reason I had to do `sidecar=False` is because you can't VBox a sidecar...gets wonky lol
> 
> but I think what you said after that about having VBox kwarg is good if you want to group things in sidecar, but if `sidecar=False`...would we still want to use the VBox kwarg to group things in the general cell outputs?

I mean to `VBox` everything and then display that `VBox` in a sidecar.

clewis7: oh wow, I think I was having a moment when I wrote this

for some reason I thought there was functionality to add a widget after calling `show()`

kushalkolar: yea this is just an ipywidget thing, it's not possible :)

fastplotlib/layouts/_gridplot.py
```diff
@@ -360,8 +365,18 @@ def show(
                 0, 0
             ].camera.maintain_aspect
 
+        # validate vbox if not None
+        if vbox is not None:
+            for widget in vbox:
+                if not isinstance(widget, Widget):
```
kushalkolar, line 371: I don't think we can use `Widget` or any ipywidget imports directly within `show()` because if `ipywidgets` is not installed in the environment (if they are using fastplotlib with only `glfw` or `Qt`), this will raise a error.

kushalkolar: I would just put it in the `VBox` and let `VBox` deal with it if it's not the right type.

fastplotlib/layouts/_gridplot.py
```diff
@@ -360,8 +365,18 @@ def show(
...
+                    raise ValueError(f"Items in vbox must be ipywidgets. Item: {widget} is of type: {type(widget)}")
+            self.vbox = VBox(vbox)
+
         if not sidecar:
-            return VBox([self.canvas, self.toolbar.widget])
+            if self.vbox is not None:
+                return VBox([self.canvas, self.toolbar.widget, self.vbox])
```
kushalkolar, line 377: from the previous comment, I just realized that we can't use `VBox` in `show()` because it'll raise an error if using in an environment without `ipywidgets`. Perhaps we need to split `show()` into `show_desktop()` and `show_notebook()`?

Or have two subclasses of `PlotFrame`, one for notebook and one for desktop, I think this will be the cleaner solution.

kushalkolar (approved):

kushalkolar:
So this has an issue where it will not work on desktop, but we need to fix that with #310 since this PR has already gotten long.

PR #313, opened 2023-10-03
Plot frame

### LINKED ISSUE #268: Setting the coordinate reference system
When plotting images with `add_image()` it doesn't show up in the conventional orientation, but upside down (same when using `ImageWidget()`.

```
face = np.array([
   [0.6, .5, .5,  0.6, 0.5, 0.5, 0.6,], # row 0
   [0.6, 0.4, .5, 0.6, 0.4, 0.5, 0.6,], # row 0.6
   [0.6, 0.6, 0.6, 0.6, 0.6, 0.6, 0.6,], # row 2
   [0.6, 0.6, 0.6, 0.4, 0.6, 0.6, 0.6,], # row 3
   [0.6, 0.4, 0.6, 0.6, 0.6, 0.4, 0.6,], # row 4
   [0.6, 0.6, 0.4, 0.4, 0.4, 0.6, 0.6,], # row 5
   [0.5, 0.6, 0.6, 0.6, 0.6, 0.6, 0.5,], # row 6
]);
```
Matplotlib  `plt.imshow(face, cmap='gray')` output:

mpl_imshow

Fastplotlib:
```
import fastplotlib as fpl
plot = fpl.Plot()
plot.add_image(data=face, cmap='gray')
plot.show()
```
add_image

Obviously there are conventions at play here with yaxis being switched being the norm for images.

### LINKED ISSUE #310: make a `PlotFrame` class for `Toolbar` and `sidecar` functionality
TODO once #300 is merged

### LINKED ISSUE #342: make `sidecar=False` the default?
@clewis7 @EricThomson your thoughts would be great. Most libraries put plotting outputs right after the cell so maybe we should follow that as the default? It also avoids things like this: https://github.com/fastplotlib/fastplotlib/issues/337

kushalkolar:
closes #310 closes #342 closes #268

This PR adds a `Frame` class that manages how plots are shown. `Plot` and `Gridplot`, and any future flexilbe gridplot, inherit from `Frame`. 

### Toolbars
Refactored ipywidget Toolbar that is common between Plot, Gridplot and any future gridplots

### Output Contexts
These are classes that "hold" the output of a Plot/Gridplot. They determine how the canvas, toolbar, and any additional widgets are shown. When `Plot.show()` is called, they return an instance of one of the following output contexts based on the curent plot canvas.

  - `JupyterOutputContext`
    - Inherits from `ipywidgets.VBox` and contains the canvas, toolbar, and any other widgets. Manages sidecar if desired.
  - `QOutputContext`
    - Inherits from `QWidget`, contains the canvas, qt toolbar, and any other widgets.
  - If the canvas is not jupyter or qt (ex. if it's glfw or offscreen like in our test suite), the output context is just the plot canvas itself. 

This cleans up the messy `show()` calls we previously had everywhere, and cleans up other stuff too. Now the output context manages closing plots too. 

BTW, the `QOutputContext` can also be used in notebooks with if you execute `%gui qt` before creating plots :smile:


clewis7:
#310

kushalkolar:
This is working now with `JupyterOutputContext`! Also fixed the wonky issue with auto-scale and center scene when you had a single controller controlling multiple cameras for diff subplots. 

Now onto Qt output context and toolbar.

kushalkolar:
Everything works on my end, I tested every combination with ipywidgets, sidecar/no sidecar, Qt, and with ImageWidget. 

Right now the toolbar implementations are quite repetitive, we can make that nicer in a future PR after we have some time to think about it. Someone must have made a common interface between ipywidgets and Qt that we can use. 

https://github.com/fastplotlib/fastplotlib/assets/9403332/462d7813-850a-492c-95fc-12c691e2fbb9

fastplotlib/layouts/_frame/_frame.py
```diff
@@ -0,0 +1,184 @@
...
+        if add_widgets is None:
+            add_widgets = list()
+
+        # flip y axis if ImageGraphics are present
+        if hasattr(self, "_subplots"):
+            for subplot in self:
+                for g in subplot.graphics:
+                    if isinstance(g, ImageGraphic):
+                        subplot.camera.local.scale_y = -1
+                        break
```
clewis7, line 144: closes #268

fastplotlib/layouts/_frame/_frame.py
```diff
@@ -0,0 +1,184 @@
...
+        # show was already called, return existing output context
+        if self._output is not None:
+            return self._output
+
+        self.start_render()
+
+        if sidecar_kwargs is None:
+            sidecar_kwargs = dict()
+
+        if add_widgets is None:
```
clewis7, line 135: I assume these are any additional widgets someone would want VBoxed with the plot?

fastplotlib/layouts/_frame/_ipywidget_toolbar.py
```diff
@@ -0,0 +1,297 @@
...
+
+        # set initial values for some buttons
+        self._maintain_aspect_button.value = self.current_subplot.camera.maintain_aspect
+
+        if copysign(1, self.current_subplot.camera.local.scale_y) == -1:
+            self._y_direction_button.icon = "arrow-down"
+        else:
+            self._y_direction_button.icon = "arrow-up"
+
+        HBox.__init__(self, widgets)
```
clewis7, line 137: why this notation?

fastplotlib/layouts/_frame/_ipywidget_toolbar.py
```diff
@@ -0,0 +1,297 @@
...
+)
+
+from ...graphics.selectors import PolygonSelector
+from ._toolbar import ToolBar
+
+
+class IpywidgetToolBar(HBox, ToolBar):
+    def __init__(self, plot):
+        """
+        Basic toolbar for a GridPlot instance.
```
clewis7, line 29: for GridPlot or Plot...correct?

fastplotlib/layouts/_frame/_jupyter_output.py
```diff
@@ -0,0 +1,84 @@
...
+                f"add_widgets must be list of ipywidgets, you have passed:\n{add_widgets}"
+            )
+
+        self.use_sidecar = use_sidecar
+
+        if not make_toolbar:  # just stack canvas and the additional widgets, if any
+            self.output = (frame.canvas, *add_widgets)
+
+        if make_toolbar:  # make toolbar and stack canvas, toolbar, add_widgets
+            self.toolbar = IpywidgetToolBar(frame)
```
clewis7, line 53: what if ImageWidget? where is logic for if plot type is imagewidget?

clewis7:
did you update the requirements file?

fastplotlib/layouts/_frame/_frame.py
clewis7: One thing I notice is that this logic is not implemented for imagewidget

clewis7:
Another thing I see is that when I have two subplots in an imagewidget and the camera y_scales are opposite that when I change click between subplots the flip button is not changing in accordance with the camera y_scale...just stays the same

fastplotlib/layouts/_frame/_frame.py
kushalkolar: ImageWidget uses a Gridplot so it's not necessary

fastplotlib/layouts/_frame/_frame.py
kushalkolar: Yes

fastplotlib/layouts/_frame/_ipywidget_toolbar.py
kushalkolar: Instantiate the subclass, I could do a super instead too

kushalkolar:
> did you update the requirements file?

No dependency changes here.

clewis7:
> > did you update the requirements file?
> 
> No dependency changes here.

not even if someone only wants qt? they would need to have pyqt6 installed right?

kushalkolar:
> > > did you update the requirements file?
> > 
> > 
> > No dependency changes here.
> 
> not even if someone only wants qt? they would need to have pyqt6 installed right?

Qt land is a bit complicated, so they will have to install Qt themselves separately. We can do toggles later, because to do this properly we need to give users the option of PySide etc and decide how broad we want this support.

fastplotlib/layouts/_frame/_ipywidget_toolbar.py
kushalkolar: yes!

fastplotlib/layouts/_frame/_ipywidget_toolbar.py
kushalkolar: changed to super

fastplotlib/layouts/_frame/_jupyter_output.py
kushalkolar: `ImageWidget` just uses the `Gridplot` context:

```python
    def show(self, toolbar: bool = True, sidecar: bool = False, sidecar_kwargs: dict = None):
        """
        Show the widget

        Returns
        -------
        OutputContext
        """
        if self.gridplot.canvas.__class__.__name__ == "JupyterWgpuCanvas":
            self._image_widget_toolbar = IpywidgetImageWidgetToolbar(self)

        elif self.gridplot.canvas.__class__.__name__ == "QWgpuCanvas":
            self._image_widget_toolbar = QToolbarImageWidget(self)

        return self.gridplot.show(
            toolbar=toolbar,
            sidecar=sidecar,
            sidecar_kwargs=sidecar_kwargs,
            add_widgets=[self._image_widget_toolbar]
        )
```

kushalkolar:
> > > Another thing I see is that when I have two subplots in an imagewidget and the camera y_scales are opposite that when I change click between subplots the flip button is not changing in accordance with the camera y_scale...just stays the same
> > 
> > 
> > In ipywidgets or Qt?
> 
> ipwidgets

I think now I see what you mean, it is wonky! So this is because in ImageWidget all cameras are under one controller. I did not decide how to deal with flipping when controllers are diff. What I will do is if the flip button is clicked, it will flip the camera for the current subplot but also any camera that has the same controller (similar to what I did for panzoom and maintain_aspect).

PR #314, opened 2023-10-03
add desktop-only CI, add py312, remove `ipywidgets` import in `plot.py`, pin `pygfx>=0.1.14`

kushalkolar:
1. This adds CI to test that the examples work in an environment without jupyter and ipywidgets, to make sure that the examples work in a "minimal" env with only glfw.
2. Add python 3.12 to CI
3. remove `ipywidgets` import from `layouters/_plot.py`
4. pin `pygfx>=0.1.14`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,10 +22,10 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.9
+    - name: Set up Python 3.11
       uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: 3.11
     - name: Install llvmpipe and lavapipe for offscreen canvas, and git lfs
       run: |
         sudo apt-get update -y -qq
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -36,8 +36,8 @@ jobs:
         sudo apt-get install ./pandoc-3.1.4-1-amd64.deb
     - name: Install dev dependencies
       run: |
-        python -m pip install --upgrade pip
-        # remove pygfx from requirements, we install a specific commit of pygfx since both fpl and pygfx are fast evolving
+        python -m pip install --upgrade pip setuptools
+        # remove pygfx from install_requires, we install using pygfx@main
```
clewis7, line 40: is this the best thing to do? we often fall behind pygfx `main` means tests could be failing in the event PRs are made before we catch-up

or is this to force us to be caught up always lol

clewis7 (commented):
otherwise LGTM

.github/workflows/ci.yml
kushalkolar: yea this way we catch things before they sneak up on us :D

PR #320, opened 2023-10-13
allow getting and setting cmaps in image widget

kushalkolar:
Usage:

```python
iw.cmap

# returns list of cmaps, one item per subplot/managed graphic in image widget

# set specific subplot managed graphics
iw.cmap = ["gray", "gnuplot2"]

# set all
iw.cmap = "viridis"
```

After #166 is done we can make it easier to set cmaps in individual subplots for image graphics, but I think this will still be useful for image widget ever after we have that.

### DIFF
--- a/fastplotlib/widgets/image.py
+++ b/fastplotlib/widgets/image.py
@@ -114,6 +114,34 @@ class ImageWidget:
         return iw_managed
 
     @property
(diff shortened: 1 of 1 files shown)


clewis7 (commented):
LGTM

PR #321, opened 2023-10-14
add `in` operator to `PlotArea` base

kushalkolar:
allows:

```python
"graphic_name" in plot

"graphic_name" in gridplot[0, 0]

graphic_obj in plot

graphic_obj in gridplot[0, 0]
```

usecase example from mesviz, makes it easy to check if a subplot has contours or not, etc. 

image


clewis7 (commented):
LGTM

PR #324, opened 2023-10-14
remove jupyter check for ImageWidget

kushalkolar:
For some reason this check doesn't work if the ImageWidget is "spawned" within a `@interact_manual` in jupyter, this will also break it in the mesviz voila app.

Need to find a more reliable way to check if running in jupyter that also works in voila.


kushalkolar:
@clewis7 just FYI

PR #330, opened 2023-10-21
add codecov 

kushalkolar:
so we get test coverage report with all PRs


kushalkolar:
Looks like runpy is currently broken, so this won't be possible to integrate until the fix gets into python, which might not be until v3.13 or something: 
https://github.com/python/cpython/issues/99437

PR #333, opened 2023-10-25
more polygon selector stuff

kushalkolar:
WIP

- [x] move vertices
- [ ] add vertices
- [ ] delete segments
- [ ] mesh for fill area
  - [ ] move entire polygon
- [ ] make polygon from points
- [ ] make the lines all have z-position under the vertex points
- [ ] vertex0 from the last segment is a bit wonky, I think 2 last segments are being added 

Might need to look into Almar's mesh morphing for changing the fill area


kushalkolar:
One issue with Delaunay triangulation is that it only seems to produce convex hulls which is a problem for shapes like this. Will need to look into QHull options

image

kushalkolar:
Will look into this: https://github.com/bellockk/alphashape

PR #335, opened 2023-10-25
refactor text graphic

clewis7:
#305

Refactored the text graphic to use properties and some more features of `pygfx TextGeometry`

### DIFF
--- a/docs/source/api/graphics/TextGraphic.rst
+++ b/docs/source/api/graphics/TextGraphic.rst
@@ -29,7 +29,6 @@ Properties
     TextGraphic.position_y
     TextGraphic.position_z
     TextGraphic.text
-    TextGraphic.text_position
     TextGraphic.text_size
     TextGraphic.visible
     TextGraphic.world_object
--- a/docs/source/api/graphics/TextGraphic.rst
+++ b/docs/source/api/graphics/TextGraphic.rst
@@ -21,10 +21,16 @@ Properties
     :toctree: TextGraphic_api
 
     TextGraphic.children
+    TextGraphic.face_color
+    TextGraphic.outline_color
+    TextGraphic.outline_size
     TextGraphic.position
     TextGraphic.position_x
     TextGraphic.position_y
     TextGraphic.position_z
+    TextGraphic.text
+    TextGraphic.text_position
+    TextGraphic.text_size
     TextGraphic.visible
     TextGraphic.world_object
 
@@ -33,10 +39,4 @@ Methods
 .. autosummary::
     :toctree: TextGraphic_api
 
-    TextGraphic.update_face_color
-    TextGraphic.update_outline_color
-    TextGraphic.update_outline_size
-    TextGraphic.update_position
-    TextGraphic.update_size
-    TextGraphic.update_text
 
--- a/fastplotlib/graphics/text.py
+++ b/fastplotlib/graphics/text.py
@@ -145,12 +145,3 @@ class TextGraphic(Graphic):
 
         self.world_object.material.outline_color = color
 
-    @property
-    def text_position(self):
-        """Returns the position of this graphic."""
-        return self.world_object.local.position
-
-    @text_position.setter
-    def text_position(self, pos: Tuple[int, int, int]):
-        """Set the position of this graphic."""
-        self.world_object.local.position = pos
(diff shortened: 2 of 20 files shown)


clewis7:
Screen Shot 2023-10-25 at 3 20 09 PM

fastplotlib/graphics/text.py
```diff
@@ -43,11 +44,16 @@ def __init__(
...
 
         """
+        super(TextGraphic, self).__init__(*args, **kwargs)
 
-        super(TextGraphic, self).__init__(name=name)
+        self._text = text
 
         world_object = pygfx.Text(
-            pygfx.TextGeometry(text=str(text), font_size=size, screen_space=False),
+            pygfx.TextGeometry(text=str(text),
```
kushalkolar, line 52: can you use `black` for now

fastplotlib/graphics/text.py
```diff
@@ -43,11 +44,16 @@ def __init__(
...
+                               font_size=size,
+                               screen_space=True,
```
kushalkolar, line 54: make a kwarg for this?

fastplotlib/graphics/text.py
```diff
@@ -43,11 +44,16 @@ def __init__(
...
+                               anchor="middle-center"
```
kushalkolar, line 55: make a kwarg for this

fastplotlib/graphics/text.py
```diff
@@ -43,11 +44,16 @@ def __init__(
...
+                               ),
             pygfx.TextMaterial(
```
kushalkolar, line 57: any other useful kwargs to have for material?

fastplotlib/graphics/text.py
```diff
@@ -59,22 +65,78 @@ def __init__(
...
-    def update_face_color(self, color: Union[str, np.ndarray]):
+    @property
+    def face_color(self):
+        """Returns the face color of this graphic."""
+        return self.world_object.material.color
+
+    @face_color.setter
+    def face_color(self, color: Union[str, np.ndarray]):
+        """Set the face color of this graphic."""
+        if not (isinstance(color, str) or isinstance(color, np.ndarray)):
```
kushalkolar, line 103: I think the text material color should be able to take anything in the form of `pygfx.Color`, so just convert the input color to `pygfx.Color` and use that?

fastplotlib/graphics/text.py
```diff
@@ -59,22 +65,78 @@ def __init__(
...
         self.world_object.material.outline_thickness = size
 
-    def update_outline_color(self, color: Union[str, np.ndarray]):
+    @property
+    def outline_color(self):
+        """Returns the outline color of this graphic."""
+        return self.world_object.material.outline_color
+
+    @outline_color.setter
+    def outline_color(self, color: Union[str, np.ndarray]):
```
kushalkolar, line 127: I think the text material color should be able to take anything in the form of pygfx.Color, so just convert the input color to pygfx.Color and use that?

kushalkolar (changes requested):

fastplotlib/graphics/text.py
clewis7: no, I don't think so...not anything that should be exposed to a user

kushalkolar:
can you run black on only the textgraphic, else this will cause merge conflicts with every other WIP PR. 

we should look into ruff longterm

docs/source/api/graphics/TextGraphic.rst
kushalkolar, line 1: the api docs need to be auto-generated by the `generate_api.py` script

kushalkolar (changes requested):

kushalkolar:
seems like the wrong commit got reverted, this is still changing the rest of the codebase

fastplotlib/graphics/text.py
```diff
@@ -59,22 +77,80 @@ def __init__(
...
+        """Set the outline color of this graphic"""
+        if not (isinstance(color, str) or isinstance(color, np.ndarray)):
+            raise ValueError("Outline color must be of type str or np.ndarray")
+
         self.world_object.material.outline_color = color
 
-    def update_position(self, pos: Tuple[int, int, int]):
-        self.world_object.position.set(*pos)
+    @property
+    def text_position(self):
```
kushalkolar, line 149: These aren't necessary, it inherits from Graphic which has position properties

kushalkolar (changes requested):
Also regenerate the add graphics mixin

kushalkolar (approved):

PR #338, opened 2023-10-26
example desktop video scrubber using linear selector

joeyo:
I wanted to scrub through videos without needing to use jupyter/ipywidgets and figured out a way to do it using a `linear_selector`. Adding as an example in case it's useful.

### DIFF
--- a/examples/desktop/image/image_scrub_linear_selector.py
+++ b/examples/desktop/image/image_scrub_linear_selector.py
@@ -1,0 +1,29 @@
+import fastplotlib as fpl
+import imageio.v3 as iio
+
+
+movie = iio.imread("imageio:cockatoo.mp4", plugin="pyav", index=None)
+
+num_frames = movie.shape[0]
+movie_width = movie.shape[2]
+
+plot = fpl.Plot()
+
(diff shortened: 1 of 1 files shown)


kushalkolar:
Thanks for the example! This will put a linear selector on top of the image.

We want to have in-canvas UI elements eventually, I'm planning to get started on this soon unless someone wants to try it out. See https://github.com/pyimgui/pyimgui/issues/336

kushalkolar:
You could also embed it in a Qt app and use Qt for your UI.

joeyo:
> You could also embed it in a Qt app and use Qt for your UI.

I actually initially started going down that route but could not get FPL master to work with PyQT/PySide at all.

kushalkolar:
What errors were you getting, happy to take a look if you post an issue. I
think the Qt canvas is just a QWidget. You can also force the Qt canvas by
doing `plot = Plot(canvas="Qt")`. Wgpu sometimes selects glfw by default, I
think they fixed it recently.

joeyo:
Okay, following-up here: https://github.com/fastplotlib/fastplotlib/issues/340

PR #344, opened 2023-10-27
histogram lut widget

### LINKED ISSUE #166: Histogram vmin vmax widget

kushalkolar:
closes #166

Also moved animation and render methods from `Subplot` to `PlotBase`, because we need `DockedPlotArea` to use animation functions which is required for `LinearRegionSelector` plot hooks.

https://github.com/fastplotlib/fastplotlib/assets/9403332/c7b10c5d-dc02-48e2-9b08-b664862a4aa4

remaining todo:
- [x] integrate with image widget, removing existing ipywidget sliders
- [x] bidirectional events with associated `ImageGraphic`
- [x] garbage collection, make sure events are cleaned up if switching ImageGraphic too
- ~text to display min and max next to the lines?~ Will do in another PR since text was refactored 
- [x] allow adjusting flank size
- [x] prevent `LinearRegionSelector` from re-enabling the controller when used in docks

For now this is not easily exposed by any imports and only used by `ImageWidget`. We can expose after it stabilizes more and if we're happy with keeping this in `fastplotlib/widgets`.

### DIFF
--- a/fastplotlib/layouts/_base.py
+++ b/fastplotlib/layouts/_base.py
@@ -1,5 +1,7 @@
+from inspect import getfullargspec
 from typing import *
 import weakref
+from warnings import warn
 
 import numpy as np
 
@@ -91,6 +93,9 @@ class PlotArea:
         self.controller.register_events(
             self.viewport,
         )
+
+        self._animate_funcs_pre = list()
+        self._animate_funcs_post = list()
 
         self.renderer.add_event_handler(self.set_viewport_rect, "resize")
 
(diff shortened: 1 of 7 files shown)


fastplotlib/layouts/_base.py
```diff
@@ -92,6 +94,9 @@ def __init__(
             self.viewport,
         )
 
+        self._animate_funcs_pre = list()
+        self._animate_funcs_post = list()
```
clewis7, line 98: add comment for what these are for

fastplotlib/layouts/_base.py
```diff
@@ -224,12 +229,90 @@ def set_viewport_rect(self, *args):
...
+        """
+        for f in funcs:
+            if not callable(f):
+                raise TypeError(
+                    f"all positional arguments to add_animations() must be callable types, you have passed a: {type(f)}"
+                )
+            if pre_render:
+                self._animate_funcs_pre += funcs
+            if post_render:
+                self._animate_funcs_post += funcs
```
clewis7, line 291: edge case where user sets `pre_render` and `post_render` to true...would this call issues if the animation function would then be getting called twice?

fastplotlib/widgets/histogram_lut.py
```diff
@@ -0,0 +1,148 @@
...
+            axis="y",
+            edge_thickness=8
+        )
+
+        widget_wo = Group()
+        widget_wo.add(self.line.world_object, self.linear_region.world_object)
+
+        self._set_world_object(widget_wo)
+
+        self.world_object.local.scale_x *= -1
```
clewis7, line 49: is this for orientation of the histogram widget? I would maybe add comment

fastplotlib/widgets/histogram_lut.py
```diff
@@ -0,0 +1,148 @@
...
+    def _set_selection_from_cmap(self, ev):
+        vmin, vmax = ev.pick_info["vmin"], ev.pick_info["vmax"]
+
+        self.linear_region.selection.block_events(True)
+
+        self.linear_region.selection = (vmin, vmax)
+
+        self.linear_region.selection.block_events(False)
+
+    def set_data(self, data):
```
clewis7, line 120: what is this method for? doc strings would be helpful :D

fastplotlib/layouts/_base.py
kushalkolar: It wouldn't be an issue, it'd just get called twice. We could change to a kwarg that takes in a string of "pre" or "post"

fastplotlib/widgets/histogram_lut.py
kushalkolar: So that the bins are left facing

fastplotlib/widgets/histogram_lut.py
kushalkolar: Yes I still haven't finalized the design of this yet so no docstrings

PR #351, opened 2023-10-31
make hlut widget optional

clewis7:


clewis7:
would be nice to add an hlut button to the toolbar so it can be generated after imagewidget instantiation if a user wanted one for a specific subplot!

fastplotlib/widgets/image.py
```diff
@@ -288,6 +289,9 @@ def __init__(
         names: Optional[str]
             gives names to the subplots
 
+        histogram_widget: bool, default False
+            determines whether each subplot is rendered with a Histogram LUT widget
```
kushalkolar, line 293: More concise: make histogram LUT widget for each image?

kushalkolar (changes requested):

kushalkolar:
> would be nice to add an hlut button to the toolbar so it can be generated after imagewidget instantiation if a user wanted one for a specific subplot!

I think this is feature creep, we they can create and add it programmatically if they need that level of customization?

kushalkolar:
Also add kwarg to specify subsample dims for HistogramLut widget?

fastplotlib/widgets/image.py
```diff
@@ -225,6 +225,7 @@ def __init__(
         grid_shape: Tuple[int, int] = None,
         names: List[str] = None,
         grid_plot_kwargs: dict = None,
+        histogram_widget: bool = False,
```
kushalkolar, line 228: I think default should be `True` :D

clewis7, line 228:
```suggestion
        histogram_widget: bool = True,
```

fastplotlib/widgets/image.py
clewis7, line 293:
```suggestion
            make histogram LUT widget for each subplot
```

kushalkolar:
> > Also add kwarg to specify subsample dims for HistogramLut widget?
> 
> @kushalkolar anything more than number of bins? or you want a dict that allows user to pass nbins and flank?

Just a tuple of its to specify the max size?

kushalkolar:
going to merge this, see #359 for the subsample

PR #280, opened 2023-07-07
More controllers

kushalkolar:
- adds the new controllers from `pygfx`:
  - panzoom
  - fly (new)
  - orbit
  - trackball (new)

works properly in simple `Plot`, needs more thought for `GridPlot`, especially considering #267

PR #382, opened 2023-11-26
cameras and controllers refactor

### LINKED ISSUE #381: Allow changing the cameras and controllers dynamically
Allow changing the cameras and controllers dynamically and make this accessible from the toolbar

### LINKED ISSUE #383: Use PerspectiveCamera for everything
Technically the pygfx Orthographic camera is just the perspective camera with a fixed FOV of 0. https://github.com/pygfx/pygfx/blob/main/pygfx/cameras/_orthographic.py#L28

Therefore if we just use the perspective camera for everything it makes it much easier to switch controllers without making things wonky!

kushalkolar:
superseeds #280 #267
closes #381, closes #383

`PlotArea` allows changing the controller types, so all plots and subplots can have their controllers changed at any time. In gridplots, if the same controller is used for multiple subplots then it will change the controller for the other respective subplots too if it's changed for only one of the subplots within that synced-subset. Slightly inelegant implementation, we can make it better later.

Merged `layouts._defaults` into `layouts._utils.`

Adds the new controllers from `pygfx`:
- panzoom
- fly (new)
- orbit
- trackball (new)

Gridplot handles controllers different now. I think the implementation isn't too complex, I realized we have no reason to have consecutive integers for `controller_ids` so that makes life easier! @clewis7 @ArjunPutcha please let me know asap!!

```
controller_types: str, list or np.ndarray, optional
            list or array that specifies the controller type for each subplot.

        controller_ids: str, list or np.ndarray of int or str ids, optional
            | If `None` a unique controller is created for each subplot
            | If "sync" all the subplots use the same controller
            | If ``numpy.array``, its shape must be the same as ``grid_shape``.

            This allows custom assignment of controllers

            | Example with integers:
            | sync first 2 plots, and sync last 2 plots: [[0, 0, 1], [2, 3, 3]]
            | Example with str subplot names:
            | list of lists of subplot names, each sublist is synced: [[subplot_a, subplot_b], [subplot_f, subplot_c]]
            | this syncs subplot_a and subplot_b together; syncs subplot_e and subplot_c together
```


fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
             This allows custom assignment of controllers
 
-            | Example:
-            | unique controllers for a 2x2 gridplot: np.array([[0, 1], [2, 3]])
-            | same controllers for first 2 plots and last 2 plots: np.array([[0, 0, 1], [2, 3, 3]])
+            | Example with integers:
+            | sync first 2 plots, and sync last 2 plots: [[0, 0, 1], [2, 3, 3]]
+            | Example with str subplot names:
+            | list of lists of subplot names, each sublist is synced: [[subplot_a, subplot_b], [subplot_f, subplot_c]]
+            | this syncs subplot_a and subplot_b together; syncs subplot_e and subplot_c together
```
clewis7, line 68:
```suggestion
            | this syncs subplot_a and subplot_b together; syncs subplot_f and subplot_c together
```

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+        # create the cameras
+        self._cameras = np.empty(self.shape, dtype=object)
+        for i, j in product(range(self.shape[0]), range(self.shape[1])):
+            self._cameras[i, j] = create_camera(camera_type=cameras[i, j])
+
+        if controller_ids is None:
+            # individual controller for each subplot
+            controller_ids = np.arange(self.shape[0] * self.shape[1]).reshape(self.shape)
+
+        elif isinstance(controller_ids, str):
```
clewis7, line 116: Are there any other options besides `"sync"`...if not should we throwing an error if `str` passed in not valid?

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+            if controller_ids == "sync":
+                controller_ids = np.zeros(self.shape, dtype=int)
+
+        # list controller_ids
+        elif isinstance(controller_ids, (list, np.ndarray)):
+            flat = list(chain(*controller_ids))
```
clewis7, line 122: better name here instead of `flat`

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+                ids_init = np.arange(self.shape[0] * self.shape[1]).reshape(self.shape)
 
-        cameras = to_array(cameras)
+                # set id based on subplot position for each synced sublist
+                for i, sublist in enumerate(controller_ids):
+                    for name in sublist:
+                        ids_init[self.names == name] = -(i + 1)  # use negative numbers because why not
 
-        self._controllers = np.empty(shape=cameras.shape, dtype=object)
+                controller_ids = ids_init
```
clewis7, line 148: should `controller_ids` be user-facing as a read only attr?

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+
+        if controller_ids.shape != self.shape:
+            raise ValueError("Number of controller_ids does not match the number of subplots")
+
+        if controller_types is None:
+            # `create_controller()` will auto-determine controller for each subplot based on defaults
+            controller_types = np.array(["default"] * self.shape[0] * self.shape[1]).reshape(self.shape)
+
+        # validate controller types
+        flat = list(chain(*controller_types))
```
clewis7, line 168: better name here instead of `flat`

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+            if controller_type is None:
+                continue
+
+            if (controller_type not in valid_str) and (not isinstance(controller_type, valid_instances)):
+                raise ValueError(
+                    f"You have passed an invalid controller type, valid controller_types arguments are:\n"
+                    f"{valid_str} or instances of {[c.__name__ for c in valid_instances]}"
+                )
+
+        controller_types = to_array(controller_types).reshape(self.shape)
```
clewis7, line 184: should this be user-facing as read only?

fastplotlib/layouts/_gridplot.py
```diff
@@ -25,128 +25,197 @@ def to_array(a) -> np.ndarray:
...
+                    "All controllers with the same id must use the same type of controller."
+                )
+
+            cont_type = cont_type[0]
+
+            # get all the cameras that use this controller
+            cams = self._cameras[controller_ids == cid].ravel()
+
+            if cont_type == "default":
+                # hacky fix for now because of hwo `create_controller()` works
```
clewis7, line 202:
```suggestion
                # hacky fix for now because of how `create_controller()` works
```

fastplotlib/layouts/_subplot.py
```diff
@@ -27,12 +18,11 @@ def __init__(
         parent: Any = None,
         position: Tuple[int, int] = None,
         parent_dims: Tuple[int, int] = None,
-        camera: str = "2d",
-        controller: Union[PanZoomController, OrbitController] = None,
-        canvas: Union[str, WgpuCanvas, Texture] = None,
-        renderer: WgpuRenderer = None,
+        camera: Union[str, pygfx.PerspectiveCamera] = "2d",
+        controller: Union[str, pygfx.Controller] = None,
+        canvas: Union[str, WgpuCanvas, pygfx.Texture] = None,
```
clewis7, line 23: how can a texture be a canvas?

fastplotlib/layouts/_gridplot.py
kushalkolar: thanks for catching this! :D

fastplotlib/layouts/_gridplot.py
kushalkolar: yes you're right, we can have more extensive checks

fastplotlib/layouts/_gridplot.py
kushalkolar: I was thinking about this. I'm undecided whether Gridplots should allow these to be accessible, because then the user could try to change one of these controllers in the object array, but then it won't change the corresponding controller_id in the subplot(s). Right now the architecture is that the subplots _own_ the controllers and camera. Gridplot is mostly a way to create subplots. I was thinking of leaving the decision for when we make more comprehensive/flexible gridplots.

fastplotlib/layouts/_gridplot.py
kushalkolar: Similar to discussion above, right now the subplots own the controllers.

fastplotlib/layouts/_subplot.py
kushalkolar: See docstring for the `pygfx.WgpuRenderer`: https://pygfx.readthedocs.io/en/stable/_autosummary/renderers/pygfx.renderers.WgpuRenderer.html#pygfx.renderers.WgpuRenderer

Basically you can render a rasterized scene to a Texture :smile: . A texture is just a way to represent a 2D image. I think that's part of how the remote frame buffer works, _the canvas is a texture_ .

Issue #388, opened 2023-12-04
Tests: OSError: [Errno 24] Too many open files


tlambert03:
There appears to be a leaked file handle somewhere in the tests

- cloned repo
- `pip install -e ".[notebook,docs,tests]"`
- ran `pytest examples`

```
OSError: [Errno 24] Too many open files: '/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/pygments/formatters/terminal.py'
```

<details>

```
module = PosixPath('/Users/talley/dev/forks/fastplotlib/examples/desktop/gridplot/gridplot.py'), force_offscreen = None

Traceback (most recent call last):
  File "/Users/talley/miniforge3/envs/fastplotlib/bin/pytest", line 8, in <module>
    sys.exit(console_main())
             ^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/_pytest/config/__init__.py", line 192, in console_main
    code = main()
           ^^^^^^
  File "/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/_pytest/config/__init__.py", line 169, in main
    ret: Union[ExitCode, int] = config.hook.pytest_cmdline_main(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/pluggy/_hooks.py", line 493, in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/pluggy/_manager.py", line 115, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/fastplotlib/lib/python3.11/site-packages/pluggy/_callers.py", line 113, in _multicall
    raise exception.with_traceback(exception.__traceback__)
[...]

PR #390, opened 2023-12-04
use cmap library for colormaps

### LINKED ISSUE #387: consider cmap library?
Hey @kushalkolar et al,

great package!  i was just looking around the image utils here and noticed that you [require matplotlib](https://github.com/fastplotlib/fastplotlib/blob/2873a677462551f47881e4c6dad637165f5230f6/fastplotlib/utils/generate_colormaps.py#L2) to generate colormaps but, understandably, [don't include matplotlib in your dependencies](https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib/utils/functions.py#L57C1-L58).

when working on napari and other related image libs, I also wanted to use colormaps without having to depend on matplotlib, and made [cmap](https://cmap-docs.readthedocs.io/en/latest/), a lib that only depends on numpy and provides all the matplotlib colormaps (and [many more](https://cmap-docs.readthedocs.io/en/latest/catalog/)) and basic mapping functionality.  the lib exports to many [third-party colormap objects](https://cmap-docs.readthedocs.io/en/latest/colormaps/#usage-with-external-visualization-libraries).  Would you be interested in looking at a PR that uses cmap here?

tlambert03:
this PR is an example of using https://github.com/tlambert03/cmap instead of matplotlib for colormaps.

It does add cmap to `setup.py` ... but if you'd prefer not to go that far, we could guard it all behind an import and say "must install `cmap` to display stuff with colormaps"

closes #387


kushalkolar:
Seems like there are slight color differences between the cmap library and matplotlib ones? You can download the screenshot diffs at the bottom: https://github.com/fastplotlib/fastplotlib/actions/runs/7090258613

kushalkolar:
Seems like there are slight color differences between the cmap library and matplotlib ones? You can download the screenshot diffs at the bottom: https://github.com/fastplotlib/fastplotlib/actions/runs/7090258613

I don't think it's an issue, but I'm curious why it's different.

tlambert03:
thanks for the heads up @kushalkolar.  I'll have a look at it and figure out what's going on.

I do actually test for parity with matplotlib [here](https://github.com/tlambert03/cmap/blob/b3b64db1c17753809d156194670789aa70603073/tests/test_data.py#L60-L79) (both naming and the actual effect of applying the cmap to an image), so i suspect that the differences might be in the way that colormaps are actually applied in the `functions` module.  I did notice a very slight difference in the values of `get_cmap` as well, so I suspect that might be the root of it.  I'll do some digging and get back to you :)

tlambert03:
btw, I'd like to run these tests locally, but i ran into https://github.com/fastplotlib/fastplotlib/issues/388 ... any thoughts there?

kushalkolar:
Thanks! 

Everything in the regenerate screenshots workflow seems fine visually, so you can replace the ground truth of the failing tests with those.

fastplotlib/utils/functions.py
```diff
@@ -1,31 +1,13 @@
...
-    "Set3",
-    "tab10",
-    "tab20",
-    "tab20b",
-    "tab20c",
-]
-
-
-def get_cmap(name: str, alpha: float = 1.0) -> np.ndarray:
+def get_cmap(name: str, alpha: float = 1.0, gamma: float = 1.0) -> np.ndarray:
```
kushalkolar, line 10: Add gamma to docstring

kushalkolar (changes requested):
Other than that docstring lgtm! The colormap text files can now be removed too.

Curious what the difference was previously when the colors weren't matching?

tlambert03:
> Curious what the difference was previously when the colors weren't matching?

oh good.  it's working on CI too! 
Yeah, the difference is actually a good one for you to know about:

the `cmap.Colormap` object behaves very much like a `matplotlib.colors.Colormap`, with the main API being the `__call__` method, which takes an array of scalars and returns an array of colors.  And, just like the matplotlib API, if the input array is of dtype `float`, it expects values from 0-1, and it's of dtype `int`, it expects 0-255. 

```python
In [27]: import cmap

In [28]: viridis = cmap.Colormap('viridis')

In [29]: viridis(np.linspace(0, 255, 5, dtype=int))
Out[29]:

array([[0.267004, 0.004874, 0.329415, 1.      ],
       [0.231674, 0.318106, 0.544834, 1.      ],
       [0.128729, 0.563265, 0.551229, 1.      ],
       [0.360741, 0.785964, 0.387814, 1.      ],
       [0.993248, 0.906157, 0.143936, 1.      ]])

In [31]: from matplotlib import colormaps

In [32]: mpl_viridis = colormaps['viridis']

In [33]: mpl_viridis(np.linspace(0, 255, 5, dtype=int))
Out[33]:

array([[0.267004, 0.004874, 0.329415, 1.      ],
       [0.231674, 0.318106, 0.544834, 1.      ],
       [0.128729, 0.563265, 0.551229, 1.      ],
       [0.360741, 0.785964, 0.387814, 1.      ],
       [0.993248, 0.906157, 0.143936, 1.      ]])

```

 There's of course a subtle difference in the input array there, for N evenly spaced numbers:

```python
In [35]: np.linspace(0, 1, 5, dtype=float)
Out[35]: array([0.  , 0.25, 0.5 , 0.75, 1.  ])

In [36]: np.linspace(0, 255, 5, dtype=int)/255
Out[36]: array([0.        , 0.24705882, 0.49803922, 0.74901961, 1.        ])
```

and

```python
In [29]: viridis(np.linspace(0, 255, 5, dtype=int))
Out[29]:

array([[0.267004, 0.004874, 0.329415, 1.      ],
       [0.231674, 0.318106, 0.544834, 1.      ],
       [0.128729, 0.563265, 0.551229, 1.      ],
       [0.360741, 0.785964, 0.387814, 1.      ],
       [0.993248, 0.906157, 0.143936, 1.      ]])

In [30]: viridis(np.linspace(0, 1, 5, dtype=float))
Out[30]:

array([[0.267004, 0.004874, 0.329415, 1.      ],
       [0.229739, 0.322361, 0.545706, 1.      ],
       [0.127568, 0.566949, 0.550556, 1.      ],
       [0.369214, 0.788888, 0.382914, 1.      ],
       [0.993248, 0.906157, 0.143936, 1.      ]])
```

... so, in my original PR, I used a slightly lower level construct to retrieve the N evenly spaced colors.  Rather than using `Colormap.__call__(some_array)`, I used `Colormap.lut(n_colors)` ... which was slightly different than the existing code you had here which used `np.take` on a `linspace` array of `int`:

```python
In [37]: viridis.lut(5)
Out[37]:

array([[0.267004  , 0.004874  , 0.329415  , 1.        ],
       [0.23022275, 0.32129725, 0.545488  , 1.        ],
       [0.1281485 , 0.565107  , 0.5508925 , 1.        ],
       [0.36285925, 0.786695  , 0.386589  , 1.        ],
       [0.993248  , 0.906157  , 0.143936  , 1.        ]])
```


so I switched it to the `__call__` approach and then everything matched the current screenshots.  probably visually imperceptible, but might as well keep the same behavior :)  Don't hesitate to ask if you have any questions about the cmap object.  But in general, the `__call__` method was designed (and tested) to behave exactly like matplotlibs.

fastplotlib/utils/functions.py
tlambert03: fixed

tlambert03:
> The colormap text files can now be removed too.

oh wow... to be honest, I somehow missed the distinction there that matplotlib was only being used as a fallback.

So, a question for you: I can *definitely* understand wanting to keep dependencies to the absolute minimum.  So, would you like me to update this PR slightly to keep the current behavior for named colormaps that you already bundle, remove cmap from `setup.py` and fallback to cmap only in the case that a string name is unrecognized?  That would still allow you to support custom colormaps using all the previously discussed syntax, but the user would have to opt-in to that behavior.  Thoughts?

kushalkolar:
> There's of course a subtle difference in the input array there, for N evenly spaced numbers:
> 
> ```python
> In [35]: np.linspace(0, 1, 5, dtype=float)
> Out[35]: array([0.  , 0.25, 0.5 , 0.75, 1.  ])
> 
> In [36]: np.linspace(0, 255, 5, dtype=int)/255
> Out[36]: array([0.        , 0.24705882, 0.49803922, 0.74901961, 1.        ])
> ```

That is a subtle difference, thanks for tracking it down! I think the reason we used the approach that we did, and use `int`, is because it makes it easier to work with qualitative colormaps when you give `n_colors`. (not entirely sure, I wrote that part of the code years ago in a previous library)

> So, a question for you: I can _definitely_ understand wanting to keep dependencies to the absolute minimum. So, would you like me to update this PR slightly to keep the current behavior for named colormaps that you already bundle, remove cmap from `setup.py` and fallback to cmap only in the case that a string name is unrecognized? That would still allow you to support custom colormaps using all the previously discussed syntax, but the user would have to opt-in to that behavior. Thoughts?

Since `cmap` only depends on `numpy` I'm fine with keeping it in, if there are issues down the road I would rather vendor it in, but I hope you're able to keep producing all the cool things you've been doing in the imaging ecosystem! :)

I like that this will now enable us to do things like `line.cmap = ["blue", "green", "yellow"]` to create custom arbitrary colormaps. Speaking of which, would you be interested in implementing that? No worries if you don't have time, I can easily do it in another PR :smile:

kushalkolar:
One more thing, does the `cmap` lib allow centering diverging colormaps arbitrarily? 

For example if the following values exist and you want to center the `seismic` colormap around `0`, instead of the middle value here which would be `-2`? 

```python
[-6, -5, -4, -3, -2, -1, 0, 1, 2]
```

tlambert03:
> I like that this will now enable us to do things like line.cmap = ["blue", "green", "yellow"] to create custom arbitrary colormaps.

I just changed `examples/notebooks/lines_cmap.ipynb` to the following code:

```python
plot.graphics[0].cmap = ["blue", "green", "yellow", "#FA499E", (0, 0.2, 0.99)]
```

and it seemed to work just fine:

Screenshot 2023-12-08 at 10 47 32 AM


... was there something else you were expecting would be needed?

kushalkolar:
This example: https://github.com/fastplotlib/fastplotlib/blob/main/examples/desktop/line_collection/line_stack.py

If you have `pyqt6` installed you can use the `%gui qt` hook. Otherwise you can force the `glfw` canvas with `plot = fpl.Plot(canvas="glfw")`

tlambert03:
thanks.  Yeah that example works, but it does get more confusing due to the overloading of the argument as a colormap or list of colormaps.  So, one has to be rather careful there.  If I do this:

```python
# line stack takes all the same arguments as line collection and behaves similarly
cmaps = ["jet"] * len(data)
cmaps[-2:] = ["cubehelix", ["red", "orange", "yellow", "green", "blue", "indigo", "violet"]]
plot.add_line_stack(data, cmap=cmaps)
```

then it works and I see this:
Screenshot 2023-12-08 at 1 29 10 PM

... but it's rather easy to get into trouble there

kushalkolar:
Yes you are right, because then this will raise: 

```python
plot.add_line_stack(data, cmap=["red", "blue", "green"])`
```

But if a line collection is already created, then this will work:

```python
plot.graphics[0].cmap = ["red", "blue", "green"]
```

```python
File ~/repos/fastplotlib/fastplotlib/graphics/line_collection.py:125, in LineCollection.__init__(self, data, z_position, thickness, colors, alpha, cmap, cmap_values, name, metadata, *args, **kwargs)
    123 elif isinstance(cmap, (tuple, list)):
    124     if len(cmap) != len(data):
--> 125         raise ValueError(
    126             "cmap argument must be a single cmap or a list of cmaps "
    127             "with the same length as the data"
    128         )
    129     single_color = False
    130 else:

ValueError: cmap argument must be a single cmap or a list of cmaps with the same length as the data
```

Question: We could enforce that the `cmap` kwarg should represent a single cmap. If the user wants to pass a list of cmaps we make a new kwarg called `cmaps`? 

After the line collection has been created, this still works to change cmaps of individual lines:

```python
plot.graphics[0][2:5].cmap = ["blue", "yellow", "green"]
```

tlambert03:
> Question: We could enforce that the cmap kwarg should represent a single cmap. If the user wants to pass a list of cmaps we make a new kwarg called cmaps?

I can see arguments for and against.  (notably, it's a breaking change).  With the appropriate type checking and exception messages, it should be possible to keep a single cmap argument, but you just need to be careful about the inspection of the object and provide clear help/feedback

kushalkolar:
> > Question: We could enforce that the cmap kwarg should represent a single cmap. If the user wants to pass a list of cmaps we make a new kwarg called cmaps?
> 
> I can see arguments for and against. (notably, it's a breaking change). With the appropriate type checking and exception messages, it should be possible to keep a single cmap argument, but you just need to be careful about the inspection of the object and provide clear help/feedback

Hi, sorry for the late reply, been slow getting back into this after the winter holidays!

I think we can allow the following types of use for the `cmap` argument in `LineCollection`:

1. If it is a `str` of a valid cmap name, ex. jet, then it is handled how it currently is; first line would be dark blue, last line would be red.
2. If it is a `list` of `str` check if it is possible to make a cmap from it using your library (so this should be a list of valid color names). If not go to possiblity (3) below.
3. If it is a `list` of `str` with the same length as the number of lines, then check to make sure every `str` in the list is a valid cmap and each individual line uses this cmap across the length of the line.
4. If it is a `list` of `list`, if `len(outerlist) == n_lines`, check that each sublist can be made into a cmap from your library and use this to color each individual line.

I think this should cover all the cases? :smile:

PR #404, opened 2024-02-19
add `Deleted` as a graphic feature

kushalkolar:
This makes it possible to create callbacks to do things when a `Graphic` is deleted from a `PlotArea`.

~garbage collection is actinng wonky again so I need to fix that too...~ nevermind, I think was accessing stuff users wouldn't normally access in jupyter so ipython was clinging onto it, as usual :smile: 

This also changes all `feature_events` class attributes to be a `set`.

Useage:
image

### DIFF
--- a/fastplotlib/graphics/_features/_deleted.py
+++ b/fastplotlib/graphics/_features/_deleted.py
@@ -1,0 +1,41 @@
+from ._base import GraphicFeature, FeatureEvent
+
+
+class Deleted(GraphicFeature):
+    """
+    Used when a graphic is deleted, triggers events that can be useful to indicate this graphic has been deleted
+
+    **event pick info:**
+
+     ==================== ======================== =========================================================================
+      key                  type                     description
+     ==================== ======================== =========================================================================
+      "collection-index"   int                      the index of the graphic within the collection that triggered the event
+      "world_object"       pygfx.WorldObject        world object
+     ==================== ======================== =========================================================================
+    """
+
+    def __init__(self, parent, value: bool):
+        super(Deleted, self).__init__(parent, value)
+
+    def _set(self, value: bool):
+        value = self._parse_set_value(value)
+        self._feature_changed(key=None, new_data=value)
+
+    def _feature_changed(self, key, new_data):
+        # this is a non-indexable feature so key=None
+
+        pick_info = {
+            "index": None,
+            "collection-index": self._collection_index,
+            "world_object": self._parent.world_object,
+            "new_data": new_data,
+        }
+
+        event_data = FeatureEvent(type="deleted", pick_info=pick_info)
+
+        self._call_event_handlers(event_data)
+
+    def __repr__(self) -> str:
+        s = f"DeletedFeature for {self._parent}"
+        return s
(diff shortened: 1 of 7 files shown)


kushalkolar:
@clewis7 notebook tests are failing due to a new change in `pygfx` which we haven't caught up to yet, but I want to merge this anyways and deal with the `pygfx` changes later.

fastplotlib/graphics/_features/_deleted.py
```diff
@@ -0,0 +1,41 @@
...
+     ==================== ======================== =========================================================================
+      "collection-index"   int                      the index of the graphic within the collection that triggered the event
+      "world_object"       pygfx.WorldObject        world object
+     ==================== ======================== =========================================================================
+    """
+
+    def __init__(self, parent, value: bool):
+        super(Deleted, self).__init__(parent, value)
+
+    def _set(self, value: bool):
```
clewis7, line 21: So, this will get triggered when `__del__` is called and `self.deleted = True` is set?

clewis7: How are we preventing users from if a graphic has been deleted to still having reference to the graphic and trying to say something silly like `graphic.deleted=False`? Will this throw an error if the graphic has already been deleted? I would assume so because there will be no reference.

fastplotlib/graphics/_features/_deleted.py
```diff
@@ -0,0 +1,41 @@
...
+        value = self._parse_set_value(value)
+        self._feature_changed(key=None, new_data=value)
+
+    def _feature_changed(self, key, new_data):
+        # this is a non-indexable feature so key=None
+
+        pick_info = {
+            "index": None,
+            "collection-index": self._collection_index,
```
clewis7, line 30: What is the difference here between `index` vs `collection-index`? Also, `collection-index` could be `None` if the graphic is not a `CollectionGraphic`, no?

clewis7: It seems like `index` will always be `None`?

clewis7 (commented):
Had a couple of clarifying questions? Has been too long since I have looked at this. Otherwise, LGTM!

fastplotlib/graphics/_features/_deleted.py
kushalkolar: > So, this will get triggered when `__del__` is called and `self.deleted = True` is set?

Yes, `Graphic.__del__` sets `self.deleted = True` https://github.com/fastplotlib/fastplotlib/pull/404/files#diff-a57994ecff1cde4e05c8175265219f0b5200da144ce8733b9eb560c09bcea7ddR179 

> How are we preventing users from if a graphic has been deleted to still having reference to the graphic and trying to say something silly like `graphic.deleted=False`? Will this throw an error if the graphic has already been deleted? I would assume so because there will be no reference.

Since users are only given acccess to weakreferences, it will throw a reference error if they try to do _anything_ with a deleted `Graphic`.

fastplotlib/graphics/_features/_deleted.py
kushalkolar: index is buffer index, used for things like `data` and `colors` features. Not used here since it's not relevant

PR #405, opened 2024-02-19
fix bug when remove_graphic() is used

kushalkolar:


kushalkolar:
Tests are failing due to new changes in `pygfx` we haven't caught up with yet, merging this since it's working locally. Will catchup with `pygfx@main` soon.

PR #406, opened 2024-02-19
legends, part 1

### LINKED ISSUE #403: Legends
It's time to add legends :smile: ! Laying out ideas first, with some inspiration from `pyqtgraph`

## Basics

Multiple legends will be allowed in a `Plot`, `Subplot` or `Dock`. They can exist in moveable and non-moveable states.

A `Legend` is created and then graphics can be added to it:

```python
import fastplotlib as fpl

plot = fpl.Plot()

line_a = plot.add_line(line_data, name="A")

# create a legend
legend = fpl.add_legend(plot.docks["top"])   # can also just be `plot`

# add a "legend item" by adding the graphic to the legend
legend.add_graphic(line_a, label=None)  # if label is None, uses graphic name to set the label
```

## Line collections or scatters

For line collections, use a standard legend if a quantitative colormap was used.

```python
line_coll = plot.add_line_collection(data, colors="quantitative")

# use the same legend box or create a new one
legend.add_graphic(line_coll, labels=["B", "B", "C", "C", "D"])  # label for each line in the collection, legend only contains unique labels
```

If a quantitative cmap is used, use colorbars instead

```python
line_coll_jet = plot.add_line_collection(data, colors="jet")  # quantitative cmap

cbar = plot.add_colorbar()

cbar.add_graphic (line_coll_jet)  # colorbar if quantitative data are used for lines and scatter
```

## More thoughts:
* For `LineCollection`, do we need to do a sanity check to make sure that when a user passes `labels`, that these lines all have the same color? Likewise for other aspects like line `thickness`.
* For scatter, there can be different colors and sizes too, so that will also have to be dealt with just like qualitative vs. quantitatively.
* ImageGraphic Colorbars are made per-ImageGraphic, and auto-made and placed next to the HistogramLUTWidget when present.

kushalkolar:
starts #403 

major WIP


kushalkolar:
got some very basic stuff working, can add simple `LineGraphics` to the legend

image

kushalkolar:
I should determine a way to make a sorting method for LegendItems in space (based on the size of each) and then organize into an optimal column, row or grid of items. Get the bbox for all LegendItems and set the Legend mesh to the bbox.

kushalkolar:
If legend is placed in the main plot area scale the text size and spacing based on current scene bbox.

fastplotlib/legends/legend.py
```diff
@@ -0,0 +1,344 @@
...
+            pass
+
+        elif graphic.name is not None:
+            pass
+
+        else:
+            raise ValueError("Must specify `label` or Graphic must have a `name` to auto-use as the label")
+
+        if np.unique(graphic.colors(), axis=0).shape[0] > 1:
+            raise ValueError("Use colorbars for multi-colored lines, not legends")
```
clewis7, line 61: Can you add code comment here explaining this better?

I assume you are checking to see if there is more than 1 color for a `Graphic`, and if so, a `Legend` is not appropriate to use?

clewis7 (commented):
Is there a plan to have something like `plot.add_legend()`, that would take a `plot_area` and auto generate a legend based on existing graphics in the plot? 

Notes:
1. Interesting functionality, if I don't specify a dock to put the legend in, it places it in the middle of the plot. Should we have a default dock to use ("top" or "right") if someone just passes the plot?
2. Also, if I don't change the size of the dock before adding the legend, it doesn't appear? Should we automatically change the size of the dock area that the legend is being added to?
3. Legend location: it appears like the legend is not centered in the middle of the plot. 
image

clewis7:
I see here you provided images exemplifying use, I think we should add a `legends.ipynb`. 

@kushalkolar thoughts?

clewis7:
>Interesting functionality, if I don't specify a dock to put the legend in, it places it in the middle of the plot. Should we have a default dock to use ("top" or "right") if someone just passes the plot?

I see you can pass a `Plot`, `Dock`, or `Subplot`. My only worry is that when I just passed the same plot that my data is in it put the legend over the graphic. Guess maybe it is up to the user to know that you can put legend in a plot or subplot by itself, or you can put it in a dock.

fastplotlib/legends/legend.py
```diff
@@ -0,0 +1,344 @@
...
+
+        width, height, _ = bbox.ptp(axis=0)
+
+        self._mesh.geometry.positions.data[mesh_masks.x_right] = width + 7
+        self._mesh.geometry.positions.data[mesh_masks.x_left] = -5
+        self._mesh.geometry.positions.data[mesh_masks.y_bottom] = 0
+        self._mesh.geometry.positions.data[mesh_masks.y_bottom] = -height - 3
+        self._mesh.geometry.positions.update_range()
+
+    def remove_graphic(self, graphic: Graphic):
```
clewis7, line 277: Does deleting a graphic from the plot automatically remove it from the legend if it is in the legend?

fastplotlib/legends/legend.py
kushalkolar: We made some more thoughts on this here, but for now only single colored lines in this PR https://github.com/fastplotlib/fastplotlib/issues/403#issuecomment-1961981225

kushalkolar:
> Is there a plan to have something like `plot.add_legend()`, that would take a `plot_area` and auto generate a legend based on existing graphics in the plot?

We can think about this later once we have all the legend functionality we want.

> 1. Interesting functionality, if I don't specify a dock to put the legend in, it places it in the middle of the plot. Should we have a default dock to use ("top" or "right") if someone just passes the plot?

Nope, it can make sense to put the legend in the main plot area too, the user can move them out of the way of existing graphics if required.

> 2. Also, if I don't change the size of the dock before adding the legend, it doesn't appear? Should we automatically change the size of the dock area that the legend is being added to?

Done

> 3. Legend location: it appears like the legend is not centered in the middle of the plot.

We can fix this later XD

> I see here you provided images exemplifying use, I think we should add a `legends.ipynb`.

We can do this after implementing all legend functionality that we want, basically when we fully resolve #403 

> it is up to the user to know that you can put legend in a plot or subplot by itself, or you can put it in a dock.

Yup!

fastplotlib/legends/legend.py
kushalkolar: yes, see: https://github.com/fastplotlib/fastplotlib/pull/406/files#diff-042597b7ed37138aa42fa3fdbdf41423475a26bbb17c0d585b8449a0c0419ad9R271

clewis7: ahh, always the partials

kushalkolar:
> Is there a plan to have something like `plot.add_legend()`,

Now after the recent References refactor I realize this is probably the best way to do it, keep the API uniform. Can properly look into it after or during axes.

PR #410, opened 2024-02-23
gpu info upon notebook import

kushalkolar:
Displays gpu info when imported in notebooks

image

### DIFF
--- a/fastplotlib/__init__.py
+++ b/fastplotlib/__init__.py
@@ -3,7 +3,6 @@ from pathlib import Path
 from .layouts import Plot, GridPlot
 from .graphics import *
 from .graphics.selectors import *
-from .legends import *
 from .utils import print_adapter_info
 
 from wgpu.gui.auto import run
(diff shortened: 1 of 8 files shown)


fastplotlib/__init__.py
```diff
@@ -3,6 +3,8 @@
 from .layouts import Plot, GridPlot
 from .graphics import *
 from .graphics.selectors import *
+from .legends import *
```
clewis7, line 6: I think you might have checked this branch out from somewhere you were working on legends. I don't think legends should be in this PR.

clewis7 (commented):
Tried it locally on my mac and after removing the legends import statement seems to work fine. LGTM

kushalkolar:
I also need to add `git lfs` installs to the pypi publish action.

Should probably also wrap the `print_gpu_info()` call in `__init__.py` in a `try except` so that the info banner doesn't block `import` for any weird reason that doesn't affect fpl functionality.

PR #413, opened 2024-02-27
Draggable Graphics

### LINKED ISSUE #407: Base class for moveable `Graphic`
It might be useful to have a general base class for all moveable graphics, or maybe allow all `Graphic` objects to be moveable with a toggle? Basically a base class that takes care of pointer and keyboard events and disabling the controller during the pointer movement.

Resizable which changes the object's world scale. We can do rotation modes later.

kushalkolar:
closes #407


fastplotlib/graphics/_base.py
```diff
@@ -66,6 +68,12 @@ def __init__(
 
         metadata: Any, optional
             metadata attached to this Graphic, this is for the user to manage
+            
+        collection_index: int
+            if relevant, the sub-index of this Graphic within the GraphicCollection
+
+        draggable: bool, default ``False``
```
clewis7, line 75: to clarify here: at this point in time all graphics are draggable right?

fastplotlib/graphics/_features/_drag.py
```diff
@@ -0,0 +1,102 @@
...
+class DragFeature(GraphicFeature):
+    """
+    Manages mouse and keyboard drag events
+    """
+    def __init__(
+            self,
+            parent,
+            multiplier: np.ndarray = None,
+            clamp: np.ndarray = None,
+    ):
```
clewis7, line 17: doc string here?

clewis7: can you explain `clamp` and `multiplier`? names are not super explanatory

clewis7 (commented):
So, purpose of this PR is to create base class for handling moving `Graphics`?

LGTM codewise 

I'm not sure I understand the relevancy of `clip` and `multiplier`. It looks like they are mostly always going to be `None`?

fastplotlib/graphics/_base.py
```diff
@@ -67,6 +77,13 @@ def __init__(
 
         metadata: Any, optional
             metadata attached to this Graphic, this is for the user to manage
+            
+        collection_index: int
+            if relevant, the sub-index of this Graphic within the GraphicCollection
+
+        draggable: one of ``None``, ``"mouse"``, ``"kb"``, or ``"mouse+kb"``, default ``None``
```
clewis7, line 84: I would maybe specify that "kb" here is keyboard

fastplotlib/graphics/_features/_drag.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        multiply the drag vector by this multplier vector. final_vector = delta * multiplier.
+        by default the multiplier is set to (1, 1, 1). For example, you can disable movement along an
+        axis by setting the multiplier for that axis component to zero, for example to restrict movement
+        only along the x-axis: use ``multiplier = (1, 0, 0)``
+        """
+        return self._multiplier
+
+    @multiplier.setter
+    def multiplier(self, value: np.ndarray):
```
clewis7, line 58: Should there be more parsing here for size of `value` passed? Should always be a `(1,3)`

Also, should we allow passing a single `float` that would apply to `xyz`?

kushalkolar:
on hold until https://github.com/fastplotlib/fastplotlib/pull/413

kushalkolar:
I forgot what usecase this was for, let's close unless we come across a usecase again

PR #426, opened 2024-03-08
Just use asarray

almarklein:
Small cleanup


kushalkolar:
I just noticed this is in Gridplot and not GraphicFeature (which is where buffers are handled), yup this makes sense!

Issue #384, opened 2023-11-28
Use qtpy so we can use PySide etc


kushalkolar:
The Qt output widget and toolbar is something I quickly put together because I needed it.

I'd need to add the pyside and qtpy stuff to the qt modules (output and toolbar) here: https://github.com/fastplotlib/fastplotlib/tree/main/fastplotlib%2Flayouts%2F_frame

And here: https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib%2Flayouts%2F_utils.py#L18

PR #429, opened 2024-03-12
Image widget

almarklein:
Just a small example to show the image widget on desktop. 

image

### DIFF
--- a/examples/desktop/image/image_cmap.py
+++ b/examples/desktop/image/image_cmap.py
@@ -3,6 +3,7 @@ Simple Plot
 ============
 Example showing simple plot creation and subsequent cmap change with Standard image from imageio.
 """
+
 # test_example = true
 
 import fastplotlib as fpl
--- a/examples/desktop/image/image_rgb.py
+++ b/examples/desktop/image/image_rgb.py
@@ -3,6 +3,7 @@ Simple Plot
 ============
 Example showing the simple plot creation with 512 x 512 2D RGB image.
 """
+
 # test_example = true
 
 import fastplotlib as fpl
--- a/examples/desktop/image/image_rgbvminvmax.py
+++ b/examples/desktop/image/image_rgbvminvmax.py
@@ -3,6 +3,7 @@ Simple Plot
 ============
 Example showing the simple plot followed by changing the vmin/vmax with 512 x 512 2D RGB image.
 """
+
 # test_example = true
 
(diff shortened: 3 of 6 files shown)


almarklein:
BTW, there are a lot of files that get touched when I run black. What is the policy related to formatting and linting?

almarklein:
> For this example, I think it'd be useful to add a note that if you want sliders to scroll through one or two dimensions you should use the ImageWidget in either Qt or jupyter. Since we don't have within canvas UI right now.

👍 

Odd though, because I have PySide6 installed, and it selects glfw. Will have a look why this is.

almarklein:
Oh, it only looks for PyQt6 for now. Ref #384

kushalkolar:
> Oh, it only looks for PyQt6 for now. Ref #384 

Yup! I updated that issue with the specifics.

almarklein:
Backend selection is a bit iffy. Will still select glfw because wgou prefers it over qt. Something for another PR :)

kushalkolar:
Yup I think that's why our Qt example explicitly chooses the canvas.

kushalkolar:
Yup I think that's why our Qt example explicitly chooses the canvas

kushalkolar:
If this is good to go you can mark it as ready for review to trigger the CI to run, and then copy the new screenshot from the regenerate action to the corresponding examples dir.

almarklein:
> CI: `AttributeError: module 'fastplotlib' has no attribute 'widgets'`

But when I `import fastplotlib as flp` I can use `flp.widgets`. Which behavior is to be expected? :)

kushalkolar:
ImageWidget is intended to be used from top level, i.e. fpl.ImageWidget. Not entirely sure how this is happening.

almarklein:
This is. Perhaps at some point the ImageWidget relied on ipywidgets?

https://github.com/fastplotlib/fastplotlib/blob/97dd7581598e08d2320c1a208737164be2cdbdb7/fastplotlib/__init__.py#L11-L16

kushalkolar:
> This is. Perhaps at some point the ImageWidget relied on ipywidgets?
> 
> https://github.com/fastplotlib/fastplotlib/blob/97dd7581598e08d2320c1a208737164be2cdbdb7/fastplotlib/__init__.py#L11-L16

Ah yes good catch! That shouldn't be there.

PR #430, opened 2024-03-12
Remove duplicate code (mesh_masks)

almarklein:


fastplotlib/graphics/selectors/_rectangle_region.py
```diff
@@ -3,10 +3,10 @@
 
 import pygfx
 
+from ...utils import mesh_masks
 from .._base import Graphic
 from .._features import GraphicFeature
 from ._base_selector import BaseSelector
-from ._mesh_positions import x_right, x_left, y_top, y_bottom
```
almarklein, line 9: This import was impossible, so I guess this module is not imported by default?

kushalkolar: Yes, most of the code in selectors is a playground for now. Only linear and linear region are functional and publicly exposed right now.

PR #431, opened 2024-03-13
Improvements to GUI selection and Qt support

### LINKED ISSUE #384: Use qtpy so we can use PySide etc

almarklein:
Closes #384 

* [x] Adds a new `fpl.utils/gui` module where the GUI selection is handled and nothing else.
* [x] Fastplotlib lists its own GUI preference order, but respects users who override it.
* [x] Avoid importing multiple GUI libraries.
  * This makes fpl lighter.
  * In my experience importing multiple GUI frameworks can also result in odd bugs.  
  * Remove all imports of `wgpu.gui.auto` except in `fpl.utils.gui`.
  * Users can still use a custom canvas and thus mix GUI toolkits if they want to go crazy :)
* [x] The `fpl.run()` method now matches the GUI backend in use (this was not necessarily the case previously).
* [x] Support for all modern Qt libraries!
* [x] The cases where `ipywidgets` is imported is now conditional (only happens in Jupyter notebook / lab).

Note on performance: The import time is about 0.2 s faster on desktop (because it does not import Jupyter stuff). In the notebook we can expect a similar improvement because glfw and qt are not  imported by default (when available).


fastplotlib/widgets/image.py
```diff
@@ -927,9 +919,17 @@ def show(
             ImageWidget just uses the Gridplot output context
         """
         if self.gridplot.canvas.__class__.__name__ == "JupyterWgpuCanvas":
+            from ..layouts._frame._ipywidget_toolbar import (
```
kushalkolar, line 922: do these types of imports work with pyinstaller?

fastplotlib/layouts/_frame/_frame.py
```diff
@@ -158,6 +124,8 @@ def show(
 
         # return the appropriate OutputContext based on the current canvas
         if self.canvas.__class__.__name__ == "JupyterWgpuCanvas":
+            from ._jupyter_output import JupyterOutputContext  # noqa - inline import
```
kushalkolar, line 127: wonder how pyinstaller and other tools that make distributables deal with this

fastplotlib/utils/gui.py
```diff
@@ -0,0 +1,56 @@
...
+# create our own preferred order. If we're in Jupyter, wgpu will prefer and
+# select Jupyter, so no no action needed. If one of the GUI backends is already
+# imported, this is likely because the user want to use that one, so we should
+# honor that. Otherwise, we try importing the GUI backends in the order that
+# fastplotlib prefers. When wgpu-py loads the gui.auto, it will see that the
+# respective lib has been imported and then prefer the corresponding backend.
+
+# A list of wgpu GUI backend libs, in the order preferred by fpl
+gui_backend_libs = ["PySide6", "PyQt6", "PySide2", "PyQt5", "glfw"]
+
```
kushalkolar, line 29: Unless I missed it, we also need a check for if it's running in ipython/jupyter and if `%gui qt` is being used, I found that [this](https://github.com/fastplotlib/fastplotlib/pull/431/files#diff-4fb24c475c133664d23618fd061b4ae9c63fd610f0cf57962a1abfe52bae9ef1L41-L42) worked to do that determination.

I tried to ask here but maybe it's the wrong place to ask :laughing: https://discourse.jupyter.org/t/what-is-the-right-way-to-determine-if-a-gui-qt-event-loop-is-running/22160

kushalkolar (changes requested):

fastplotlib/utils/gui.py
kushalkolar: Also, when the user has invoked `%gui qt` we must use Qt because `jupyter_rfb` stops working in that case, even if you force `canvas = jupyter`.

fastplotlib/layouts/_utils.py
kushalkolar, line 1: this is really great :joy:

fastplotlib/widgets/image.py
almarklein: Just as well as it did before this change ;) because they were in a conditional there as well. It might be a good idea to implement a pyinstaller hook like we [do in wgpu](https://github.com/pygfx/wgpu-py/blob/main/wgpu/__pyinstaller/hook-wgpu.py) to just include all modules.

fastplotlib/utils/gui.py
almarklein: Oh interesting! I had assumed (also in wgpu-py) that if you're in Jupyter you can assume an asyncio event loop and ipywidets to be alove 🤔

almarklein: Can we assume that the `%gui  something`  is called once, before fpl/wgpu is imported, and not changed halfway?

almarklein: I mean, surely a user can do so, but is it reasonable to expect users to expect things to break then 😉 ?

kushalkolar: > Can we assume that the `%gui something`  is called once, before fpl/wgpu is imported, and not changed halfway?

Yes, changing it in the middle leads to very weird behavior. When they open a nb they should choose jupyter (default) or gui Qt before making any plot.

fastplotlib/widgets/image.py
kushalkolar: I made #434 for this, can look into it later.

almarklein:
Funny how https://github.com/pygfx/pygfx/issues/674 gets reported while we're working on this 😄   I think I'll put this PR on hold, and first address the GUI backend selection mechanics. I think your code to check what GUI Jupyter has integrated should be used in wgpu-py. And perhaps it makes sense to provide more ways to influence backend preference, that fpl can then use (in the new `gui.py`).

kushalkolar:
yup makes sense to do it upstream, less wheel reinventing :smile:

kushalkolar:
After https://github.com/pygfx/wgpu-py/pull/478 it should be possible to greatly simplify this

almarklein:
I made some updates to make it work with the latest wgpu. Ready from my end!

clewis7 (approved):

kushalkolar (approved):

PR #432, opened 2024-03-13
Lazy import av

almarklein:
After #431 I did a review of all imports. I found 2 small interesting cases, which are addressed in this PR:

* One absolute import of a fastplotlib module.
* The optional dependency `av` can easily be lazy-imported (shaving nearly 0.1 off the import time on my machine).

### DIFF
--- a/fastplotlib/layouts/_frame/_toolbar.py
+++ b/fastplotlib/layouts/_frame/_toolbar.py
@@ -1,4 +1,4 @@
-from fastplotlib.layouts._subplot import Subplot
+from .._subplot import Subplot
 
 
 class ToolBar:
--- a/fastplotlib/layouts/_record_mixin.py
+++ b/fastplotlib/layouts/_record_mixin.py
@@ -3,12 +3,17 @@ from pathlib import Path
 from multiprocessing import Queue, Process
 from time import time
 
-try:
-    import av
-except ImportError:
-    HAS_AV = False
-else:
-    HAS_AV = True
+
+def _get_av():
+    try:
+        import av
+    except ImportError:
+        raise ModuleNotFoundError(
+            "Recording to video file requires `av`:\n"
+            "https://github.com/PyAV-Org/PyAV"
+        ) from None
+    else:
+        return av
 
 
 class VideoWriterAV(Process):
@@ -28,6 +33,7 @@ class VideoWriterAV(Process):
         super().__init__()
         self.queue = queue
 
+        av = _get_av()
         self.container = av.open(path, mode="w")
 
         self.stream = self.container.add_stream(codec, rate=fps, options=options)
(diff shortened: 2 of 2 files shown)


fastplotlib/layouts/_frame/_toolbar.py
```diff
@@ -1,4 +1,4 @@
-from fastplotlib.layouts._subplot import Subplot
+from .._subplot import Subplot
```
kushalkolar, line 1: This definitely slipped through when I used pycharm to refactor moving stuff to different modules, it likes to change relative imports to absolute.

fastplotlib/layouts/_record_mixin.py
```diff
@@ -3,12 +3,17 @@
...
-    HAS_AV = True
+
+def _get_av():
+    try:
+        import av
+    except ImportError:
+        raise ModuleNotFoundError(
+            "Recording to video file requires `av`:\n"
+            "https://github.com/PyAV-Org/PyAV"
+        ) from None
```
kushalkolar, line 14: What is the `from None`, never seen that before 😄

almarklein: This is related to exception chaining:
```py
try:
    1/0
except ZeroDivisionError:
    raise ValueError("meh")
```
Causes:
```
  File "/Users/almar/dev/py/tmp.py", line 2, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "/Users/almar/dev/py/tmp.py", line 4, in <module>
    raise ValueError("meh")
ValueError: meh
```

A chain can be made explicit using the `from`  keyword:
```py
try:
    1/0
except ZeroDivisionError as err:
    raise ValueError("meh") from err
```
Then:
```
  File "/Users/almar/dev/py/tmp.py", line 2, in <module>
    1/0
    ~^~
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "/Users/almar/dev/py/tmp.py", line 4, in <module>
    raise ValueError("meh") from err
ValueError: meh
```

And it can be used (as in the code this comment refers to) to tell that the exception being raised is "standalone":
```py
try:
    1/0
except ZeroDivisionError as err:
    raise ValueError("meh") from None
```

Then:
```
  File "/Users/almar/dev/py/tmp.py", line 4, in <module>
    raise ValueError("meh") from None
ValueError: meh
```

almarklein: TL;DR: The `from None` disables exception chaining.

kushalkolar: Ah TIL! Cleaner messages is nice :smile:

EricThomson: Very cool I didn't know this was possible.

PR #433, opened 2024-03-13
A bit of black

almarklein:
* [x] The script that generates `graphic_methods_mixin.py` now also blackens the code. (could take this nearly verbatim from the wgpu codegen)
* [x] Blackend a few more files.

Now only the examples are unblacked. I can add that too if you want, but I was not sure if certain PR's may be affected.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -7,9 +7,9 @@ import fastplotlib
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
-project = 'fastplotlib'
-copyright = '2023, Kushal Kolar, Caitlin Lewis'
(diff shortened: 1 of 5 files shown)


almarklein:
I think it makes sense to move the `generate_x` scripts out of the `fastplotlib` dir into a special `scripts`  dir in the root of the repo, since they're not part of the library. But `generate_colormaps.py` is touched in #390, so I left it in place.

kushalkolar:
> I think it makes sense to move the `generate_x` scripts out of the `fastplotlib` dir into a special `scripts` dir in the root of the repo, since they're not part of the library.

Makes sense!

> But `generate_colormaps.py` is touched in #390, so I left it in place.

Yup I need to get back to that PR, probably next week. But `generate_colormaps.py` will disappear after we incorporate Talley's `cmap` lib

PR #436, opened 2024-03-18
Add black to setup in optional feature set for tests

### LINKED ISSUE #435: Should black be in setup.py optional deps for tests?
Maybe I'm missing something : shouldn't we add `black` to optional deps for `test` set of optional deps?

```
    "tests": [
        "pytest<8.0.0",
        "nbmake",
        "scipy",
        "imageio[pyav]",
        "jupyterlab",
        "jupyter-rfb>=0.4.1",
        "ipywidgets>=8.0.0,<9",
        "scikit-learn",
        "tqdm",
        "sidecar",
    ],
```

EricThomson:
Closes #435. 

Also changes `.gitignore`, adding VS Code config file.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -6,5 +6,5 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: psf/black@stable
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,10 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: psf/black@stable
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -9,4 +9,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: psf/black@stable
         with:
-            jupyter: true
+          src: "./fastplotlib"
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -8,3 +8,5 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: psf/black@stable
+        with:
+            jupyter: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,9 @@ on:
       - ready_for_review
 
 jobs:
+  black-check:
+    - name: Black Code Formatter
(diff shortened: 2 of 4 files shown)


kushalkolar:
Might also be time to add a GitHub action to check for black

.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ on:
       - ready_for_review
 
 jobs:
+  black-check:
```
kushalkolar, line 17: I think the format needs to be something like this:

```yaml
black-check:
  name: black
  steps:
    - uses: lgeiger/....
```

kushalkolar (changes requested):

.github/workflows/ci.yml
kushalkolar: ok seems like this is not the only thing, I'll look at it later

kushalkolar: or create a new workflow entirely, `black.yml`

EricThomson: We should probably look into https://black.readthedocs.io/en/stable/integrations/github_actions.html

Which is referenced in https://github.com/marketplace/actions/run-black-formatter

kushalkolar: oh that looks better, just add a yml with that!

EricThomson:
At point now where not sure if `black.yml` is failing because it is written incorrectly or because the linter is finding problems in the code.  :shrug:

kushalkolar:
ok so we need to black our nbs too :laughing:

.github/workflows/black.yml
```diff
@@ -0,0 +1,12 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: psf/black@stable
+        with:
```
kushalkolar, line 11: I actually don't want to black our examples because things like this are useful in the examples:

Ours:
```python
cmap_values = [
    0, 1, 1, 2,
    0, 0, 1, 1,
    2, 2, 3, 3,
    1, 1, 1, 5
]
```

black:
```
cmap_values = [0, 1, 1, 2, 0, 0, 1, 1, 2, 2, 3, 3, 1, 1, 1, 5]
```

So let's set jupyter to false and only check the actual sources, i.e. `./fastplotlib`

kushalkolar (changes requested):
one last thing and should be good to go! :D

.github/workflows/black.yml
EricThomson: Ok I removed the jupyter line, it defaults to false, and am trying to direct just to the fastplotlib dir.

once nice thing with fpl this stuff runs really fast!

kushalkolar (approved):

PR #437, opened 2024-03-18
sphinx gallery for docs

clewis7:
starts #486

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -4,6 +4,26 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 import fastplotlib
 
+from sphinx_gallery.sorting import ExplicitOrder
+import wgpu.gui.offscreen
+import os
+import sys
+
+ROOT_DIR = os.path.abspath(os.path.join(__file__, "..", ".."))
+sys.path.insert(0, ROOT_DIR)
+
+# -- Sphix Gallery Hackz -----------------------------------------------------
+# When building the gallery, render offscreen and don't process
+# the event loop while parsing the example
+
+
+def _ignore_offscreen_run():
+    wgpu.gui.offscreen.run = lambda: None
+
+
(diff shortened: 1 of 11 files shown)


clewis7:
Currently, the issue that I am running into is that in the `conf.py` I am forcing offscreen rendering but then when I go to make the html the wgpu canvas is rendering which causes the build process to stall

docs/source/conf.py
```diff
@@ -4,6 +4,26 @@
...
+ROOT_DIR = os.path.abspath(os.path.join(__file__, "..", ".."))
+sys.path.insert(0, ROOT_DIR)
+
+# -- Sphix Gallery Hackz -----------------------------------------------------
+# When building the gallery, render offscreen and don't process
+# the event loop while parsing the example
+
+
+def _ignore_offscreen_run():
+    wgpu.gui.offscreen.run = lambda: None
```
almarklein, line 21: This bit should no longer be necessary. It's from a time when the offscreen backend still fired up asyncio (it does not do so anymore). The `WGPU_FORCE_OFFSCREEN` is the only thing needed.

almarklein:
I happen to be working on https://github.com/pygfx/pygfx/pull/679.

docs/source/conf.py
```diff
@@ -33,6 +54,41 @@
...
+    #     "Controller",
+    #     "Camera",
+    #     "show",
+    #     "Display",
+    #     "Group",
+    #     "Scene",
+    #     "Light",
+    # },
+    "doc_module": ("fastplotlib",),
+    "image_scrapers": ("fastplotlib",),
```
almarklein, line 78: If you put "pygfx" here instead, I *think* you can use our scraper, so you don;t need the `gallery_scraper.py`.

almarklein:
> [...] the wgpu canvas is rendering which causes the build process to stall

Fastplotlib is eagerly trying to use glfw when I try it too, ignoring the offscreen setting. I expect that this should be fixed with #431

PR #439, opened 2024-03-19
Rotation

### LINKED ISSUE #427: method for rotating graphics
A method on `Graphic` for rotating graphics around an arbitrary line given by 2 vectors, etc. @BalzaniEdoardo wants to try this since his fav mathmetical object is quaternions :smile:

BalzaniEdoardo:
Added a rotation property and a setter to `Graphic` and a method to set a rotation of a given angle over one of the axis, with respect to the world.

closes #427

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 
 import numpy as np
+import pylinalg as la
 
 from pygfx import WorldObject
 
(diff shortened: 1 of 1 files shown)


kushalkolar:
good to go after a test :D

BalzaniEdoardo:
> good to go after a test :D

Nice!!

kushalkolar:
Some notes on testing here: https://github.com/fastplotlib/fastplotlib/blob/main/CONTRIBUTING.md#running-tests

almarklein:
A general remark about transformations for the graphics object:

I think `world_object.local` makes more sense that `.world`. In that case it represents the rotation with respect to the parent. When the graphics are flat, its the same thing (possibly more efficient). And when there is a scene graph (graphics being children of other graphics) in the majority of cases you want to transform with respect to the parent.

kushalkolar:
Thanks! If all `WorldObjects` in a subplot exist within the same scene, then `.local` and `.world` have the same effect right? Is there a scientific visualization usecase where you might want to differentially use `.local` and `.world`?

fastplotlib/graphics/_base.py
```diff
@@ -196,6 +205,26 @@ def __del__(self):
         self.deleted = True
         del WORLD_OBJECTS[self.loc]
 
+    def rotate(self, alpha: float, axis: Literal["x", "y", "z"] = "y"):
+        """Rotate the Graphic with respect to the world.
+
+        Parameters
+        ----------
+        alpha :
+            Rotation angle in radiants.
```
kushalkolar, line 214: radians?

kushalkolar (changes requested):

kushalkolar (dismissed):

kushalkolar (approved):

PR #441, opened 2024-03-20
see if mac can work in CI

kushalkolar:
Specifying `macos-14` is supposed to select for M1 mac: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories 

I wonder if this comes with GPU support and whether WGPU uses Metal here :smile:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,6 @@ jobs:
           - name: Test py312
             pyversion: '3.12'
     steps:
-    - name: checkout
     - uses: actions/checkout@v3
       with:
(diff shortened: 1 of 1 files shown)


kushalkolar:
So it has some kind of GPU: https://github.com/fastplotlib/fastplotlib/actions/runs/8354227427/job/22867218279?pr=441#step:5:7

diffs are quite large though, I wonder if MSE suggested here would work: https://github.com/pygfx/wgpu-py/issues/400 

diff-line_stack-rgb

PR #442, opened 2024-03-20
add codeowners file

clewis7:


kushalkolar:
Looks like pygfx@main just changed points material api, fix that in another PR and then do this?

almarklein:
`vertex_sizes=True` -> `size_mode='vertex'`

PR #443, opened 2024-03-21
update for changes in pygfx@main

clewis7:
Should update to most recent changes in pygfx


clewis7:
@kushalkolar do we need to update screenshots as well?

almarklein:
> do we need to update screenshots as well?

To provide some context: we improved the points shader to consolidated the API, add support to express sizes in world/model space, and improve the shader code, making it more similar to the line shader. As part of that, the points are likely a tiny bit bigger, to counter for the edge we need to do anti-aliasing, so that aa and no-aa points look equally big.

PR #445, opened 2024-03-22
settable bg color

### LINKED ISSUE #15: background
- [ ] set background in subplots, maybe subplots should keep an instance of a class that's just used for theme-stuff.
  - Something like `subplot.theme.set_background(color='w')`
  - @clewis7 easy first contrib for you :smile:

kushalkolar:
closes #15


kushalkolar:
looks like adding a background changes some things that I didn't expect, will look into later.

PR #453, opened 2024-03-27
Use simpler form of super() where possible

almarklein:
These caught my eye in my earlier review. Simple fix that makes the code a bit simplere / easier to read, but possibly also avoids bugs.

----

Context:

```py
class Foo:    
    def __init__(self):
        print("init Foo")

class Bar(Foo):
    def __init__(self):
        super().__init__()
        print("init Bar")

class Spam(Bar):
    def __init__(self):
        super().__init__()
        print("init Spam")

Spam()
```

Produces:
```
init Foo
init Bar
init Spam
```

Replacing the `super()` in `Spam`  with `super(Spam, self).__init__()` produces the same result.
But if it is replaced with `super(Foo, self).__init__()`, it produces:
```
init Foo
init Spam
```

### DIFF
--- a/fastplotlib/graphics/_features/_colors.py
+++ b/fastplotlib/graphics/_features/_colors.py
@@ -251,7 +251,8 @@ class CmapFeature(ColorFeature):
     """
 
     def __init__(self, parent, colors, cmap_name: str, cmap_values: np.ndarray):
-        super().__init__(parent, colors)
+        # Skip the ColorFeature's __init__
+        super(ColorFeature, self).__init__(self, parent, colors)
 
         self._cmap_name = cmap_name
         self._cmap_values = cmap_values
--- a/fastplotlib/graphics/_features/_colors.py
+++ b/fastplotlib/graphics/_features/_colors.py
@@ -252,7 +252,7 @@ class CmapFeature(ColorFeature):
 
     def __init__(self, parent, colors, cmap_name: str, cmap_values: np.ndarray):
         # Skip the ColorFeature's __init__
-        super(ColorFeature, self).__init__(self, parent, colors)
+        super(ColorFeature, self).__init__(parent, colors)
 
         self._cmap_name = cmap_name
         self._cmap_values = cmap_values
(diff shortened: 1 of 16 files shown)


fastplotlib/graphics/_features/_colors.py
```diff
@@ -253,7 +251,7 @@ class CmapFeature(ColorFeature):
     """
 
     def __init__(self, parent, colors, cmap_name: str, cmap_values: np.ndarray):
-        super(ColorFeature, self).__init__(parent, colors)
```
almarklein, line 256: This case was a potential bug because it'd skip the `ColorFeature.__init__`. Did not yet result in a problem yet, because that init does nothing.

almarklein: Or was this intentional?

almarklein: Looks like it is.

almarklein:
There's a lot of multiple inheritance when you get to the plots and graphics, and that results in some juggling with super and direct calls to `SomeSuperclass.__init__`. Will not burn my hands on that right now. Let's keep this pr small 😅

fastplotlib/layouts/_plot.py
```diff
@@ -45,7 +45,7 @@ def __init__(
             passed to Subplot, for example ``name``
 
         """
-        super(Plot, self).__init__(
+        super().__init__(
```
kushalkolar, line 48: What exactly happens with `super` calls in these situations? Because we want to pass these args to `Subplot` and `Frame` is instantiated separately with different args

almarklein: Using `super()` is equivalent to `super(CurrentClass, self)`. I double-checked that this is the case for all changes in this PR.

It will take the next class in the mro, which is `SubPlot` in this case. You can check the resolution order using `Plot.mro()`. With multiple inheritance, the classes on the right are higher up the tree.

kushalkolar (approved):

PR #459, opened 2024-03-29
Refactor imwidget

### LINKED ISSUE #284: allow RGB for ImageWidget

apasarkar:
Closes #284


fastplotlib/widgets/image.py
```diff
@@ -22,9 +22,12 @@
     2: "xy",
     3: "txy",
     4: "tzxy",
-    5: "tzcxy",
 }
 
+DEFAULT_SLIDER_DIMS = {0: "t", 1: "z"}
+
+ALLOWED_WINDOW_KEYS = {"t", "z"}
```
kushalkolar, line 29: maybe `ALLOWED_WINDOW_DIMS`

fastplotlib/widgets/image.py
```diff
@@ -22,9 +22,12 @@
     2: "xy",
     3: "txy",
     4: "tzxy",
-    5: "tzcxy",
 }
 
+DEFAULT_SLIDER_DIMS = {0: "t", 1: "z"}
```
kushalkolar, line 27: `ALLOWED_SLIDER_DIMS` since this is no longer just the "default" but the only supported order.

fastplotlib/widgets/image.py
```diff
@@ -247,7 +248,7 @@ def __init__(
 
         Can display a single n-dimensional image array or a grid of n-dimensional images.
```
kushalkolar, line 249: modify since the purpose is no longer n-dim

fastplotlib/widgets/image.py
```diff
@@ -247,7 +248,7 @@ def __init__(
...
 
-        Default dimension orders:
+        Standard dimension orders:
```
kushalkolar, line 251: Interpreted dimension orders?

fastplotlib/widgets/image.py
```diff
@@ -262,31 +263,19 @@ def __init__(
...
-            | dimension/axes can be specified using ``str`` such as "t", "z" etc. or ``int`` that indexes the dimension
-            | if window_size is not an odd number, adds 1
-            | use ``None`` to disable averaging for a dimension, example: ``{"t": 5, "z": None}``
+            | You may want to apply functions along certain axes of your data before displaying.
+            | This parameter allows you pass in a dictionary; each key is an axis (allowed axes: "t" and "z").
+            | The value is a tuple: (Callable, window_size). Every callable must take "axis" as an input.
+            | A typical signature might be def my_func(img: np.ndarray, axis=0) -> np.ndarray:
+            | To use window_funcs, you do something like: window_funcs = {"t": (t_callable, window_size_t),
+            | "z": (z_callable, window_size_z)}. So at time index "50", if window_size_t is 10, we look ed at frames
+            | (50 - 5, 50 + 5), etc.
```
kushalkolar, line 273:
```suggestion
            | Apply function(s) with rolling windows along "t" and/or "z" dimensions of the `data` arrays.
            | Pass a dict in the form: {dimension: (func, window_size)}, `func` must take a slice of the data array as the
            | first argument and must take `axis` as a kwarg.
            | Ex: mean along "t" dimension: {"t": (np.mean, 11)}, if `current_index` of "t" is 50, it will pass frames 45 to 50 to `np.mean` with `axis = 0`.
            | Ex2: max along z dim: {"z": (np.max, 3)}, passes current, previous and next frame to `np.max` with `axis = 1`
```

More concise

fastplotlib/widgets/image.py
```diff
@@ -262,31 +263,19 @@ def __init__(
...
 
         frame_apply: Union[callable, Dict[int, callable]]
-            | apply a function to slices of the array before displaying the frame
-            | pass a single function or a dict of functions to apply to each array individually
+            | apply a function to frames of data along the "t" and "z" axes before displaying a single image
+            | (think time-averaging in local windows, etc.).
+            | Pass a single function or a dict of functions to apply to each array individually
             | examples: ``{array_index: to_grayscale}``, ``{0: to_grayscale, 2: threshold_img}``
```
kushalkolar, line 279: spatial filters is another good example, various kernels

fastplotlib/widgets/image.py
```diff
@@ -262,31 +263,19 @@ def __init__(
...
+            | A typical signature might be def my_func(img: np.ndarray, axis=0) -> np.ndarray:
+            | To use window_funcs, you do something like: window_funcs = {"t": (t_callable, window_size_t),
+            | "z": (z_callable, window_size_z)}. So at time index "50", if window_size_t is 10, we look ed at frames
+            | (50 - 5, 50 + 5), etc.
 
         frame_apply: Union[callable, Dict[int, callable]]
-            | apply a function to slices of the array before displaying the frame
-            | pass a single function or a dict of functions to apply to each array individually
+            | apply a function to frames of data along the "t" and "z" axes before displaying a single image
+            | (think time-averaging in local windows, etc.).
```
kushalkolar, line 277:
```suggestion
```

fastplotlib/widgets/image.py
```diff
@@ -372,149 +360,29 @@ def __init__(
                 f"You have passed the following type {type(data)}"
             )
 
-        # default dims order if not passed
-        # updated later if passed
+        if self.ndim not in DEFAULT_DIMS_ORDER.keys():
+            raise ValueError(f"{self.ndim} dimensions not supported")
```
kushalkolar, line 364: add something like "only xy, txy, and tzxy data with or without RGB(A) is supported"

fastplotlib/widgets/image.py
```diff
@@ -372,149 +360,29 @@ def __init__(
...
-                        "If passing a list for `slider_dims` each element must be either an <int> or <str>"
-                    )
-
-        else:
-            raise TypeError(
-                f"`slider_dims` must a <int>, <str> or <list>, you have passed a: {type(slider_dims)}"
-            )
+        # Sliders are made for all dimensions except the last 2
+        self._slider_dims = list()
+        for dim in range(self.ndim - 2):
```
kushalkolar, line 375: I think you can do something like:

```python
for dim_ix, dim_name in zip(range(self.ndim - 2), ("t", "z"))
```

fastplotlib/widgets/image.py
```diff
@@ -372,149 +360,29 @@ def __init__(
...
+            if dim in DEFAULT_SLIDER_DIMS.keys():
+                self.slider_dims.append(DEFAULT_SLIDER_DIMS[dim])
+            else:
+                self.slider_dims.append(f"{dim}")
```
kushalkolar, line 379: I think this will all simplify with the above comment?

fastplotlib/widgets/image.py
```diff
@@ -372,149 +360,29 @@ def __init__(
...
             )
 
-        ao = np.array([sorted(v) for v in self.dims_order])
-
-        if not np.all(ao == ao[0]):
-            raise ValueError(
-                f"`dims_order` for all arrays must contain the same combination of dimensions, your `dims_order` are: "
-                f"{self.dims_order}"
-            )
-
```
kushalkolar, line 435: it's really great to get rid of this deep nest :laughing:

fastplotlib/widgets/image.py
```diff
@@ -615,61 +483,42 @@ def window_funcs(self) -> Dict[str, _WindowFunctions]:
         return self._window_funcs
 
     @window_funcs.setter
-    def window_funcs(self, sa: Union[int, Dict[str, int]]):
+    def window_funcs(self, sa: Dict[str, int]):
```
kushalkolar, line 486: I don't know why I called the arg `sa`, maybe a better name?

fastplotlib/widgets/image.py
```diff
@@ -615,61 +483,42 @@ def window_funcs(self) -> Dict[str, _WindowFunctions]:
...
-
-            dim_str = self.slider_dims[0]
-            self._window_funcs = dict()
-            self._window_funcs[dim_str] = _WindowFunctions(self, *sa)
-
-        # for multiple dims
         elif isinstance(sa, dict):
+            if not set(sa.keys()).issubset(ALLOWED_WINDOW_KEYS):
+                raise ValueError(f"The only allowed keys to window funcs are {list(ALLOWED_WINDOW_KEYS)} " 
+                                 f"Your window func passed in these keys: {list(sa.keys())}")
```
kushalkolar, line 496: nice exception handling :smile:

fastplotlib/widgets/image.py
```diff
@@ -615,61 +483,42 @@ def window_funcs(self) -> Dict[str, _WindowFunctions]:
...
-                if v is not None:
-                    if not callable(v[0]) or not (
-                        isinstance(v[1], int) or v[1] is None
-                    ):
-                        raise TypeError(
-                            "dict argument to `window_funcs` must be in the form of: "
-                            "`{dimension: (func, window_size)}`. "
-                            "See the docstring."
-                        )
+                if not callable(v[0]):
```
kushalkolar, line 506: should also check that `v[1] is int` right?

fastplotlib/widgets/image.py
```diff
@@ -703,23 +552,22 @@ def _process_indices(
...
 
         numerical_dims = list()
+
+        data_ix = None
+        for i in range(len(self.data)):
+            if self.data[i] is array:
+                data_ix = i
+                break
+        if data_ix is None:
+            raise ValueError(f"Given `array` not found in `self.data`")
```
kushalkolar, line 562: I'm wondering why this is here, this exception should never be reached because this is a method that the user never calls :thinking:

fastplotlib/widgets/image.py
```diff
@@ -703,23 +552,22 @@ def _process_indices(
...
-                numerical_dim = self.dims_order[data_ix].index(dim)
-            else:
-                numerical_dim = dim
+            # get axes order for that specific array
+            numerical_dim = self.dims_order[data_ix].index(dim)
 
             indices_dim = slice_indices[dim]
 
-            # takes care of averaging if it was specified
+            # takes care of index selection for this specific axis
```
kushalkolar, line 570: maybe "takes care of window slicing..." ?

kushalkolar (changes requested):

fastplotlib/widgets/image.py
apasarkar: @kushalkolar spatial filters is a good example for frame_apply not window_function right?

kushalkolar: yup!

fastplotlib/widgets/image.py
apasarkar: @kushalkolar I think we should move away from self.ndim in general - as we add support for things like RGBA, keeping track of the "2" will get annoying. This is nicer imo: 

`
self._slider_dims = list()
for dim in range(self.ndim):
    if dim in ALLOWED_SLIDER_DIMS.keys():
        self.slider_dims.append(ALLOWED_SLIDER_DIMS[dim])
`

fastplotlib/widgets/image.py
apasarkar: @kushalkolar good call

fastplotlib/widgets/image.py
```diff
@@ -483,43 +482,54 @@ def window_funcs(self) -> Dict[str, _WindowFunctions]:
         return self._window_funcs
 
     @window_funcs.setter
-    def window_funcs(self, sa: Dict[str, int]):
-        if sa is None:
+    def window_funcs(self, callable_dict: Dict[str, int]):
```
kushalkolar, line 485: Looks like the type annotation is also leftover from when I only used this for averaging. The type annotation can also be more precise in py3.10+

At the top add the import `from typing import Literal`

and then:

```python
def window_funcs(self, callables_dict: dict[Literal["t", "z"], tuple(callable, int)])
```

I'm curious if this works for dicts now :thinking: 

```python
def window_funcs(self, callables_dict: dict[Literal["t", "z"]: tuple(callable, int)])
```

kushalkolar (changes requested):

fastplotlib/widgets/image.py
kushalkolar: ah right, so then it'll stay restricted to just t and z sliders

kushalkolar:
I think we can use the `a[np.newaxis]` method as you mentioned, it returns views of the array which makes sense: `a[np.newaxis, np.newaxis].ctypes.data == a.ctypes.data`

Currently `ImageWidget` uses the lowest "t" and/or "z" dimension sizes to set the max bounds "`_dims_max_bounds`" for the `ImageWidget` and the sliders. If we instead set `_dims_max_bounds` based on the largest sizes in each dimension

https://github.com/fastplotlib/fastplotlib/blob/5ec3fb29946bdff1ade94654b5ec37d034f1b0aa/fastplotlib/widgets/image.py#L540-L546

and then in `@current_index.setter`

https://github.com/fastplotlib/fastplotlib/blob/5ec3fb29946bdff1ade94654b5ec37d034f1b0aa/fastplotlib/widgets/image.py#L218-L221

when iterating through each individual data array, we do something like this:

```python
current_index = dict()
for k in self._current_index:
  dim_ix = dict(reversed(ALLOWED_SLIDER_DIMS))[k]
  current_index[k] = min(self._current_index[k], data.shape[dim_ix])
```

fastplotlib/widgets/image.py
apasarkar: @kushalkolar Can do this, but it might be harder to maintain. Hard coding the accepted keys in two locations (the global variable at the top of the file and in this signature) seems like something we'd want to avoid. 

Also from typing import * is already at the top so I think we are good to go

kushalkolar: Yea you're right, let's just use str for the key then

fastplotlib/widgets/image.py
```diff
@@ -3,28 +3,35 @@
...
+# How many dimensions to display a single image
+# If key is 0 this means no RGB (so each image is 2 dims). If key is 1, it's 3
+RGB_DIM_MAP = {0: 2, 1: 3}
 
-DEFAULT_DIMS_ORDER = {
-    2: "xy",
-    3: "txy",
-    4: "tzxy",
-    5: "tzcxy",
+DEFAULT_IMAGE_FORMAT = "gray"
```
kushalkolar, line 22:
```suggestion
```

kushalkolar (changes requested):

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
...
+            elif k in color_scheme.keys():
+                curr_scheme = color_scheme[k].lower()
+                if curr_scheme not in RGB_DIM_MAP.keys():
+                    raise ValueError(
+                        f"Color Schemes can only be one of {list(RGB_DIM_MAP.keys())} "
+                        f"You passed in {color_scheme[k]} for one of the keys"
+                    )
+                else:
+                    full_color_scheme[k] = curr_scheme
+        return full_color_scheme
```
kushalkolar, line 246:
```suggestion
```

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
...
+
+    def _compute_and_validate_formats(self) -> list[int]:
+
```
kushalkolar, line 249: add docstring

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
...
+        num_scrollable_dims_list = []
```
kushalkolar, line 250: `n_scrollable_dims`

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
...
+        for k in range(len(self._data)):
```
kushalkolar, line 251: ``for array_ix in range(len(self._data))``

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
         """
         return self._current_index
 
+    @property
+    def num_img_dims(self) -> list[int]:
+        """
+
```
kushalkolar, line 211: better docstring

kushalkolar, line 211: `num` -> `n`

fastplotlib/widgets/image.py
```diff
@@ -157,13 +164,19 @@ def data(self) -> List[np.ndarray]:
...
-        return self._dims_order
+    def num_scrollable_dims(self) -> List[int]:
+        """Returns a List describing, for each array, how many of the first dimensions are "scrollable".
+        All other dimensions are used for displaying images"""
+        return self._num_scrollable_dims
+
+    @property
+    def num_img_dims(self) -> List[int]:
+        """Returns a List describing, for each array, how many of the last dimensions are used to display a single img."""
+        return self._num_img_dims
```
kushalkolar, line 179:
```suggestion
```

fastplotlib/widgets/image.py
```diff
@@ -305,19 +375,26 @@ def __init__(
         histogram_widget: bool, default False
             make histogram LUT widget for each subplot
 
+        rgb_disp: list, default None
```
kushalkolar, line 378: just call it `rgb`

kushalkolar: should be a `list[bool]`, of `len(data)`

fastplotlib/widgets/image.py
```diff
@@ -192,6 +205,77 @@ def current_index(self) -> Dict[str, int]:
...
+                if curr_scheme not in RGB_DIM_MAP.keys():
+                    raise ValueError(
+                        f"Color Schemes can only be one of {list(RGB_DIM_MAP.keys())} "
+                        f"You passed in {color_scheme[k]} for one of the keys"
+                    )
+                else:
+                    full_color_scheme[k] = curr_scheme
+        return full_color_scheme
+
+    def _compute_and_validate_formats(self) -> list[int]:
```
kushalkolar, line 248: This is probably useful when `set_data()` is used.

This should take args, `data: List[<array_like>], rgb: List[bool]`

maybe rename this to `_get_n_scrollab_dims(data, rgb)`

fastplotlib/widgets/image.py
```diff
@@ -3,28 +3,35 @@
...
-
 if CANVAS_OPTIONS_AVAILABLE["jupyter"]:
     from ..layouts._frame._ipywidget_toolbar import IpywidgetImageWidgetToolbar
 
 if CANVAS_OPTIONS_AVAILABLE["qt"]:
     from ..layouts._frame._qt_toolbar import QToolbarImageWidget
 
+# How many dimensions to display a single image
+# If key is 0 this means no RGB (so each image is 2 dims). If key is 1, it's 3
+RGB_DIM_MAP = {0: 2, 1: 3}
```
kushalkolar, line 20: `IMAGE_DIM_COUNTS = {"gray": 2, "rgb": 3}`

and then map `rgb: list[bool]` to `IMAGE_DIM_COUNTS`

fastplotlib/widgets/image.py
```diff
@@ -328,17 +405,32 @@ def __init__(
...
-                self._data: List[np.ndarray] = data
-                self._ndim = self.data[0].ndim  # all ndim must be same
+                self._num_img_dims = [RGB_DIM_MAP[0] for i in range(len(data))]
+                for i in rgb_disp:
+                    if 0 <= i < len(data):
+                        self._num_img_dims[i] = RGB_DIM_MAP[1]
+
+                self._num_scrollable_dims = self._compute_and_validate_formats()
+
+                # Compute the largest dimension
```
kushalkolar, line 424: you mean "compute the largest number of dims" ?

fastplotlib/widgets/image.py
```diff
@@ -692,34 +647,36 @@ def _process_indices(
...
+        data_ix = None
+        for i in range(len(self.data)):
+            if self.data[i] is array:
+                data_ix = i
+                break
 
         numerical_dims = list()
+        curr_ndim = self.data[data_ix].ndim
+        indexer = [slice(None)] * curr_ndim
+        curr_scrollable_format = SCROLLABLE_DIMS_ORDER[
```
kushalkolar, line 668: put a comment that this maps from n_scrollab_dims to one of: `"", "t", or "tz"`

generally explain these lines with line-by-line comments

kushalkolar (changes requested):

fastplotlib/widgets/image.py
```diff
@@ -874,15 +833,21 @@ def set_data(
                     f"does not equal current data ndim {current_array.ndim}"
                 )
 
+            # Computes the number of scrollable dims and also validates new_array
+            new_scrollable_dims = self._get_n_scrollable_dims(new_array, self.rgb[i])
+
+            if self.n_scrollable_dims[i] != new_scrollable_dims:
+                raise ValueError(f"Scrollable dims do not match")
```
kushalkolar, line 840:
```suggestion
                raise ValueError(f"number of dimensions of data arrays must match number of dimensions of existing data arrays")
```

I think this is what you mean right? because we can't create new sliders etc.

fastplotlib/widgets/image.py
```diff
@@ -894,21 +859,32 @@ def set_data(
...
+                    elif max_lengths[scroll_dim] != new_length:
+                        raise ValueError(
+                            f"New arrays have differing values along dim {scroll_dim}"
+                        )
 
             # set histogram widget
-            subplot.docks["right"]["histogram_lut"].set_data(
-                new_array, reset_vmin_vmax=reset_vmin_vmax
-            )
+            if self.histogram_widget:
```
kushalkolar, line 878: private variable

fastplotlib/widgets/image.py
```diff
@@ -894,21 +859,32 @@ def set_data(
...
+                subplot.docks["right"]["histogram_lut"].set_data(
+                    new_array, reset_vmin_vmax=reset_vmin_vmax
+                )
 
         # set slider maxes
         # TODO: maybe make this stuff a property, like ndims, n_frames etc. and have it set the sliders
         for key in self.sliders.keys():
+            print("key is {}".format(key))
+            print("curr max is {}".format(max_lengths[key]))
```
kushalkolar, line 887:
```suggestion
```

I guess this was for debugging :joy:

kushalkolar (changes requested):

apasarkar:
@kushalkolar It seems the tests for pre-existing cells in image_widget_test.ipynb are failing. The rgba_to_rgb function seems to complain that the input has 2 indices but 3 indices are indexed etc. Let me know what you think, maybe we need to update the rgba_to_rgb function? 

Once we have that fixed, I can re-run to generate + update new screenshots.

kushalkolar:
I've made this a full PR from the draft so the CI will run on GitHub actions, we can check the result tomorrow.

fastplotlib/widgets/image.py
```diff
@@ -184,6 +197,56 @@ def current_index(self) -> Dict[str, int]:
...
+        ----------
+        curr_arr: np.ndarray
+            np.ndarray or a list of array-like
+
+        rgb: bool
+            True if we view this as RGB(A) and False if grayscale
+
+        Returns
+        -------
+        list[int]
```
kushalkolar, line 224:
```suggestion
        int
```
@apasarkar I'm editing this in-place, let me know if it's correct.

PR #467, opened 2024-03-31
add `References` object to `PlotArea`, other cleanup, better garbage collection

### LINKED ISSUE #463: refactor graphic, selector and legend instances dicts in `_plot_area`
Something like:

```python

class References:
  graphics: Dict[str, Graphic] = dict()
  selectors = dict()
  legends = dict()

REFERENCES = References()
```

kushalkolar:
closes #463

Adds new `References` object for which there is a single global instance for one instance of Python.

This helps segregate `Graphic`, `Selector`, `Legend` and any other objects we might want to have in the future. 

start using prefix `_fpl_*` for any attributes/methods that are for fpl internal use only, to distinguish them from actual private attributes

Moved selector `_fpl_cleanup` to `Graphic._fpl_cleanup`

Need to add tests for:
- [x] remove and adding things from `PlotArea`
- [x] garbage collection, maybe a starting point: https://github.com/fastplotlib/fastplotlib/blob/0cbe36802431c2928394fd33004804d50018a87d/examples/garbage_collection.ipynb

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -96,10 +96,16 @@ class Graphic(BaseGraphic):
 
     @name.setter
     def name(self, name: str):
+        if self.name == name:
+            return
+
         if not isinstance(name, str):
             raise TypeError("`Graphic` name must be of type <str>")
+
         if self._plot_area is not None:
             self._plot_area._check_graphic_name_exists(name)
+
+        self._name = name
 
     @property
     def world_object(self) -> WorldObject:
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -166,6 +166,9 @@ class Graphic(BaseGraphic):
         """Return the children of the WorldObject."""
         return self.world_object.children
 
+    def _fpl_add_plot_area_hook(self, plot_area):
+        self._plot_area = plot_area
+
     def __setattr__(self, key, value):
         if hasattr(self, key):
             attr = getattr(self, key)
@@ -192,7 +195,7 @@ class Graphic(BaseGraphic):
 
         return False
 
-    def _cleanup(self):
+    def _fpl_cleanup(self):
         """
         Cleans up the graphic in preparation for __del__(), such as removing event handlers from
         plot renderer, feature event handlers, etc.
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal
+from typing import Any, Literal, TypeAlias
 import weakref
 from warnings import warn
 from abc import ABC, abstractmethod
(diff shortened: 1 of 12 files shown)


kushalkolar:
Using the linear region selectors notebook to try adn understand what holds onto references:

image

kushalkolar:
A base `Graphic._fpl_cleanup()` that removes all possible event handlers on the renderer and world objects seems to work, tested with `LinearRegionSelector`

image

fastplotlib/graphics/_base.py
```diff
@@ -187,23 +199,52 @@ def __eq__(self, other):
...
+
+        feature_names = getattr(self, "feature_events")
+        for n in feature_names:
+            fea = getattr(self, n)
+            fea.clear_event_handlers()
 
     def __del__(self):
         self.deleted = True
-        del WORLD_OBJECTS[self.loc]
+        del WORLD_OBJECTS[self._fpl_address]
```
clewis7, line 247: does this method not need to call `_fpl_cleanup`?

fastplotlib/graphics/_base.py
```diff
@@ -450,7 +492,7 @@ def graphics(self) -> np.ndarray[Graphic]:
         """The Graphics within this collection. Always returns a proxy to the Graphics."""
         if self._graphics_changed:
```
clewis7, line 493: what was the reason for having this bool?

fastplotlib/graphics/_base.py
clewis7: oh wait, I see it getting called in subclasses

fastplotlib/graphics/selectors/_base_selector.py
```diff
@@ -136,8 +137,10 @@ def _add_plot_area_hook(self, plot_area):
 
         for fill in self._fill:
             if fill.material.color_is_transparent:
-                pfunc_fill = partial(self._check_fill_pointer_event, fill)
-                self._plot_area.renderer.add_event_handler(pfunc_fill, "pointer_down")
+                self._pfunc_fill = partial(self._check_fill_pointer_event, fill)
```
clewis7, line 140: you love partials lol

clewis7 (commented):
LGTM, `REFERENCES` class very elegant

fastplotlib/graphics/_base.py
```diff
@@ -11,9 +11,12 @@
 
 from ._features import GraphicFeature, PresentFeature, GraphicFeatureIndexable, Deleted
 
+
+HexStr: TypeAlias = str
```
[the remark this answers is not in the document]
kushalkolar: this is just a type alias, they were added in py3.10 https://peps.python.org/pep-0613/

fastplotlib/graphics/_base.py
kushalkolar: `_fpl_cleanup()` has to be called explicitly before in order to allow `__del__()` to be called, `__del__` gets called only when the object is actually gc.

fastplotlib/graphics/_base.py
kushalkolar: prevent remaking the array of proxies which can take a few millisecond

Issue #456, opened 2024-03-28
events refactor


kushalkolar:
From today's meeting, superseeds #345, #182, #100. Much of the ideas from those issues either belong in examples or a neuro-widget library @clewis7 

1. I think we will keep a very simple event system in fastplotlib. I will explore `observ` and see whether it makes sense to incorporate it within fastplotlib or not. Otherwise we will make a very simple event system in fastplotlib and give users examples of how to use observ or other libs for more complex event handling
2. Using what Almar and Korijn call a store model is probably a better paradigm than circular events when those types of interactions are required. Have a concept of a "state", for example "selected cell" and events modify that state. This way you don't have to worry about circular events. For example, if you click on a cell it modifies "selected cell", if you click on a temporal component it modifies "selected cell" etc.
3. Pull out the `Interaction` class from `fastplotlib` and put it into a new neurowidgets that Caitlin will work on.

Proposed basic API for adding event handlers suggested by Almar:

```python
graphic.add_event_handler("feature", callable)

# could also be a pygfx event and use the same API?
graphic.add_event_handler("click", callable)
```

clewis7:
Just to clarify...

Under this model we are completely revamping the interactivity abstraction layer to handle all interactivity using the `add_event_handler` function? A user will always have to pass a callable defining the behavior they want? 

What about for something where you are linking graphic features together? You would still need to pass a callable? 

For example:
```
xs = np.linspace(-10, 10, 100)
# sine wave
ys = np.sin(xs)
sine = np.dstack([xs, ys])[0]

# cosine wave
ys = np.cos(xs) + 5
cosine = np.dstack([xs, ys])[0]

plot_l = Plot()

# plot sine wave, use a single color
sine_graphic = plot_l.add_line(data=sine, thickness=5, colors="magenta")

# you can also use colormaps for lines!
cosine_graphic = plot_l.add_line(data=cosine, thickness=12, cmap="autumn")

sine_graphic.link(
    "click",
    target=cosine_graphic,
    feature="colors", 
    new_data="white", 
)
```

We would be removing the `link` function in favor of:
```
def callable():
   cosine_graphic.colors = "white"

sine_graphic.add_event_handler("click", callable)
```

kushalkolar:
We would completely remove the `link` functionality from fastplotlib in favor of just keeping the graphic-features events system (and pygfx events system), in favor of a state-management model instead of circular events.

[...]

Issue #449, opened 2024-03-24
Get rid of heatmap tiling 


kushalkolar:
If there is no longer a wgpu texture dimension limit. Test on all OSes and backends.

Issue #369, opened 2023-11-15
Choose dtype for isolated buffer


kushalkolar:
I just noticed when using `uint16` imaging data along with a `frame_apply` function in `ImageWidget`: `lambda frame: frame - movie_mean`, which results in negative values that overflow with numpy `uint16` without raising warnings or exceptions. We could either always use `float32` as the isolated buffer for `ImageGraphic` and/or allow choosing the dtype for the isolate buffer.

kushalkolar:
I propose always casting to `float32` for the `ImageDataFeature` buffer, even simple things like Gaussian filters on uint8 or uint16 images result in overflow and meaningless results. 

without casting:
![image](https://github.com/fastplotlib/fastplotlib/assets/9403332/792ab6d3-971e-4414-a1d3-0a96a4b96735)

with casting:
![image](https://github.com/fastplotlib/fastplotlib/assets/9403332/f9b92dba-5406-431b-a85c-5d363706ba3f)

I do not think we should expect users to worry about the limitations of the dtype of their raw image data if they are using `ImageWidget` with frame_apply functions etc. to interactively play with different types of processing :smile: So we should make casting to `float32` automatic, unless there are downsides I cannot think of?

clewis7:
Sounds good to me...I think `float32` gives enough flexibility

ArjunPutcha:
Agreed. Might still be useful to allow users to pass in dtype if necessary with default float 32. Some image data I acquire with sensor is uint16 or uint8 by default.

kushalkolar:
I'm not sure how useful it would be to allow users to set the dtype of the
internal buffer. uint isn't useful for any basic image processing. Do you
have an example use case?

ArjunPutcha:
Most image data from sensors I typecast as uint 16 (in firmware) - I'm pretty sure some national instruments tools I use also do the same; generally I want to assess the raw data first to see if analog filtration is going well, not do any digital signal processing, so I'd just read in the data and visualize directly (no need for any additional typecasting).

kushalkolar:
ah ok, so in this case you would still want to use an isolated buffer but of the same type so it's 1-1 copying into the buffer.

ArjunPutcha:
[...]

Issue #316, opened 2023-10-11
add features section for doc page for each graphic


kushalkolar:
I just noticed we don't put the feature docs in the main doc page for each graphic, they are only within the constructor doc pages: https://github.com/fastplotlib/fastplotlib/blob/master/docs/source/api/graphics/ImageGraphic.rst

PR #473, opened 2024-04-03
graphic features refactor

### LINKED ISSUE #389: consider descriptor pattern for `GraphicFeature`.
While playing around with some examples, I saw this type error reported by mypy:

Screenshot 2023-12-04 at 12 01 06 PM

I dug into a bit and learned about the `Graphic`/`GraphicFeature` relationship.  Apologies if you already know this and specifically decided against it... but the general pattern of an abstract `_set` method on `GraphicFeature` objects that retain a pointer to some parent object looks very much like it could be replaced/improved with [python descriptors](https://docs.python.org/3/howto/descriptor.html).  

For example, the current pattern (highly simplified) is:

```python
class GraphicFeature:
    def __init__(self, parent):
        self._parent = weakref.proxy(parent)

    def _set(self, value):
        raise NotImplementedError


class SomeFeature(GraphicFeature):
    def _set(self, value):
        print(f"setting {self.name!r} on {self._parent} to {value!r}")


class Graphic:
    def __setattr__(self, key, value):
        if hasattr(self, key):
            attr = getattr(self, key)
            if isinstance(attr, GraphicFeature):
                attr._set(value)
                return

        super().__setattr__(key, value)


class SomeGraphic(Graphic):
    def __init__(self) -> None:
        self.feature = SomeFeature()
```

with the descriptor pattern, that all reduces to 

```python
class SomeFeature:
    def __get__(self, instance, owner):
        ...

    def __set__(self, obj, value):
        print(f"setting feature on {obj} to {value!r}")


class SomeGraphic:
    feature = SomeFeature()

```

and when you set the `feature` attribute on an instance of `SomeGraphic`,  then `SomeFeature.__set__` will be called with the instance and value:

```python
In [18]: g = SomeGraphic()

In [19]: g.feature = 1
setting feature on <__main__.SomeGraphic object at 0x106c33750> to 1
```

this also works very nicely for typing:

Screenshot 2023-12-04 at 12 11 15 PM

note that it neither complains when setting  `.feature = [1,2,3]`, and it also recognizes that accessing `.feature` will return a `str` (i.e. it's been coerced from list to some internal value), based on the type hints of `__set__` and `__get__`

kushalkolar:
WIP 

closes #389, #489, #499, #456, #369, #322, #316, #309, #283, #174, #147, #126

Aim to also address: #507, #449, #123, #483, #214

# Remaining TODOs
- [ ] Update `Graphic._fpl_cleanup()`, must be possible to simplify this now that a `Graphic` instance keeps track of all the event handlers. Probably possible to have a simple `Graphic.clear_event_handlers()`
- [ ] Clear any pygfx rendering engine event handlers in `Graphic._fpl_cleanup()`
- [ ] When sharing buffers, events are also shared. When deleting a graphic from which the event handler was originally added, we need to think about how to deal with the event handlers since other graphics share that buffer :thinking: 
- [ ] test attach/detach buffers
- [x] Use tiling for `ImageGraphic`
- [ ] refactor `GraphicCollection`, make it possible to get values directly from a collection, decide:
  - `line_stack.colors`, or `line_stack[:].colors`? I think either should be possible with a `@property`
- [ ] more `GraphicCollection` implementation:
  - cleanup implementation of: `line_stack.colors = "r"`, `line_stack[50:].colors = "r"`, `line_stack[50:].colors[::2] = "r"`
  - likewise, getters should work for all of these, `line_stack[:10].colors[10:30]` should return a list of 10 arrays of shape `[20, 4]`
  - similar for non-sliceable graphics and those with uniform buffer for colors, ex:
    -  `line_stack.name = [f"line-{i}" for i in range(len(line_stack))]` 
- [ ] update `TextGraphic` with the feature changes.
- [ ] All tests

# Summary

## Graphic Features
- [x] Changes how our Graphic Features are implemented, makes it easier for static type checking tools to understand them, slightly changes how they behave, see #389 for the original inspiration of this refactor, thanks @tlambert03 !!
- [x] Modifes how events handlers can be added to a graphic, #456
- [x] Makes the feature events more uniform across features, and simpler, #174
- [x] UniformBuffer can be used for lines and scatters, cannot be shared between graphics but maybe we can implement this later.
- [x] Adds new `GraphicFeatures`, basically every graphic property which makes sense to be evented is a `GraphicFeature`, see #489. These features are common among all graphics:
    * `Name`: `str`, an optional name for the graphic
    * `Offset`: `np.ndarray, [x, y, z]`, offset from the origin, NOT related to the data shown in the graphic, #499
      * example: `LineStack`, lines are stacked along the y-axis w.r.t. to the rendering engine. This is done by incrementing the `Offset`. We DO NOT add to the actual data buffer that the graphic uses!
      * `Rotation`: `np.ndarray`, quaternion
      * `Visible`: `bool`
      * `Deleted`: `bool` - indicates if a a `PlotArea` has requested the deletion of a graphic. This is useful because you can add event handlers for `"deleted"`, and this event handler can perform any necessary cleanup (disconnted events, etc.) so that the RAM for the graphic (i.e. the buffers) can be cleared.
- [x] vmin, vmax are now their own features, no longer properties of `ImageCmap`. It was just cleaner this way and makes more sense because RGBA images have no cmap, but they can have vmin and vmax.
- [x] Ownership of things is much cleaner. Graphic feature instances no longer keep the parent graphic as an attribute. Similarly, `Graphics` do not keep `collection_index` as an attribute.

### Graphic Features that are sliceable, subclasses `BufferManager`
- [x]  Makes it possible to share buffers across graphics :smile:  #126
  * buffers can be attached/detached, not yet tested. `Graphic.attach_feature("feature_name")` and `Graphic.detach_feature("feature_name")`
- [x]  All sorts of indexing works, including negative indexing, 16 test cases: https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/tests/utils.py 
- [x]  The slice parser is much much simpler now, no more :spaghetti: 
- [x]  Only one color parser used for constructor and setter, also much simpler, less spaghetti
- [x]  A bunch of parameterized tests for slicing and setting these buffers
- [x]  `BufferManager.__init__` manages whether the user's passed array should be isolated or not, the default is to create an isolated buffer. Makes much more sense to manage it here rather than in each `SubGraphic.__init__`. #147
- [x]  Everything just gets casted to `float32`, #369

### Selector tools
- [x]  Deleted Sychronizer, this is something that a user should manage on their own, tailored to their usecase.
- [x]  The logic for getting the selected, index, indices, and selected data is much simpler.
  * All selector graphics set their offset to match the parent graphic's offset. This greatly simplifies things, no more adding/subtracting offsets to map from world-space -> data space :smile: 
  * The geometry of the selector tool's underlying `WorldObjects` are used to get the position/region under the selection. Since the offsets match those of the parent `Graphic`, we don't need to do anything more.
  * the `"selection"` feature now returns value(s) in **data space, NOT INDICES**. This makes more sense since for example, you would want to set or get the bounds of a `LinearRegionSelector` in terms of actual `(data xmin, data xmax)`. Likewise for `LinearSelector`.
- [x] Selectors behave a bit better, example: `LinearRegionSelector` behaves more intuitively when at/close to the limits. It is not actually possible to "smash" the selector right up against the limits and it actually touches the limit, whereas before if the delta was a bit larger it would be stuck a few pixels away from the limit.
- [x]  The events are simplified (see below)

#### Basic features (no buffer)

```python
# before this PR, setting and getting a feature value
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness()
3.5

# this PR, getter & setter are symmetric when possible
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness
3.5
```

#### Sliceable features (manages a buffer)

```python
# setting and getting the "entire value"
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

# this PR, similar but value can be directly access if desired
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

>> line_graphic.colors.value
<returns the actual array>
```

**Sharing buffers** :smile: !

```python
# let's say you have a line with some vertex colors
>>> line1.colors
<returns array for red>

>>> line2 = subplot.add_line(data, colors=line1.colors)  # share buffer :D !

# modify colors
>>> line2.colors[50:] = "b"

# change of colors is reflected in BOTH graphics :D 
```

The implementation is very simple, for example in `PositionsGraphic.__init__`:

```python
class PositionsGraphic(Graphic):
  def __init__(self, data, colors, ...):
    if isinstance(data, VertexPositions):
        self._data = data
    else:
        self._data = VertexPositions(data, isolated_buffer=isolated_buffer)
```

As we can see, it actually shares the `BufferManager` (i.e. feature) instance. This is simpler and makes it so changing the buffer causes events to be emitted even if it's from another graphic. In the above example, if an event handler is added to `line1`, i.e. `line1.add_event_handler(<callable>, "colors")`, then changing the colors of `line2` would also trigger the event which I assume is what you would want since `line1` colors have _also_ changed.

### Events

```python
# before this PR
line_graphic.thickness.add_event_handler(<callable>)

# this PR
line_graphic.add_event_handler(<callable>, "thickness")

# can also use decorators :D

@line_graphic.add_event_handler("thickness")
def thickness_changed_handler(ev):
  pass

# exactly the same for features that subclass BufferManager
@line_graphic.add_event_handler("colors", "data")
def line_colors_or_data_changed(ev):
  if ev.type == "colors":
    print(
      f"colors changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )
  if ev.type == "data":
    print(
      f"data changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )

# rendering engine events added in exactly the same way
@image_graphic.add_event_handler("click")
def image_clicked(ev):
  ev.graphic # the graphic that caused the event
```

The event adding and handling is managed here:

https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/fastplotlib/graphics/_base.py#L173-L278

Before this PR, the event info was a mess that we probably don't need to dig into :laughing: . Now every `FeatureEvent` inherits from `pygfx.Event` and has the following attributes:

| attribute  | type        | description                                   |
|------------|-------------|-----------------------------------------------|
| type       | `str`         | name of the event, ex: "colors"                  |
| graphic    | `Graphic`     | graphic instance that the event is from       |
| info       | `dict`        | event info dictionary (see below)             |
| target     | `pygfx.WorldObject` | pygfx rendering engine object for the graphic |
| time_stamp | `float`       | time when the event occured, in ms            |

The "info" dict for all "simple" features has one key, "value", which is the user passed new value. Example if `line_graphic.name` is changed, then the info dict will be: `{"value": "new_name_str"}`

#### `BufferManager` example info dicts:

**colors** (vertex colors, line or scatter)

| dict key | value type | value description |
|-------|---------------------------|-------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which colors were indexed/sliced |
| value | `np.ndarray` | new color values for points that were changed, shape is \[n_points_changed, RGBA\] |
| user_value | `str` \| `np.ndarray` \| `tuple[float]` \| `list[float]` \| `list[str]` | user input value that was parsed into the RGBA array |

**data** (points data, line or scatter)

| dict key | value type | value description |
|------|--------------------------|----------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which vertex positions data were indexed/sliced |
| value | `np.ndarray` \| `float` \| `list[float]` | new data values for points that were changed, shape depends on the indices that were set |

#### Selectors

Adding event handlers to a selector is identical to how they're added for "regular" graphics:

```python
# one of the selectors will change the line colors when it moves
@selector.add_event_handler("selection")
def set_color_at_index(ev):
    # changes the color at the index where the slider is
    ix = ev.get_selected_index()
    g = ev.graphic.parent
    g.colors[ix] = "green"
```

**Selector feature event structure**

`LinearSelector` `"selection"`

**additional event attributes:**

| attribute          | type     | description                        |
|--------------------|----------|------------------------------------|
| get_selected_index | callable | returns indices under the selector |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new x or y value of selection |

`LinearRegioonSelector`

**additional event attributes:**

| attribute            | type     | description                        |
|----------------------|----------|------------------------------------|
| get_selected_indices | callable | returns indices under the selector |
| get_selected_data    | callable | returns data under the selector    |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new `[min, max]` of selection |

This is how it's implemented:

https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/fastplotlib/graphics/_features/_selection_features.py#L57-L76

`selector` is the `LinearSelector` (in this case) or `LinearRegionSelector` that the feature manages, and it has the `get_selected_index`, or `get_selected_indices` and `get_selected_data` methods for `LinearRegionSelector`

## Graphics

Some major refactor to graphics not mentioned above:

* line and scatter graphics now inherit from a common `PositionsGraphic` that handles things that are common between lines and scatters (data, colors, etc.)


kushalkolar:
Working on this, Graphic features should have no access to `collection_index`. When we emit events, we can have the graphic accessible and then the collection index of the graphic can be accessed if required.

kushalkolar:
ok so `PointsDataFeature` working, and we have negative slicing now :partying_face:

kushalkolar:
I should double check to make sure we don't get any off by 1 errors in the slicing and with `size` in `update_range`

kushalkolar:
Ideas for event adding:

```python
# graphic base class
Graphic:
  def add_event_handler(func, "event-name")
```

Where "event-name" is either a feature name or pygfx event name (ex: "click" etc.). And emitted feature events use a `pick_info` dict that holds event information just like pygfx events. Might make sense to subclass `pygfx.Event` here for feature events, will see.

kushalkolar:
@almarklein I've arrived at this parser that takes the key used to index the buffer and auto determines the offset & size to mark for upload, do you think this would be useful to have directly in pygfx?

snippet:

```python
    def _update_range(self, key: int | slice | np.ndarray[int | bool] | list[bool | int] | tuple[slice, ...]):
        """
        Uses key from slicing to determine the offset and
        size of the buffer to mark for upload to the GPU
        """
        # number of elements in the buffer
        upper_bound = self.value.shape[0]

        if isinstance(key, tuple):
            # if multiple dims are sliced, we only need the key for
            # the first dimension corresponding to n_datapoints
            key: int | np.ndarray[int | bool] | slice = key[0]

        if isinstance(key, int):
            # simplest case
            offset = key
            size = 1

        elif isinstance(key, slice):
            # parse slice
            start, stop, step = key.indices(upper_bound)

            # account for backwards indexing
            if (start > stop) and step < 0:
                offset = stop
            else:
                offset = start

            # slice.indices will give -1 if None is passed
            # which just means 0 here since buffers do not
            # use negative indexing
            offset = max(0, offset)

            # number of elements to upload
            # this is indexing so do not add 1
            size = abs(stop - start)

        elif isinstance(key, (np.ndarray, list)):
            if isinstance(key, list):
                # convert to array
                key = np.array(key)

            if not key.ndim == 1:
                raise TypeError(key)

            if key.dtype == bool:
                # convert bool mask to integer indices
                key = np.nonzero(key)[0]

            if not np.issubdtype(key.dtype, np.integer):
                # fancy indexing doesn't make sense with non-integer types
                raise TypeError(key)

            if key.size < 1:
                # nothing to update
                return

            # convert any negative integer indices to positive indices
            key %= upper_bound

            # index of first element to upload
            offset = key.min()

            # number of elements to upload
            # add 1 because this is direct
            # passing of indices, not a start:stop
            size = np.ptp(key) + 1

        else:
            raise TypeError(key)

        self.buffer.update_range(offset=offset, size=size)
```

kushalkolar:
I need to do a thing in the tests where we just draw frames on demand to test that the right things are in `buffer._gfx_pending_uploads`, and that after a frame is draw this list is cleared and take a screenshot and see that it looks right.

tests/test_buffer_manager.py
```diff
@@ -0,0 +1,159 @@
...
+            indices = [2, 4, 6]
+
+        case 9:
+            # positive, negative, negative
+            s = slice(8, -9, -2)
+            indices = [8, 6, 4, 2]
+
+    others = [i for i in a if i not in indices]
+
+    return {"slice": s, "indices": indices, "others": others}
```
kushalkolar, line 82: @apasarkar do we have other ways of slicing that applies to lines/scatters along the first (n_datapoints, or n_timepoints) direction?

tests/test_buffer_manager.py
```diff
@@ -0,0 +1,222 @@
...
+
+    s = name
+    a = np.array(color)
+    l = list(color)
+    t = tuple(color)
+
+    return [s, a, l, t]
+
+
+def generate_slice_indices(kind: int):
```
kushalkolar, line 25: @apasarkar do we have other ways of slicing that applies to lines/scatters along the first (n_datapoints, or n_timepoints) direction?

tests/test_buffer_manager.py
```diff
@@ -0,0 +1,222 @@
...
+            indices = [1, 4, 8]
+
+        case 15:
+            # array indices
+            s = np.array([1, 4, -7, 9])
+            indices = [1, 4, 3, 9]
+
+    others = [i for i in a if i not in indices]
+
+    return {"slice": s, "indices": indices, "others": others}
```
kushalkolar, line 117: @apasarkar do we have other ways of slicing that applies to lines/scatters along the first (n_datapoints, or n_timepoints) direction?

kushalkolar:
Since this will easily allow sharing a buffer between graphics I should probably have a method for detatching from a buffer and acquiring a new one.

almarklein:
> @almarklein I've arrived at this parser that takes the key used to index the buffer and auto determines the offset & size to mark for upload, do you think this would be useful to have directly in pygfx?

Ah thanks! I added a link to your comment in https://github.com/pygfx/pygfx/issues/600. I'm sure we can re-use some of that.

kushalkolar:
should make text properties a feature too

kushalkolar:
I'm changing the selectors so that `selector.selection` returns the x or y axis value, i.e. the location in data space, as opposed to data index. I think this makes sense and will usually be what users want. I think it also makes the parsing easier.

clewis7 (commented):
Started reviewing things...

Things with lines appear to be working well - shared buffers, event handling 

Images are a bit wonking - updating data, changing cmap, etc.

fastplotlib/graphics/_base.py
```diff
@@ -35,30 +37,72 @@
...
+        self._visible.set_value(self, value)
+
+    @property
+    def deleted(self) -> bool:
+        """used to emit an event after the graphic is deleted"""
+        return self._deleted.value
+
+    @deleted.setter
+    def deleted(self, value: bool):
+        self._deleted.set_value(self, value)
```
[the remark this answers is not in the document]
kushalkolar: This feature is not meant to be set by the user, it's used to indicate that a graphic has been requested to be deleted by a `PlotArea`

fastplotlib/graphics/_base.py
```diff
@@ -269,6 +362,165 @@ def rotate(self, alpha: float, axis: Literal["x", "y", "z"] = "y"):
...
+            self.world_object.geometry.positions = self._sizes.buffer
+            self._sizes._shared -= 1
+
+    def attach_feature(self, feature: VertexPositions | VertexColors | PointsSizesFeature):
+        if isinstance(feature, VertexPositions):
+            # TODO: check if this causes a memory leak
+            self._data._shared -= 1
+
+            self._data = feature
+            self._data._shared += 1
```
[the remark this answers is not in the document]
kushalkolar: these are different instnaces. subtract from previous instance, and then attach other instance and add

kushalkolar:
Thought from Billy: make the simplest use cases default kwargs, ex: uniform_colors for line and scatter.

clewis7:
TODO: need to fix HLUT in `ImageWidget` to mirror changes in `LinearRegionSelector`

fastplotlib/graphics/_features/_positions_graphics.py
```diff
@@ -0,0 +1,376 @@
...
+        self._vertex_colors = vertex_colors
+        self._cmap_name = cmap_name
+        self._cmap_values = cmap_values
+
+        if self._cmap_name is not None:
+            if not isinstance(self._cmap_name, str):
+                raise TypeError
+            if self._cmap_values is not None:
+                if not isinstance(self._cmap_values, np.ndarray):
+                    raise TypeError
```
clewis7, line 310: In a previous example with lines we had setting the `cmap_values` from a list, with this change we would no longer be supporting

for now, I changed the example to just cast the list to an array

kushalkolar:
see if this can reopen

PR #479, opened 2024-04-07
`Gridplot` -> `Figure`

### LINKED ISSUE #457: `GridPlot` -> `Figure` as the only user-facing layout
Layouts currently has a lot of multiple inheritance, this is partly due to us having a separate `Plot` and `GridPlot` class. After meeting with Almar I propose:

1. New `Figure` class which merges `GridPlot`, `Frame` and `RecordMixin`. 
2. Remove `Plot` in favor of creating a `Figure` with shape `(1, 1)`, i.e. one subplot. We can have `shape=(1, 1)` as default for `Figure`

@almarklein Does this make it easier to have axes etc.? Anything else to add?

kushalkolar:
closes #457

This PR breaks everything :smile: 

- [x] `Gridplot` renamed to `Figure`
- [x] `Plot` is removed in favor of just doing `Figure()` since the default `shape=(1, 1)`, i.e. "gridplot" with a single subplot
- [x] `Frame` merged into `Figure`
- [x] `RecordMixin` -> `FigureRecorder`
  * `record` is now an instance of `Figure`, use composition instead of inheritance, I haven't tested that this works yet, right now it's still a very experimental feature
- [x] `_frame` module renamed to `output`, some reorganization of what is public/private within that module to make it more logical
- [x] updated all the examples and tests
- [x] update `ImageWidget`

Still need to do this for `ImageWidget`, waiting for #459 , rest of it is good to go and tests are passing on my end :partying_face:


kushalkolar:
when reviewing don't be scared by the size of the diff, it's mostly stuff being moved around, and the changes to the examples are mostly renaming stuff from plot -> figure, gridplot -> figure :smile:

fastplotlib/layouts/_figure.py
```diff
@@ -97,7 +102,7 @@ def __init__(
         if names is not None:
             if len(list(chain(*names))) != len(self):
                 raise ValueError(
-                    "must provide same number of subplot `names` as specified by gridplot shape"
+                    "must provide same number of subplot `names` as specified by Figure shape"
```
clewis7, line 105:
```suggestion
                    "must provide same number of subplot `names` as specified by Figure `shape`"
```

clewis7 (approved):
LGTM, mostly just a big refactor but I think this cleans up a lot of the soup and better organizes the repo

kushalkolar:
Idea: if `figure.shape == (1, 1)`, should be make `figure.__getitem__` just pass to `subplot.__getitem__` so the user doesn't have to do `figure[0, 0]` :thinking:

PR #495, opened 2024-04-25
Dockerfile example

kushalkolar:
We should make some example dockerfiles, this one builds off of the nvidia cuda docker. It builds python3.11 from source and makes a venv, I think we can just keep this in here because if someone wants to use conda they can just replace that part and adjust the rest accordingly.

@apasarkar @clewis7 you guys are more experienced with this than me, let me know what you think.


kushalkolar:
Actually the better way to do this might be to put it on dockerhub with a few variations:
1. Barebones, only with `wgpu`
2. With `fastplotlib` installed using venv
3. With `fastplotlib` and mamba :man_shrugging:

clewis7 (approved):
also LGTM

clewis7:
> Actually the better way to do this might be to put it on dockerhub with a few variations:
> 
> 1. Barebones, only with `wgpu`
> 2. With `fastplotlib` installed using venv
> 3. With `fastplotlib` and mamba 🤷‍♂️

I think the bottom two make sense to give people the option of what they are most comfortable with. I'm not sure if we really need a barebones, I think it would not get used much.

kushalkolar:
A WGPU docker image could be useful in general for others. 

Also is there a way to use github actions to "publish" to docker-hub every time we merge into main and for every release?

PR #505, opened 2024-05-17
simple Figure.export()

### LINKED ISSUE #504: Figure.export() ?
What do you think about an `Figure.export()` method. I'm not sure because `Figure.renderer.snapshot()` exists, and then users can use their preferred library for saving that RGBA array to an image. And if they want to deal with the alpha channel in a special way they can do that. But I often see people searching the docs for "save" and "save figure" . We can have a simple `Figure.export()` `imageio` that takes the snapshot and does stuff in a default way? Just saves it to an png, or a jpg by stripping the alpha channel?

kushalkolar:
Simple `Figure.export()` method that saves a snapshot of the renderer using `imageio`

closes #504


clewis7 (dismissed):
LGTM

clewis7 (dismissed):

clewis7 (approved):

clewis7 (approved):
LGTM

kushalkolar:
idk why this is being weird, going to do a new PR

PR #511, opened 2024-06-04
Graphic features refactor

### LINKED ISSUE #123: `GraphicCollection` feature data is not accessible
It would be useful to access the feature data, for example:

image

Currently resorting to this:

```python
[plot_temporal["temporal"].graphics[ix].present() for ix in range(len(plot_temporal["temporal"].graphics))]
```

### LINKED ISSUE #126: Look into sharing storage buffers
When for example the colors are the same across millions of points.

### LINKED ISSUE #147: need to improve how we handle buffers
I just noticed that before #143 when we initialized any `GraphicFeature`, the data were being copied to a new array because of `astype(np.float32)` which has `copy=True` by default: https://github.com/kushalkolar/fastplotlib/pull/143/files#diff-e9a4f655a73cc26eb39868b9b4f44940cccfa247f6c433559cec77954383a198L46

 We need to think about how to properly deal with updating the data in these buffers when the arrays are read-only, such as numpy memmaps:

```python
>>> img_graphic.data._buffer.data

memmap([[ 14.957001 ,   8.239777 ,  24.973755 , ...,   2.7485046,
           5.922577 ,   3.677765 ],
        [ 91.104004 ,  39.25052  ,   6.083435 , ...,  38.935028 ,
          57.211304 ,  23.596008 ],
        [121.86125  ,  74.614716 ,  20.688904 , ...,  25.005035 ,
          73.836426 ,  12.158752 ],
        ...,
        [ 35.85089  ,  22.899536 ,  21.668213 , ...,  93.56639  ,
          38.996155 ,  32.658997 ],
        [ 15.931152 ,  12.423279 ,  10.606384 , ...,  64.8004   ,
          21.833801 ,   8.394348 ],
        [ 54.492004 ,   6.143982 ,  18.702332 , ...,  22.917175 ,
          15.810699 ,  30.137512 ]], dtype=float32)
```

Trying to assign, results in the following:

```python
>> img_graphic.data = 0
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [44], in <cell line: 1>()
----> 1 iw_cnmf.plot.graphics[0].data[0] = 0

File ~/Insync/gdrive/repos/fastplotlib/fastplotlib/graphics/features/_data.py:114, in ImageDataFeature.__setitem__(self, key, value)
    110 def __setitem__(self, key, value):
    111     # make sure float32
    112     value = to_gpu_supported_dtype(value)
--> 114     self._buffer.data[key] = value
    115     self._update_range(key)
    117     # avoid creating dicts constantly if there are no events to handle

ValueError: assignment destination is read-only
```

 But we are only showing a part of the entire large memmap in those cases so we could just set the texture data entirely if it's a lazy-loading type, maybe a nice way to do it would be to have something like:

```python
import fastplotlib
from numpy import memmap
from mesmerize_core.arrays import LazyArray

fastplotlib.config.lazy_array_types.append(memmap, LazyArray)
``

And then in the various `GraphicFeatureIndexable._set()` we can set `_buffer.data` entirely if they are one of those `lazy_array_types`.

### LINKED ISSUE #174: Cleanup GraphicFeatures events pick_info
The pick_info dicts are becoming a bit of a mess now and inconsistent. They should be consistent with certain keys that are available across all FeatureEvents. FeatureEvents should itself probably be a dataclass with fields that correspond to the minimum info that should be present across all FeatureEvents. The rest can be in a dict probably.

### LINKED ISSUE #214: probably get rid of `CollectionIndexer` and `CollectionFeature` in favor of numpy arrays for graphic collections

### LINKED ISSUE #283: scatter data doesn't work with `int` types, only with `float`
maybe this is just the positions geometry buffer, need to check with line too

### LINKED ISSUE #309: Move Selection `Feature` classes to the corresponding selector module
Right now the selection feature classes live inside `_features`, but they are quite strongly tied to their corresponding selector so we can just move it there. It's not like `PointsDataFeature` which is shared between lines and scatter graphics. This will also make the imports nicer.

### LINKED ISSUE #316: add features section for doc page for each graphic

### LINKED ISSUE #322: add test for cmap_values `@setter`

### LINKED ISSUE #369: Choose dtype for isolated buffer

### LINKED ISSUE #389: consider descriptor pattern for `GraphicFeature`.

### LINKED ISSUE #449: Get rid of heatmap tiling

### LINKED ISSUE #456: events refactor

### LINKED ISSUE #489: Scope of graphic features
Right now graphic features mostly map to pygfx buffers, except for `present`, `deleted` and `thickness` (for lines). I'm wondering what attributes we should and shouldn't make as "graphic features". For example, the current primitive legends take a `label: str` kwarg. If `Graphic.name` is a feature, the legend labels could automatically track any changes in the graphic name. But should other things like `rotation` and `position` also be a feature? Or should the users just use groups or make their own systems for keeping track of changes to things like `name`, `rotation`, `position`.

### LINKED ISSUE #499: rename `position_x`, `position_y`, `position_z`  to offsets
We should probably rename all the `position_<axis>` properties to offsets since that's where the `WorldObject` is _placed_ and does not indicate the buffer data, ex. positions geometry.

realized when doing #498 that this can get confusing

Maybe have:

```python
class Graphic
@offset.setter
def offset(self, offset: tuple[float, float, float]):
  x, y, z = offset
  self.world_object.world.x = x
  ...
```

### LINKED ISSUE #507: `ImageGraphic` should use tiling if necessarry, get rid of `HeatmapGraphic`
The only reason `HeatmapGraphic` exists is to display images with dims larger than the WGPU max texture size (8192, 8192). `_ImageTile` just adds the chunk index to the `pick_info` dict but we can instead just use the chunk indices to set the actual `index` position of events based on the chunk size.

https://github.com/fastplotlib/fastplotlib/blob/9a309ef1359244eb3791ddcfebc8b08b2697ae9f/fastplotlib/graphics/image.py#L323-L331

kushalkolar:
WIP 

closes #389, closes #489, closes #499, closes #456, closes #369, closes #322, closes #316, closes #309, closes #283, closes #174, closes #147, closes #126

 closes #507, closes #449, closes #123, closes #214

# Remaining TODOs
- [x] Use tiling for `ImageGraphic`
- [x] refactor `GraphicCollection`
- [x] more `GraphicCollection` implementation:
  - getters work
  - similar for non-sliceable graphics and those with uniform buffer for colors, ex:
    -  `line_stack.name = [f"line-{i}" for i in range(len(line_stack))]` 
- [x] update `TextGraphic` with the feature changes.
- [x] All tests

# Summary

## Graphic Features
- [x] Changes how our Graphic Features are implemented, makes it easier for static type checking tools to understand them, slightly changes how they behave, see #389 for the original inspiration of this refactor, thanks @tlambert03 !!
- [x] Modifes how events handlers can be added to a graphic, #456
- [x] Makes the feature events more uniform across features, and simpler, #174
- [x] UniformBuffer can be used for lines and scatters, cannot be shared between graphics but maybe we can implement this later.
- [x] Adds new `GraphicFeatures`, basically every graphic property which makes sense to be evented is a `GraphicFeature`, see #489. These features are common among all graphics:
    * `Name`: `str`, an optional name for the graphic
    * `Offset`: `np.ndarray, [x, y, z]`, offset from the origin, NOT related to the data shown in the graphic, #499
      * example: `LineStack`, lines are stacked along the y-axis w.r.t. to the rendering engine. This is done by incrementing the `Offset`. We DO NOT add to the actual data buffer that the graphic uses!
      * `Rotation`: `np.ndarray`, quaternion
      * `Visible`: `bool`
      * `Deleted`: `bool` - indicates if a a `PlotArea` has requested the deletion of a graphic. This is useful because you can add event handlers for `"deleted"`, and this event handler can perform any necessary cleanup (disconnted events, etc.) so that the RAM for the graphic (i.e. the buffers) can be cleared.
- [x] vmin, vmax are now their own features, no longer properties of `ImageCmap`. It was just cleaner this way and makes more sense because RGBA images have no cmap, but they can have vmin and vmax.
- [x] Ownership of things is much cleaner. Graphic feature instances no longer keep the parent graphic as an attribute. Similarly, `Graphics` do not keep `collection_index` as an attribute.

### Graphic Features that are sliceable, subclasses `BufferManager`
- [x]  Makes it possible to share buffers across graphics :smile:  #126
  * buffers can be attached/detached, not yet tested. `Graphic.attach_feature("feature_name")` and `Graphic.detach_feature("feature_name")`
- [x]  All sorts of indexing works, including negative indexing, 16 test cases: https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/tests/utils.py 
- [x]  The slice parser is much much simpler now, no more :spaghetti: 
- [x]  Only one color parser used for constructor and setter, also much simpler, less spaghetti
- [x]  A bunch of parameterized tests for slicing and setting these buffers
- [x]  `BufferManager.__init__` manages whether the user's passed array should be isolated or not, the default is to create an isolated buffer. Makes much more sense to manage it here rather than in each `SubGraphic.__init__`. #147
- [x]  Everything just gets casted to `float32`, #369

### Selector tools
- [x]  Deleted Sychronizer, this is something that a user should manage on their own, tailored to their usecase.
- [x]  The logic for getting the selected, index, indices, and selected data is much simpler.
  * All selector graphics set their offset to match the parent graphic's offset. This greatly simplifies things, no more adding/subtracting offsets to map from world-space -> data space :smile: 
  * The geometry of the selector tool's underlying `WorldObjects` are used to get the position/region under the selection. Since the offsets match those of the parent `Graphic`, we don't need to do anything more.
  * the `"selection"` feature now returns value(s) in **data space, NOT INDICES**. This makes more sense since for example, you would want to set or get the bounds of a `LinearRegionSelector` in terms of actual `(data xmin, data xmax)`. Likewise for `LinearSelector`.
- [x] Selectors behave a bit better, example: `LinearRegionSelector` behaves more intuitively when at/close to the limits. It is not actually possible to "smash" the selector right up against the limits and it actually touches the limit, whereas before if the delta was a bit larger it would be stuck a few pixels away from the limit.
- [x]  The events are simplified (see below)

#### Basic features (no buffer)

```python
# before this PR, setting and getting a feature value
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness()
3.5

# this PR, getter & setter are symmetric when possible
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness
3.5
```

#### Sliceable features (manages a buffer)

```python
# setting and getting the "entire value"
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

# this PR, similar but value can be directly access if desired
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

>> line_graphic.colors.value
<returns the actual array>
```

**Sharing buffers** :smile: !

```python
# let's say you have a line with some vertex colors
>>> line1.colors
<returns array for red>

>>> line2 = subplot.add_line(data, colors=line1.colors)  # share buffer :D !

# modify colors
>>> line2.colors[50:] = "b"

# change of colors is reflected in BOTH graphics :D 
```

The implementation is very simple, for example in `PositionsGraphic.__init__`:

```python
class PositionsGraphic(Graphic):
  def __init__(self, data, colors, ...):
    if isinstance(data, VertexPositions):
        self._data = data
    else:
        self._data = VertexPositions(data, isolated_buffer=isolated_buffer)
```

As we can see, it actually shares the `BufferManager` (i.e. feature) instance. This is simpler and makes it so changing the buffer causes events to be emitted even if it's from another graphic. In the above example, if an event handler is added to `line1`, i.e. `line1.add_event_handler(<callable>, "colors")`, then changing the colors of `line2` would also trigger the event which I assume is what you would want since `line1` colors have _also_ changed.

### Events

```python
# before this PR
line_graphic.thickness.add_event_handler(<callable>)

# this PR
line_graphic.add_event_handler(<callable>, "thickness")

# can also use decorators :D

@line_graphic.add_event_handler("thickness")
def thickness_changed_handler(ev):
  pass

# exactly the same for features that subclass BufferManager
@line_graphic.add_event_handler("colors", "data")
def line_colors_or_data_changed(ev):
  if ev.type == "colors":
    print(
      f"colors changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )
  if ev.type == "data":
    print(
      f"data changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )

# rendering engine events added in exactly the same way
@image_graphic.add_event_handler("click")
def image_clicked(ev):
  ev.graphic # the graphic that caused the event
```

The event adding and handling is managed here:

https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/fastplotlib/graphics/_base.py#L173-L278

Before this PR, the event info was a mess that we probably don't need to dig into :laughing: . Now every `FeatureEvent` inherits from `pygfx.Event` and has the following attributes:

| attribute  | type        | description                                   |
|------------|-------------|-----------------------------------------------|
| type       | `str`         | name of the event, ex: "colors"                  |
| graphic    | `Graphic`     | graphic instance that the event is from       |
| info       | `dict`        | event info dictionary (see below)             |
| target     | `pygfx.WorldObject` | pygfx rendering engine object for the graphic |
| time_stamp | `float`       | time when the event occured, in ms            |

The "info" dict for all "simple" features has one key, "value", which is the user passed new value. Example if `line_graphic.name` is changed, then the info dict will be: `{"value": "new_name_str"}`

#### `BufferManager` example info dicts:

**colors** (vertex colors, line or scatter)

| dict key | value type | value description |
|-------|---------------------------|-------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which colors were indexed/sliced |
| value | `np.ndarray` | new color values for points that were changed, shape is \[n_points_changed, RGBA\] |
| user_value | `str` \| `np.ndarray` \| `tuple[float]` \| `list[float]` \| `list[str]` | user input value that was parsed into the RGBA array |

**data** (points data, line or scatter)

| dict key | value type | value description |
|------|--------------------------|----------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which vertex positions data were indexed/sliced |
| value | `np.ndarray` \| `float` \| `list[float]` | new data values for points that were changed, shape depends on the indices that were set |

#### Selectors

Adding event handlers to a selector is identical to how they're added for "regular" graphics:

```python
# one of the selectors will change the line colors when it moves
@selector.add_event_handler("selection")
def set_color_at_index(ev):
    # changes the color at the index where the slider is
    ix = ev.get_selected_index()
    g = ev.graphic.parent
    g.colors[ix] = "green"
```

**Selector feature event structure**

`LinearSelector` `"selection"`

**additional event attributes:**

| attribute          | type     | description                        |
|--------------------|----------|------------------------------------|
| get_selected_index | callable | returns indices under the selector |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new x or y value of selection |

`LinearRegioonSelector`

**additional event attributes:**

| attribute            | type     | description                        |
|----------------------|----------|------------------------------------|
| get_selected_indices | callable | returns indices under the selector |
| get_selected_data    | callable | returns data under the selector    |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new `[min, max]` of selection |

This is how it's implemented:

https://github.com/fastplotlib/fastplotlib/blob/1d6adc6413e861b4a4ddc1f22fbe4c754b584a30/fastplotlib/graphics/_features/_selection_features.py#L57-L76

`selector` is the `LinearSelector` (in this case) or `LinearRegionSelector` that the feature manages, and it has the `get_selected_index`, or `get_selected_indices` and `get_selected_data` methods for `LinearRegionSelector`

## Graphics

Some major refactor to graphics not mentioned above:

* line and scatter graphics now inherit from a common `PositionsGraphic` that handles things that are common between lines and scatters (data, colors, etc.)


kushalkolar:
remaining tests todo:

1. texture array buffer for images, row/col indices and slice positions of the full data array, with square, tall, and wide arrays.
  - new screenshot tests for these when in graphics
2. image cmap, vmin, vmax, interpolations
  - new screenshot tests too for these when in graphics
3. uniform color and size features for lines and points, and screenshot tests
4. VertexCmap
  - screenshot tests exist for `LineGraphic`, need to do `ScatterGraphic`
5. Tests of making all graphics and changing all their features, this is to test the setters and getters in addition to the graphic features and buffers on their own.
6. Text, features by themselves, creating graphics & testing their features then, and screenshots
7. Common features, by themselves and within graphics with tests, can add to example gallery too (offset, rotation, visible).

kushalkolar:
Need to add arrow and dashing options for line, and marker option for scatter

fastplotlib/graphics/_collection_base.py
```diff
@@ -0,0 +1,277 @@
...
+
+        self.world_object.remove(graphic.world_object)
+
+        self._graphics_changed = True
+
+    def add_event_handler(self, *args):
+        raise NotImplementedError("Slice graphic collection to add event handlers")
+
+    def remove_event_handler(self, callback, *types):
+        raise NotImplementedError("Slice graphic collection to remove event handlers")
```
clewis7, line 228: so if I have a `LineCollection` of circles, do I need to have an image underlayed in order to get a click event? but then I have manually map the click event location to the closest circle...

kushalkolar: You need to click the actual line. Or You can add a click event handler to the renderer and then map that to world space, I should add a wrapper to PlotArea so it can directly add renderer events

clewis7:
adding linear selector linecollection is broken

fastplotlib/widgets/histogram_lut.py
clewis7, line 1: clean up file, move properties above `__init__()`

fastplotlib/graphics/_features/_base.py
```diff
@@ -112,21 +93,12 @@ def block_events(self, val: bool):
...
-        if isinstance(value, GraphicFeature):
-            return value()
-
-        return value
-
     def add_event_handler(self, handler: callable):
         """
         Add an event handler. All added event handlers are called when this feature changes.
+
         The ``handler`` can optionally accept a :class:`.FeatureEvent` as the first and only argument.
```
clewis7, line 100:
```suggestion
        The ``handler`` can optionally accept a :class:`.FeatureEvent` as the first and only argument.
        .. code-block:: python
        def event_handler(ev):
            print(ev.type)
            print(ev.pick_info)
        ...
```

fastplotlib/graphics/_features/_base.py
```diff
@@ -164,196 +136,192 @@ def clear_event_handlers(self):
...
-        # since the first obj is the datapoint indices
-        if isinstance(key[0], slice):
-            key = key[0]
+            with log_exception(
+                f"Error during handling {self.__class__.__name__} event"
+            ):
+                func(event_data)
+
+
+class BufferManager(GraphicFeature):
```
clewis7, line 150: should this be moved to its own file?

fastplotlib/graphics/_positions_base.py
```diff
@@ -0,0 +1,181 @@
...
+                        n_colors=self._data.value.shape[0],
+                        alpha=alpha,
+                    )
+                    self._cmap = VertexCmap(
+                        self._colors, cmap_name=None, cmap_values=None, alpha=alpha
+                    )
+
+        super().__init__(*args, **kwargs)
+
+    def detach_feature(self, feature: str):
```
clewis7, line 135: add doc string

fastplotlib/graphics/_positions_base.py
```diff
@@ -0,0 +1,181 @@
...
+            self._data._buffer = pygfx.Buffer(self._data.value.copy())
+            self.world_object.geometry.positions = self._data.buffer
+            self._data._shared -= 1
+
+        elif feature == "sizes":
+            self._sizes._buffer = pygfx.Buffer(self._sizes.value.copy())
+            self.world_object.geometry.positions = self._sizes.buffer
+            self._sizes._shared -= 1
+
+    def attach_feature(
```
clewis7, line 158: add doc string

clewis7 (changes requested):

fastplotlib/widgets/histogram_lut.py
kushalkolar: I think we should actually move the constructor to the top consistently, I've been bad about this XD

fastplotlib/graphics/_features/_text.py
```diff
@@ -0,0 +1,92 @@
+import numpy as np
+
+import pygfx
+
+from ._base import GraphicFeature, FeatureEvent
+
+
+class TextData(GraphicFeature):
```
clewis7, line 8: add features const like the other graphics have

kushalkolar:
discussing with @clewis7 , will make mixin class for line collection properties so that we can just do `line_collection.<features>` without slicing to get the entire collection.

kushalkolar:
@clewis7 Idea, make `Graphic._features` private and add a public `Graphic.events` list, and add the pygfx event names to it as well. People are familiar with properties so we won't need to make "graphic features" seem like this special thing to the users.

We can just say:
- graphics have properties, just like how other libs use properties for objects (pygfx objects use properties for everything)
- these properties are evented, you can have handlers that are called when they change

I think this makes fpl easier for newcommers to understand and use, the less jargon/new terms that don't exist in other libs the better? 

Internally for devs we can keep calling them features since we understand the concept.

kushalkolar:
I just realized, since we have the `_fpl_plot_area_hook` when a graphic is added to a plot area, for click events we can return the click x, y in world space with `PlotArea.map_screen_to_world()` and also return data x, y by just adding the offset to that.

clewis7:
> @clewis7 Idea, make `Graphic._features` private and add a public `Graphic.events` list, and add the pygfx event names to it as well. People are familiar with properties so we won't need to make "graphic features" seem like this special thing to the users.
> 
> We can just say:
> 
> * graphics have properties, just like how other libs use properties for objects (pygfx objects use properties for everything)
> * these properties are evented, you can have handlers that are called when they change
> 
> I think this makes fpl easier for newcommers to understand and use, the less jargon/new terms that don't exist in other libs the better?
> 
> Internally for devs we can keep calling them features since we understand the concept.

Sounds good to me, and much easier to explain

kushalkolar:
ok basically all test are done except for line collection, waiting on @clewis7 to finish `LineCollection` and graphic collection.

These are kinda corner cases so let's do them later:
- When sharing buffers, events are also shared. When deleting a graphic from which the event handler was originally added, we need to think about how to deal with the event handlers since other graphics share that buffer :thinking: 
- test attach/detach buffers

These are more relevant for selector tools and we haven't solidified that part of the API so we can move it to later:
- Update `Graphic._fpl_cleanup()`, must be possible to simplify this now that a `Graphic` instance keeps track of all the event handlers. Probably possible to have a simple `Graphic.clear_event_handlers()`
- Clear any pygfx rendering engine event handlers in `Graphic._fpl_cleanup()`

kushalkolar:
I still need to do:

- docstrings and comments and messages for the exceptions :laughing: 
- check the screenshot tests, possibly add/modify some

kushalkolar:
I should modify thickness to allow switching between the line thing and line materia, and add a dash pattern feature.

clewis7:
> ok basically all test are done except for line collection, waiting on @clewis7 to finish LineCollection and graphic collection.

@kushalkolar, I just need to write some tests for these?

fastplotlib/layouts/_plot_area.py
```diff
@@ -565,6 +566,11 @@ def _add_or_insert_graphic(
         # if we don't use the weakref above, then the object lingers if a plot hook is used!
         graphic._fpl_add_plot_area_hook(self)
 
+        # need to also hook in individual graphics in a collection to plot area
+        if isinstance(graphic, GraphicCollection):
+            for g in graphic.graphics:
+                g._fpl_add_plot_area_hook(self)
+
```
kushalkolar, line 573: This should be implemented in `GraphicCollection` base

fastplotlib/graphics/_collection_base.py
```diff
@@ -149,7 +149,7 @@ class GraphicCollection(Graphic):
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        cls._features = cls.child_type._features
+        cls.features = cls.child_type._features
```
kushalkolar, line 152: we made this private

fastplotlib/graphics/line_collection.py
```diff
@@ -431,31 +413,33 @@ def _get_linear_selector_init_args(self, padding, **kwargs):
             o = np.vstack(origin)
             origin_y = (o[:, 1].min() + o[:, 1].max()) / 2
             origin = (limits[0], origin_y)
+            center = (self.graphics[-1].world_object.world.y - self.graphics[0].world_object.world.y) / 2
```
kushalkolar, line 416: this has to be calculated from the data, not the offset

fastplotlib/graphics/line_collection.py
```diff
@@ -431,31 +413,33 @@ def _get_linear_selector_init_args(self, padding, **kwargs):
...
         else:
             o = np.vstack(origin)
             origin_x = (o[:, 0].min() + o[:, 0].max()) / 2
             origin = (origin_x, limits[0])
+            center = (self.graphics[-1].world_object.world.x - self.graphics[0].world_object.world.x) / 2
 
-        return bounds, limits, size, origin, axis, end_points
+        return bounds, limits, size, origin, axis, center
```
kushalkolar, line 423: baiscally this whole computation is probably wrong

fastplotlib/graphics/selectors/_linear.py
```diff
@@ -81,12 +81,12 @@ def __init__(
         axis: str, default "x"
             "x" | "y", the axis which the slider can move along
 
+        center: float
+            center of selector along the opposite axis of `axis`
```
kushalkolar, line 85:
```selection
            center position of selector on the axis orthogonal to the axis that the selector moves on.
```

kushalkolar:
This makes me think, we should prevent a graphic from being added to another plot area if it's already added to one.

kushalkolar:
@almarklein if you have time it would be great to get your thoughts, I've summarized what this does here since it's a massive diff:

The majority of this PR implements this: https://github.com/fastplotlib/fastplotlib/issues/389#issuecomment-2028534204 

I decided to go for explict properties instead of descriptors because it shows up better for typing and docstring popups.

### API comparision

#### Basic features (no buffer)

```python
# before this PR, setting and getting a feature value
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness()
3.5

# this PR, getter & setter are symmetric when possible
>>> line_graphic.thickness = 3.5
>>> line_graphic.thickness
3.5
```

#### Sliceable features (manages a buffer)

```python
# setting and getting the "entire value"
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

# this PR, similar but value can be directly access if desired
>>> line_graphic.colors = "r"
>>> line_graphic.colors
<__repr__ prints array>

>> line_graphic.colors.value
<returns the actual array>
```

**Sharing buffers** :smile: !

```python
# let's say you have a line with some vertex colors
>>> line1.colors
<returns array for red>

>>> line2 = subplot.add_line(data, colors=line1.colors)  # share buffer :D !

# modify colors
>>> line2.colors[50:] = "b"

# change of colors is reflected in BOTH graphics :D 
```

The implementation is very simple, for example in `PositionsGraphic.__init__`:

```python
class PositionsGraphic(Graphic):
  def __init__(self, data, colors, ...):
    if isinstance(data, VertexPositions):
        self._data = data
    else:
        self._data = VertexPositions(data, isolated_buffer=isolated_buffer)
```

As we can see, it actually shares the `BufferManager` (i.e. feature) instance. This is simpler and makes it so changing the buffer causes events to be emitted even if it's from another graphic. In the above example, if an event handler is added to `line1`, i.e. `line1.add_event_handler(<callable>, "colors")`, then changing the colors of `line2` would also trigger the event which I assume is what you would want since `line1` colors have _also_ changed.

### Events

```python
# before this PR
line_graphic.thickness.add_event_handler(<callable>)

# this PR
line_graphic.add_event_handler(<callable>, "thickness")

# can also use decorators :D

@line_graphic.add_event_handler("thickness")
def thickness_changed_handler(ev):
  pass

# exactly the same for features that subclass BufferManager
@line_graphic.add_event_handler("colors", "data")
def line_colors_or_data_changed(ev):
  if ev.type == "colors":
    print(
      f"colors changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )
  if ev.type == "data":
    print(
      f"data changed for graphic: <{ev.graphic}> 
      F"with slice: {ev.info['key']} and new values: {ev.info['value']}"
  )

# rendering engine events added in exactly the same way
@image_graphic.add_event_handler("click")
def image_clicked(ev):
  ev.graphic # the graphic that caused the event
```

The event adding and handling is managed here, mostly copy-pasted from pygfx.

https://github.com/fastplotlib/fastplotlib/blob/47cecfa300d0e712418ae4c72a2944f37eff2105/fastplotlib/graphics/_base.py#L209-L332

We handle the events in a wrapper method which appends the `Graphic` instead that caused the event, and also adds data and world (x, y) for pointer events.

Before this PR, the event info for "graphic features" was a mess that we probably don't need to dig into :laughing: . Now every `FeatureEvent` inherits from `pygfx.Event` and has the following attributes:

| attribute  | type        | description                                   |
|------------|-------------|-----------------------------------------------|
| type       | `str`         | name of the event, ex: "colors"                  |
| graphic    | `Graphic`     | graphic instance that the event is from       |
| info       | `dict`        | event info dictionary (see below)             |
| target     | `pygfx.WorldObject` | pygfx rendering engine object for the graphic |
| time_stamp | `float`       | time when the event occured, in ms            |

The "info" dict for all "simple" features has one key: "value" - the user passed new value. Example if `line_graphic.name` is changed, then the info dict will be: `{"value": "new_name_str"}`. The info dict will have an additional "key" entry if it's from a buffer manager feature.

#### `BufferManager` example info dicts:

**colors** (vertex colors, line or scatter)

| dict key | value type | value description |
|-------|---------------------------|-------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which colors were indexed/sliced |
| value | `np.ndarray` | new color values for points that were changed, shape is \[n_points_changed, RGBA\] |
| user_value | `str` \| `np.ndarray` \| `tuple[float]` \| `list[float]` \| `list[str]` | user input value that was parsed into the RGBA array |

**data** (points data, line or scatter)

| dict key | value type | value description |
|------|--------------------------|----------------------------------------|
| key | `int` \| `slice` \| `np.ndarray[int \| bool]` \| `tuple[slice, ...]` | key at which vertex positions data were indexed/sliced |
| value | `np.ndarray` \| `float` \| `list[float]` | new data values for points that were changed, shape depends on the indices that were set |

#### Selectors

Adding event handlers to a selector is identical to how they're added for "regular" graphics:

```python
# one of the selectors will change the line colors when it moves
@selector.add_event_handler("selection")
def set_color_at_index(ev):
    # changes the color at the index where the slider is
    ix = ev.get_selected_index()
    g = ev.graphic.parent
    g.colors[ix] = "green"
```

**Selector feature event structure**

`LinearSelector` `"selection"`

**additional event attributes:**

| attribute          | type     | description                        |
|--------------------|----------|------------------------------------|
| get_selected_index | callable | returns indices under the selector |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new x or y value of selection |

`LinearRegioonSelector`

**additional event attributes:**

| attribute            | type     | description                        |
|----------------------|----------|------------------------------------|
| get_selected_indices | callable | returns indices under the selector |
| get_selected_data    | callable | returns data under the selector    |

**info dict:**

| dict key | value type | value description             |
|----------|------------|-------------------------------|
| value    | np.ndarray | new `[min, max]` of selection |

`selector` is the `LinearSelector` (in this case) or `LinearRegionSelector` that the feature manages, and it has the `get_selected_index`, or `get_selected_indices` and `get_selected_data` methods for `LinearRegionSelector`

## Graphics

Some major refactor to graphics not mentioned above:

* line and scatter graphics now inherit from a common `PositionsGraphic` that handles things that are common between lines and scatters (data, colors, etc.)

clewis7 (changes requested):

clewis7:
Bug when adding an existing graphic to another plot

```
import fastplotlib as fpl
import numpy as np

fig = fpl.Figure()

fig[0,0].add_image(data=data)

fig2 = fpl.Figure()

fig2[0,0].add_graphic(fig[0,0].graphics[0])
```

yields
```
TypeError                                 Traceback (most recent call last)
Cell In[3], line 3
      1 fig2 = fpl.Figure()
----> 3 fig2[0,0].add_graphic(fig[0,0].graphics[0])

File [~/repos/fastplotlib/fastplotlib/layouts/_plot_area.py:468](http://localhost:8888/home/caitlin/repos/fastplotlib/fastplotlib/layouts/_plot_area.py#line=467), in PlotArea.add_graphic(self, graphic, center)
    465     self.scene.add(graphic.world_object)
    466     return
--> 468 self._add_or_insert_graphic(graphic=graphic, center=center, action="add")
    470 if self.camera.fov == 0:
    471     # for orthographic positions stack objects along the z-axis
    472     # for perspective projections we assume the user wants full 3D control
    473     graphic.offset = (*graphic.offset[:-1], len(self))

File [~/repos/fastplotlib/fastplotlib/layouts/_plot_area.py:558](http://localhost:8888/home/caitlin/repos/fastplotlib/fastplotlib/layouts/_plot_area.py#line=557), in PlotArea._add_or_insert_graphic(self, graphic, center, action, index)
    555 REFERENCES.add(graphic)
    557 # now that it's in the dict, just use the weakref
--> 558 graphic = weakref.proxy(graphic)
    560 # add world object to scene
    561 self.scene.add(graphic.world_object)

TypeError: cannot create weak reference to 'weakref.ProxyType' object
```

also, even just trying to access a graphic
```
fig[0,0].graphics[0]
```
```
----> 1 fig[0,0].graphics[0]

File [~/repos/fastplotlib/fastplotlib/layouts/_plot_area.py:280](http://localhost:8888/home/caitlin/repos/fastplotlib/fastplotlib/layouts/_plot_area.py#line=279), in PlotArea.graphics(self)
    277 @property
    278 def graphics(self) -> tuple[Graphic, ...]:
    279     """Graphics in the plot area. Always returns a proxy to the Graphic instances."""
--> 280     return REFERENCES.get_proxies(self._graphics)

File [~/repos/fastplotlib/fastplotlib/layouts/_plot_area.py:62](http://localhost:8888/home/caitlin/repos/fastplotlib/fastplotlib/layouts/_plot_area.py#line=61), in References.get_proxies(self, refs)
     60 for key in refs:
     61     if key in self._graphics.keys():
---> 62         proxies.append(weakref.proxy(self._graphics[key]))
     64     elif key in self._selectors.keys():
     65         proxies.append(weakref.proxy(self._selectors[key]))

TypeError: cannot create weak reference to 'weakref.ProxyType' object
```

fastplotlib/graphics/_base.py
```diff
@@ -69,123 +79,266 @@ def __init__(
...
+        # all the common features
+        self._name = Name(name)
+        self._deleted = Deleted(False)
+        self._rotation = Rotation(rotation)
+        self._offset = Offset(offset)
+        self._visible = Visible(visible)
+        self._block_events = False
+
+    @property
+    def events(self) -> tuple[str]:
```
almarklein, line 118: Maybe `supported_events`?

fastplotlib/graphics/_base.py
```diff
@@ -69,123 +79,266 @@ def __init__(
...
 
-    @property
-    def position_y(self) -> float:
-        """y-axis position of the graphic"""
-        return self.world_object.world.y
+        # set rotation if it's not (0., 0., 0., 1.)
+        if not all(self.world_object.world.rotation == self.rotation):
+            self.rotation = self.rotation
+
+    def detach_feature(self, feature: str):
```
almarklein, line 195: Some methods (this one, but also e.g. `clear_event_handlers()` are missing a docstring.

almarklein (commented):
> @almarklein if you have time it would be great to get your thoughts, I've summarized what this does here since it's a massive diff

The description you posted sounds very good!

> Sharing buffers

Hooray! Feels nice and simple.

> Adding event handlers to a selector is identical to how they're added for "regular" graphics:

Nice, less stuff to learn for users.


I also glanced over the diff. Looks nice from a birds-eye view. Made a few minor comments/suggestions.

kushalkolar:
failing because camera stuff was recently merged in `pygfx`, I might as well update this PR with it :melting_face:

almarklein:
> failing because camera stuff was recently merged in pygfx

You mean https://github.com/pygfx/pygfx/pull/778? I was under the impression that that change was backwards compatible.

kushalkolar:
It's specific to how we allow changing the controller of an existing plot area: https://github.com/fastplotlib/fastplotlib/actions/runs/9492712816/job/26160279159?pr=511#step:9:895

clewis7 (approved):
YAYAY!!!

kushalkolar:
Everything passing on our ends locally, merging. Need to wait for pygfx to fix https://github.com/pygfx/pygfx/issues/790 and for things to catch up with numpy v2 before all our CI works again.

PR #529, opened 2024-06-18
disable cmap property for RGB images

clewis7:
set `cmap` feature for RGB images to `None` and provide more useful feedback is user try to set or read property


clewis7:
image

kushalkolar (approved):

PR #546, opened 2024-07-01
Use real references to `Graphics`, proper garbage collection in ipython and jupyter

### LINKED ISSUE #161: Create a Proxy subclass that we use for our Graphic classes
the purpose of this is having useful `__repr__` etc.

### LINKED ISSUE #393: GraphicCollection.__del__ method can cause confusing tracebacks
If something fails during init of a GraphicCollection, an exception will be raised, but that exception will often be hidden higher up in the stack due to a failure during cleanup in the `__del__` method.  for example, if you pass the wrong length of cmaps to a LineCollection:

```pytb
  File "/Users/talley/dev/forks/fastplotlib/examples/desktop/line_collection/line_stack.py", line 27, in <module>
    plot.add_line_stack(data, cmap=["jet"])
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/layouts/graphic_methods_mixin.py", line 314, in add_line_stack
    return self._create_graphic(LineStack, data, z_position, thickness, colors, cmap, separation, separation_axis, name, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/layouts/graphic_methods_mixin.py", line 25, in _create_graphic
    graphic = graphic_class(*args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/line_collection.py", line 547, in __init__
    super(LineStack, self).__init__(
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/line_collection.py", line 125, in __init__
    raise ValueError(
ValueError: cmap argument must be a single cmap or a list of cmaps with the same length as the data
Exception ignored in: <function GraphicCollection.__del__ at 0x1212071a0>
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/_base.py", line 477, in __del__
    super().__del__()
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/_base.py", line 171, in __del__
    del WORLD_OBJECTS[self.loc]
        ~~~~~~~~~~~~~^^^^^^^^^^
KeyError: '0x16a217250'
```

It would be nice to either ignore errors in the `__del__` method, or do a bit more careful cleanup depending on the state of the object:

```python
    def __del__(self):
        with contextlib.suppress(Exception):
            self.world_object.clear()
            for loc in self._graphics:
                del COLLECTION_GRAPHICS[loc]

            super().__del__()
```

and then you'd see:

```pytb
  File "/Users/talley/dev/forks/fastplotlib/examples/desktop/line_collection/line_stack.py", line 27, in <module>
    plot.add_line_stack(data, cmap=["jet"])
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/layouts/graphic_methods_mixin.py", line 314, in add_line_stack
    return self._create_graphic(LineStack, data, z_position, thickness, colors, cmap, separation, separation_axis, name, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/layouts/graphic_methods_mixin.py", line 25, in _create_graphic
    graphic = graphic_class(*args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/line_collection.py", line 547, in __init__
    super(LineStack, self).__init__(
  File "/Users/talley/dev/forks/fastplotlib/fastplotlib/graphics/line_collection.py", line 125, in __init__
    raise ValueError(
ValueError: cmap argument must be a single cmap or a list of cmaps with the same length as the data
```

### LINKED ISSUE #472: use of weakref.proxy
I figured it's useful to have this in an issue and publicly as well.

We return weakref.proxy objects to the real `Graphic` objects so that jupyter doesn't accumulate references when the user uses references to the Graphic.

Almar's description:

> The problem is that it may confuse the user (it shows “proxy” in the repr) and it results in a rather unconventional memory management system.
> 
> I’ve looked into at least changing the repr, but this is not possible. We’ve discussed the reasons underlying the choice, coming mostly down to event handlers holding onto objects. This is a tricky problem, and there is not a simply solution that will make it go away.
> 
> We’ve concluded that the current solution does the job. Since these proxy objects look a lot like the real objects, the burden on the user is not too bad. And if the approach is changed in the future, this will most likely not affect users code.
> 
> On the longer term, I think it might be worth looking into a way to clear handlers. This should also be done in pygfx’ world objects. Then when a plot is cleared, it can also clear/destroy all graphics objects within.

I just wanted to add, I think the one issue with proxies is that they are not hashable, so they cannot be used as dict keys.

Anyways, not something we will work on now, but #467 helps make this all cleaner internally.

kushalkolar:
closes #472, closes #161

also closes #393 

No more weakreferences to `Graphics`! :partying_face: 

Figured out how to properly garbage collect references proliferated by ipython in jupyter with some help from @lauraporta, thanks! :smile: 

The `pygfx.WorldObject` reference in `Graphic` is still a `weakref.proxy`, I thought let's keep it like that for now since we don't expect users to directly interact with the rendering engine objects much.

In `PlotArea.delete_graphic()` this is how proliferated references can be removed:

https://github.com/fastplotlib/fastplotlib/blob/675dfaa53f77ffc93626ebd2dddfc8d0f0ea38d4/fastplotlib/layouts/_plot_area.py#L642-L651


fastplotlib/layouts/_plot_area.py
```diff
@@ -696,29 +621,34 @@ def delete_graphic(self, graphic: Graphic):
...
+        if IS_IPYTHON:
+            # remove any references that ipython might have made
+            ip = get_ipython()
+            # get all references in ipython namespace
+            references = list(ip.user_ns.keys())
+            for ref in references:
+                if ip.user_ns[ref] is graphic:
+                    # we found the graphic, remove from ipython
+                    ip.del_var(ref)
+                    break
```
almarklein, line 651: How likely is it that the object is in `user_ns`? If it's only occasionally, can wrap this in:
```py
if graphic in ip.user_ns.values():
    ...
```

I guess it's possible that the graphic is present more than once, under a different name?

Alt solution that might be faster because it avoids copying the keys and looking up the values:
```py
            keys_to_delete = []
            for key, ob in ip.user_ns.items():
                if ob is graphic:
                    keys_to_delete.append(key)
            for key in keys_to_delete:
                    ip.del_var(key)
```

almarklein (commented):
> No more weakreferences to Graphics! 🥳

🥳 ❤️

clewis7 (dismissed):
LGTM aside from the minor suggestions that Almar had 

I think this will be super useful for users to get the actual object instead of a weakref back (less confusing when interacting with objects etc.). I agree that for now we just leave weakref to the world object...we can always change later.

fastplotlib/graphics/_collection_base.py
```diff
@@ -324,20 +310,22 @@ def _fpl_cleanup(self):
...
-            return weakref.proxy(COLLECTION_GRAPHICS[addr])
+            return self.graphics[key]
 
         return self._indexer(selection=self.graphics[key], features=self._features)
 
     def __del__(self):
-        self.world_object.clear()
+        # remove world object if it was created
+        with suppress(KeyError):
+            self.world_object.clear()
```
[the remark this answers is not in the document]
kushalkolar: you're right :laughing: , should just be a `clear()`

fastplotlib/layouts/_plot_area.py
kushalkolar: > How likely is it that the object is in `user_ns`? If it's only occasionally, can wrap this in:
> 
> ```python
> if graphic in ip.user_ns.values():
>     ...
> ```

`in` operators don't work if any item of the iterable is a numpy array, for example:

```python
a = np.random.rand(10)
l = [a, "bah", 123]

# will raise a ValueError
123 in l
```

I've resorted to this, couldn't think of a more performance solution:

```python
        if IS_IPYTHON:
            # remove any references that ipython might have made
            ip = get_ipython()

            # check both namespaces
            for namespace in [ip.user_ns, ip.user_ns_hidden]:
                # find the reference
                for ref, obj in namespace.items():
                    if graphic is obj:
                        # we found the reference, remove from ipython
                        ip.del_var(ref)
                        break
```

kushalkolar:
ready for another look, sorry for the delay been sick :sneezing_face:

fastplotlib/layouts/_plot_area.py
almarklein: No problems with deleting items from an object being iterated over? I guess it depends on IPython internals whether that'd be a problem.

fastplotlib/graphics/_base.py
```diff
@@ -389,9 +389,11 @@ def _fpl_prepare_del(self):
...
 
     def __del__(self):
-        with suppress(KeyError):
-            # remove world object if created
-            del WORLD_OBJECTS[hex(id(self))]
+        wo = WORLD_OBJECTS.pop(hex(id(self)), None)
+        # remove world object if created
+        # world object does not exist if an exception was raised during __init__ which is why this check exists
+        if wo is not None:
+            del wo
```
almarklein, line 396: Why not simply call the pop without storing the result?

almarklein:
Hope you get better soon!

fastplotlib/graphics/_collection_base.py
```diff
@@ -1,12 +1,9 @@
+from contextlib import suppress
```
clewis7, line 1: Is this actually being used anywhere?

fastplotlib/graphics/_base.py
kushalkolar: ah right that will still gc since it's no longer in locals right, an explicit `del` is not required?

almarklein: Exactly. And even if you would store the result in `wo`, it would still be gc'd once this method is done. But since you don't use `wo` it makes more sense to just not store it.

kushalkolar: thanks!

fastplotlib/graphics/_collection_base.py
kushalkolar: yup you're right

kushalkolar:
@clewis7 ready to go!

clewis7 (approved):

PR #549, opened 2024-07-08
add marker, edge_color, edge_with to `ScatterGraphic`

kushalkolar:
WIP, still need to do tests and think about how exactly to implement markers

@almarklein do you think it could be possible and make sense to allow markers to be set per-vertex for `pygfx.Points`? It would make it easier to create visualization with a mix of both, and even easier to change markers per-vertex.

For example, with how I've currently implemented it a `ScatterGraphic` is required per-marker type for a visualization:

```python
import fastplotlib as fpl
import pygfx
import numpy as np
from sklearn.cluster import AgglomerativeClustering
from sklearn import datasets


iris = datasets.load_iris()
data = iris["data"]
targets = iris["target"]

agg = AgglomerativeClustering(n_clusters=3)
agg.fit_predict(data)

figure = fpl.Figure()

# markers to denote the real label
markers = ["circle", "cross", "square"]


# colors to denote the clustering results
tab10 = fpl.utils.make_colors(3, "tab10")
label_colors = {i: pygfx.Color(tab10[i]).hex for i in range(3)}
# makes a list of hex strings by mapping the labels using the above dict
colors = np.fromiter(map(label_colors.get, agg.labels_), dtype="<U7")

for marker, target in zip(markers, np.unique(targets)):
    figure[0, 0].add_scatter(
        data=data[targets == target, :-1].astype(np.float32),
        sizes=15,
        alpha=0.5,
        colors=colors[targets == target],
        marker=marker,
    )

figure.show()
```

image

If not no worries, I think a workaround is to make `ScatterGraphic` manage a `pygfx.Group`, and allow `markers` to be passed as an iterable of `str` markers, and then make the materials based on the passed markers. The tricky part is changing the materials any time a user changes the markers. :thinking:

### DIFF
--- a/docs/source/api/graphics/ScatterGraphic.rst
+++ b/docs/source/api/graphics/ScatterGraphic.rst
@@ -25,7 +25,10 @@ Properties
     ScatterGraphic.colors
     ScatterGraphic.data
     ScatterGraphic.deleted
+    ScatterGraphic.edge_color
+    ScatterGraphic.edge_width
     ScatterGraphic.event_handlers
+    ScatterGraphic.marker
     ScatterGraphic.name
     ScatterGraphic.offset
     ScatterGraphic.rotation
--- a/examples/desktop/scatter/scatter_marker.py
+++ b/examples/desktop/scatter/scatter_marker.py
@@ -1,0 +1,56 @@
+"""
+Scatter Plot Markers
+====================
+
+Example showing scatter plot using different markers
+"""
+
+# test_example = true
+# sphinx_gallery_pygfx_docs = 'screenshot'
+
+import fastplotlib as fpl
+import pygfx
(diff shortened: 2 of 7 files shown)


almarklein:
Good point, I see how this would be very useful. I put my thoughts is a comment on the points issue in pygfx: https://github.com/pygfx/pygfx/issues/48#issuecomment-2213447462

kushalkolar:
Note to self: I think I'm actually going to finish this soon as-is and refactor after it's possible to do per-point markers like an array

kushalkolar:
Closing, per-vertex shapes has been implemented in pygfx so we can do a much better version of this :smile: https://github.com/pygfx/pygfx/pull/1208

PR #550, opened 2024-07-10
add colorbar to hlut tool, add `pause_events()` context manager

kushalkolar:
adds a dynamic colobar to the HistogramLUT tool

https://github.com/fastplotlib/fastplotlib/assets/9403332/ecef47df-7f72-4439-94fe-177662f018ba

@clewis7 this will modify the test screenshot, but I will do that after getting your thoughts on this :smile:

### DIFF
--- a/fastplotlib/graphics/utils.py
+++ b/fastplotlib/graphics/utils.py
@@ -13,11 +13,9 @@ def pause_events(*graphics: Graphic):
 
     original_vals = [g.block_events for g in graphics]
 
-    try:
-        for g in graphics:
-            g.block_events = True
-        yield
+    for g in graphics:
+        g.block_events = True
+    yield
 
-    finally:
-        for g, value in zip(graphics, original_vals):
-            g.block_events = value
+    for g, value in zip(graphics, original_vals):
+        g.block_events = value
--- a/fastplotlib/graphics/utils.py
+++ b/fastplotlib/graphics/utils.py
@@ -1,0 +1,23 @@
+from contextlib import contextmanager
+
+from ._base import Graphic
+
+
+@contextmanager
+def pause_events(*graphics: Graphic):
+    if not all([isinstance(g, Graphic) for g in graphics]):
+        raise TypeError(
+            f"`pause_events` only takes Graphic instances as arguments, "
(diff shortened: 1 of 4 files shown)


fastplotlib/widgets/histogram_lut.py
```diff
@@ -205,27 +239,39 @@ def _linear_region_handler(self, ev):
...
-        self.image_graphic.block_events = False
-        self._linear_region_selector.block_events = False
+        with pause_events(self.image_graphic, self._linear_region_selector):
+            # must use world coordinate values directly from selection()
+            # otherwise the linear region bounds jump to the closest bin edges
+            self._linear_region_selector.selection = (
+                value * self._scale_factor,
+                self._linear_region_selector.selection[1],
+            )
+            self.image_graphic.vmin = value
```
kushalkolar, line 270: @clewis7 you might also find this useful, just pass any number of graphics to `pause_events()` and the events are blocked for those graphics while inside the `with` block :smile:

fastplotlib/graphics/utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ._base import Graphic
+
+
+@contextmanager
+def pause_events(*graphics: Graphic):
+    if not all([isinstance(g, Graphic) for g in graphics]):
+        raise TypeError(
+            f"`pause_events` only takes Graphic instances as arguments, "
+            f"you have passed the following types:\n{[type(g) for g in graphics]}"
```
clewis7, line 11: Should we make this a set so we are only returning the different unique types passed? 

I can just think of maybe wanting to pass a lot of graphics to be paused at one time and then finding it hard to figure out which type might be erroneous

clewis7 (commented):
LGTM aside from maybe adding some code comments

I think the `pause_events()` will be super useful for pynaviz stuff

fastplotlib/graphics/utils.py
kushalkolar: If it's like this then the list of pass types is printed in the same order as the user's argument so they can figure out which item is wrong. Whereas if they passed a set you don't know which index has the non-Graphic item.

kushalkolar:
closing because I merged this into axes #551

PR #551, opened 2024-07-11
Basic axes

kushalkolar:
Started prototyping axes. A lot of this is based on the pygfx ruler example. @almarklein would be great to get your input on this, this is just a prototype so I'm more than happy if you had other ideas on how to do this! I was too excited to protype this :smile: 

Summary:

### Grid class
subclass `pygfx.Grid` to give it properties from the `GridMaterial` so the user can for example do `subplot.axes.grid["xy"].major_step` instead of ``subplot.axes.grid["xy"].material.major_step`

### Axes class

The main properties are:
  - `x`, `y`, `z` - `pygfx.Ruler` instance for each dimension
  - `grid` - dict containing the `Grid` instance for each plane, "xy", "xz", "yz" 
  - `follow` - whether or not the axes follow the camera in the bottom-left corner during pan

Has one method: `animate`, which based on the pygfx ruler example and sets the ruler start and end positions, and updates them. Main differences from the current pygfx example are:
  - uses inverse projection to find the range of the world space, accounts for camera zoom etc. 
  - the `pygfx.Ruler` start and end points are swapped if the `camera.local.scale_{dim}` is negative. 

`Subplots` auto-make an `Axes` instance. The `animate` function to update the axes are added to the subplot's render functions list.

I also had to slightly modify `PlotArea`. There's a new `pygfx.Group` `PlotArea._graphics_scene` which sequesters all the `WorldObjects` that belong to `Graphics` in the `PlotArea`. This allows us to compute the world bounding box on the actual plot-data-graphics and exclude the selectors. 

`PlotArea._graphics_scene` is added to `PlotArea.scene`, and non-`Graphic` `WorldObjects`, such as legends, the axes, and selector tools, are added directly to `PlotArea.scene`.

Todo:
- [ ] Not working with `camera.fov > 0`, need to figure out why.
- [ ] z axis ruler
- [ ] figure out how to set `PlotArea.auto_scale()` such that the rulers are in the bottom left corner, basically take into account the ruler spacing from the subplot edge so that all the `Graphics` are "contained" within the rulers when `auto_scale()` is used, right now it's wonky with the current auto_scale. I think just reducing the zoom should be sufficient?
- [ ] modify all test screenshots again :upside_down_face: 

https://github.com/fastplotlib/fastplotlib/assets/9403332/dc4ad68e-7aeb-4a9c-9660-24e96e051ede

https://github.com/fastplotlib/fastplotlib/assets/9403332/a1cbb81f-f1c0-43a5-b461-56818c0d8ed9

### DIFF
--- a/fastplotlib/layouts/_plot_area.py
+++ b/fastplotlib/layouts/_plot_area.py
@@ -99,6 +99,11 @@ class PlotArea:
         # legends, managed just like other graphics as explained above
         self._legends: list[Legend] = list()
 
+        # keep all graphics in a separate group, makes bbox calculations etc. easier
+        # this is the "real scene" excluding axes, selection tools etc.
+        self._graphics_scene = pygfx.Group()
+        self.scene.add(self._graphics_scene)
+
         self._name = name
 
         # need to think about how to deal with children better
@@ -394,7 +399,7 @@ class PlotArea:
 
         if graphic in self:
             # graphic is already in this plot but was removed from the scene, add it back
-            self.scene.add(graphic.world_object)
+            self._graphics_scene.add(graphic.world_object)
             return
 
         self._add_or_insert_graphic(graphic=graphic, center=center, action="add")
@@ -465,12 +470,15 @@ class PlotArea:
 
         if isinstance(graphic, BaseSelector):
             obj_list = self._selectors
+            self.scene.add(graphic.world_object)
 
         elif isinstance(graphic, Legend):
             obj_list = self._legends
+            self.scene.add(graphic.world_object)
 
         elif isinstance(graphic, Graphic):
             obj_list = self._graphics
+            self._graphics_scene.add(graphic.world_object)
 
         else:
             raise TypeError("graphic must be of type Graphic | BaseSelector | Legend")
@@ -481,9 +489,6 @@ class PlotArea:
             obj_list.append(graphic)
         else:
             raise ValueError("valid actions are 'insert' | 'add'")
-
-        # add world object to scene
-        self.scene.add(graphic.world_object)
 
         if center:
             self.center_graphic(graphic)
@@ -518,7 +523,7 @@ class PlotArea:
         # probably because camera.show_object uses bounding sphere
         self.camera.zoom = zoom
 
-    def center_scene(self, *, zoom: float = 1.35):
+    def center_scene(self, *, zoom: float = 1.0):
         """
         Auto-center the scene, does not scale.
 
@@ -528,13 +533,13 @@ class PlotArea:
             apply a zoom after centering the scene
 
         """
-        if not len(self.scene.children) > 0:
+        if not len(self._graphics_scene.children) > 0:
             return
 
         # scale all cameras associated with this controller
         # else it looks wonky
         for camera in self.controller.cameras:
-            camera.show_object(self.scene)
+            camera.show_object(self._graphics_scene)
 
             # camera.show_object can cause the camera width and height to increase so apply a zoom to compensate
             # probably because camera.show_object uses bounding sphere
@@ -544,7 +549,7 @@ class PlotArea:
         self,
         *,  # since this is often used as an event handler, don't want to coerce maintain_aspect = True
         maintain_aspect: None | bool = None,
-        zoom: float = 0.8,
+        zoom: float = 1.0,
     ):
         """
         Auto-scale the camera w.r.t to the scene
@@ -559,12 +564,8 @@ class PlotArea:
             zoom value for the camera after auto-scaling, if zoom = 1.0 then the graphics
             in the scene will fill the entire canvas.
         """
-        if not len(self.scene.children) > 0:
+        if not len(self._graphics_scene.children) > 0:
             return
-        # hacky workaround for now until we decide if we want to put selectors in their own scene
-        # remove all selectors from a scene to calculate scene bbox
-        for selector in self.selectors:
-            self.scene.remove(selector.world_object)
 
         self.center_scene()
 
@@ -575,8 +576,8 @@ class PlotArea:
         for camera in self.controller.cameras:
             camera.maintain_aspect = maintain_aspect
 
-        if len(self.scene.children) > 0:
-            width, height, depth = np.ptp(self.scene.get_world_bounding_box(), axis=0)
+        if len(self._graphics_scene.children) > 0:
+            width, height, depth = np.ptp(self._graphics_scene.get_world_bounding_box(), axis=0)
         else:
             width, height, depth = (1, 1, 1)
 
@@ -585,9 +586,6 @@ class PlotArea:
             width = 1
         if height < 0.01:
             height = 1
-
-        for selector in self.selectors:
-            self.scene.add(selector.world_object)
 
         # scale all cameras associated with this controller else it looks wonky
         for camera in self.controller.cameras:
@@ -609,7 +607,11 @@ class PlotArea:
 
         """
 
-        self.scene.remove(graphic.world_object)
+        if isinstance(graphic, (BaseSelector, Legend)):
+            self.scene.remove(graphic.world_object)
+
+        elif isinstance(graphic, Graphic):
+            self._graphics_scene.remove(graphic.world_object)
 
     def delete_graphic(self, graphic: Graphic):
         """
@@ -626,15 +628,19 @@ class PlotArea:
 
         if isinstance(graphic, BaseSelector):
             self._selectors.remove(graphic)
+
         elif isinstance(graphic, Legend):
             self._legends.remove(graphic)
 
         elif isinstance(graphic, Graphic):
-            self._graphics.remove(graphic)
+            self._graphics_scene.remove(graphic)
 
         # remove from scene if necessary
         if graphic.world_object in self.scene.children:
             self.scene.remove(graphic.world_object)
+
+        elif graphic.world_object in self._graphics_scene.children:
+            self._graphics_scene.remove(graphic.world_object)
 
         # cleanup
         graphic._fpl_prepare_del()
--- a/fastplotlib/layouts/_axes.py
+++ b/fastplotlib/layouts/_axes.py
@@ -1,0 +1,296 @@
+import pygfx
+
+
+# very thin subclass that just adds GridMaterial properties to this world object for easier user control
+class Grid(pygfx.Grid):
+    @property
+    def major_step(self):
+        """The step distance between the major grid lines."""
+        return self.material.major_step
+
+    @major_step.setter
+    def major_step(self, step):
+        self.material.major_step = step
+
+    @property
+    def minor_step(self):
+        """The step distance between the minor grid lines."""
+        return self.material.minor_step
+
+    @minor_step.setter
+    def minor_step(self, step):
+        self.material.minor_step = step
+
+    @property
+    def axis_thickness(self):
+        """The thickness of the axis lines."""
+        return self.material.axis_thickness
+
+    @axis_thickness.setter
+    def axis_thickness(self, thickness):
+        self.material.axis_thickness = thickness
+
+    @property
+    def major_thickness(self):
+        """The thickness of the major grid lines."""
+        return self.material.major_thickness
+
+    @major_thickness.setter
+    def major_thickness(self, thickness):
+        self.material.major_thickness = thickness
+
+    @property
+    def minor_thickness(self):
+        """The thickness of the minor grid lines."""
+        return self.material.minor_thickness
+
+    @minor_thickness.setter
+    def minor_thickness(self, thickness):
+        self.material.minor_thickness = thickness
+
+    @property
+    def thickness_space(self):
+        """The coordinate space in which the thicknesses are expressed.
+
+        See :obj:`pygfx.utils.enums.CoordSpace`:
+        """
+        return self.material.thickness_space
+
+    @thickness_space.setter
+    def thickness_space(self, value):
+        self.material.thickness_space = value
+
+    @property
+    def axis_color(self):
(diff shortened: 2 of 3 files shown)


kushalkolar:
Idea: Create Axes on a graphic, useful when the graphic has an offset. Example use cases: line stack, many images in one subplot. For LineStack especially it can be useful to have independent y-rulers for each, and just one x-ruler. 

Can be Graphic.add_axes()

Maybe we should make Axes a Graphic so then it's easy to link the offset and rotation. 

The start and end position of the rulers for this should be the box of the graphic.

almarklein:
Very nice to see this taking shape! Looks good so far to me :)

kushalkolar:
Let's try to merge a prototype of this towards the end of this week/early next week and we can evolve it over time with use. It is messy but we can refactor :D

Updates:

Can add axes to a single graphic, useful for line stacks or any graphics that are offset from the origin, example:

image

I did not make `Axes` a graphic, it doesn't require all that functionality and we avoid circular imports. I think let's keep it this way.

For perspective projections, it was non-trivial to map the screen space to world space in all 3 dimensions, so instead what it will do for perspective projections is just get the scene bbox and set the axes limits based on the scene bbox :smile: . I think this is a good enough compromise for now. 

image

fastplotlib/layouts/_figure.py
```diff
@@ -474,7 +474,7 @@ def show(
                     _maintain_aspect = subplot.camera.maintain_aspect
                 else:
                     _maintain_aspect = maintain_aspect
-                subplot.auto_scale(maintain_aspect=_maintain_aspect, zoom=0.95)
+                subplot.auto_scale(maintain_aspect=maintain_aspect)
```
kushalkolar, line 477: I removed the zoom so that the scaling is nice w.r.t. axes positions when you do `fig.show()`

fastplotlib/graphics/_axes.py
```diff
@@ -0,0 +1,423 @@
...
+            world_ymin, world_ymax = world_ymax, world_ymin
+            self.y.tick_side = "right"  # swap tick side
+            self.x.tick_side = "left"
+        else:
+            self.y.tick_side = "left"
+            self.x.tick_side = "right"
+
+        if self._plot_area.camera.local.scale_x < 0:
+            world_xmin, world_xmax = world_xmax, world_xmin
+            self.x.tick_side = "left"
```
kushalkolar, line 390: @almarklein is there a nicer way to do this? :sweat_smile:

kushalkolar:
I just realized, if axes are added to a graphic directly, it's not necessary to update them on every render cycle. They can just be fixed. Change only if the graphic offset and rotation change.

kushalkolar:
@clewis7 Once you've gotten further in `pynaviz` the Axes tick label mapping functions will be really important for you. For example, let's say you have a `HeatmapVisual` to display subsampled data of a massive recording. You could have some property like `HeatmapVisual.graphic_data_subsample = (1, 1_000)`: subsample by (1, 1_000) in [rows, columns], and columns is typically time which is usually the subsampled dimension. When you subscribe a `HeatmapVisual` causes the time to change, the `TimeStore` sees the graphic data subsample factor and adjusts accordingly to get the real time. This real time is then used for any other mappings.

fastplotlib/graphics/_axes.py
almarklein: Can probable be made shorter, but not sure if it'd be prettier :)

Does this also do the right thing if `scale_x` and `scale_y` are negative?

clewis7:
This looks amazing!! So cool :D

fastplotlib/graphics/_axes.py
clewis7, line 1: I think because you ended up not making axes a graphic, that maybe we should put it somewhere else??

fastplotlib/graphics/_axes.py
```diff
@@ -0,0 +1,423 @@
...
+# very thin subclass that just adds GridMaterial properties to this world object for easier user control
+class Grid(pygfx.Grid):
+    @property
+    def major_step(self):
+        """The step distance between the major grid lines."""
+        return self.material.major_step
+
+    @major_step.setter
+    def major_step(self, step):
+        self.material.major_step = step
```
clewis7, line 18: Add type annotation

fastplotlib/graphics/_axes.py
```diff
@@ -0,0 +1,423 @@
...
+
+    @property
+    def minor_step(self):
+        """The step distance between the minor grid lines."""
+        return self.material.minor_step
+
+    @minor_step.setter
+    def minor_step(self, step):
```
clewis7, line 26: same as above

fastplotlib/graphics/_axes.py
clewis7, line 1: all the setters in this file need type annotations

fastplotlib/graphics/_axes.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # world object for the rulers + grids
+        self._world_object = pygfx.Group()
+
+        # add rulers
+        self.world_object.add(
+            self.x,
+            self.y,
+            self._z,
+        )
```
clewis7, line 211: looks like typo here

```self.z```

File: fastplotlib/graphics/_base.py [hunk 6]
```diff
@@ -413,3 +422,21 @@ def rotate(self, alpha: float, axis: Literal["x", "y", "z"] = "y"):
                 f"`axis` must be either `x`, `y`, or `z`. `{axis}` provided instead!"
             )
         self.rotation = la.quat_mul(rot, self.rotation)
+
+    @property
+    def axes(self) -> Axes:
+        return self._axes
+
+    def add_axes(self):
```
clewis7, line 430: for images, adding axes places them on the top of the image with the numbers above the image but for the y-axis the numbers are overlayed on top of the image...I think the number should also be to the outside of the image here

image

clewis7: if I flip the image, the axes are correct...is there any easy way to make sure no matter the orientation of the image we are also displaying axes for images on the bottom of the image and on the left??

fastplotlib/graphics/utils.py
```diff
@@ -0,0 +1,21 @@
...
+        )
+
+    original_vals = [g.block_events for g in graphics]
+
+    for g in graphics:
+        g.block_events = True
+    yield
+
+    for g, value in zip(graphics, original_vals):
+        g.block_events = value
```
clewis7, line 21: :))))

fastplotlib/layouts/_figure.py
clewis7: I think we should add a `bool` in `show()` that allows axes visibility to be `False` to start 

```python
fig.show(axes=False)
```

fastplotlib/graphics/_base.py
kushalkolar: A user can call `graphic.axes.update_using_camera()` after flipping. We can't do this automatically without making it an animation function or having callbacks from cameras. Let's wait for https://github.com/pygfx/pygfx/issues/495 before persuing this further.

kushalkolar:
I subclassed `pygfx.Ruelr` and two attributes:
  - `font_size`, default `14`
  - `tick_text_mapper`, default `None`

Let's kinda keep both of these hidden for now, play with 

Example of how `tick_text_mapper` can be used:

```python
import fastplotlib as fpl
import numpy as np


a = np.random.rand(1_000)

fig = fpl.Figure()

line = fig[0, 0].add_line(a)


def basic_mapper(text: str):
    """map values by dividing by 30"""
    value = float(text)
    # divide by 30, show in seconds
    return f"{value / 30:.2f}s"


indexer = np.linspace(0, 20, 1_000)
def array_mapper(text: str):
    """map values by indexing an array, ex. time axis from pynapple arrays"""
    value = int(text)
    if value >= 0 and value < indexer.size:
        return f"{indexer[value]:.2f}s"

    return ""  # no match, return nothing


fig[0, 0].axes.x.tick_text_mapper = basic_mapper

fpl.run()
```

Basically the `tick_text_mapper` func can be anything that takes a `tick` `str` label and returns a str label.

using `basic_mapper`:

image


using `array_mapper`:

image

examples/desktop/image/image_widget.py
clewis7, line 1: did you get this to pass? I never could

fastplotlib/layouts/_plot_area.py
```diff
@@ -240,6 +254,21 @@ def name(self, name: str):
...
+            self._background_material.color_top_right,
+            self._background_material.color_bottom_right,
+            self._background_material.color_bottom_left,
+        )
+
+    @background_color.setter
+    def background_color(self, colors: str | tuple[float]):
+        """1, 2, or 4 colors, each color must be acceptable by pygfx.Color"""
+        self._background_material.set_colors(*colors)
+
```
clewis7, line 271: :DDD

clewis7 (approved):

clewis7:
I think this PR looks really good. 

Things that I think we should consider moving forward:
- where we keep axes and legends (should prob go in a submodule)
- user-facing things 
     - `axes=False` in `show()`
     - `background_color="white"` in `show()` to set a canvas color initially 
- more tests, maybe we should start using `coverage` now that our API is becoming more stable and making sure we are really well-tested
- how much do we care that our gallery images are a bit wonky?

fastplotlib/graphics/_axes.py
kushalkolar: we'll figure this out later

kushalkolar:
Either something is weird with rtd lately or the addition of axes increasing build times dramatically :thinking:

kushalkolar:
I think something is off with rtd, [this build](https://readthedocs.org/projects/fastplotlib/builds/25095344/) yesterday took 500 seconds, IDK why they're taking 1900+ seconds now and being killed for going over time.

kushalkolar:
When I generated the nb test screenshots for image widget, some of them are weird without axes, but then the following ones that test changing things on the image widget, such the cmap of the image or the timepoint look normal with proper axes :man_shrugging: . 

nb-image-widget-single

nb-image-widget-single-gnuplot2

nb-image-widget-movie-single-0

nb-image-widget-movie-single-50

Not just image widget, for example lines initialized:

nb-lines

and then when colors are changed the axes appear normal:

nb-lines-colors

Similar to the previous issues with axes not showing up, or axes not showing up properly in the test screenshots or examples gallery. I wonder if this means we need to do a 2nd draw to make it appear correct due to the axes update animation function? Maybe a forced call to `render()` in `Figure.show()` could solve it.

kushalkolar:
I'm actually going to make these changes:

- [ ] add a property `intersection` instead of `follow`, `if intersection == None` it will follow.
- [ ] make grids invisible by default, I think that users might think it's too much to see them all by default, especially with perspective projections. I think that `pygfx.WgpuRenderer` just skips invisible `WorldObjects` so it won't affect performance, I need to check the renderer again and make sure.
- [ ] make it easier to set grid plane position

kushalkolar:
I FIGURED IT OUT.

So in `subplot.get_rect()` we have:

https://github.com/fastplotlib/fastplotlib/blob/54583166e27b11e3c2a62db34e721c131c30df3d/fastplotlib/layouts/_subplot.py#L169-L172

So we should actually be using `canvas.get_logical_size()` instead. We never noticed because the canvas resize event causes the renderer resize event which is then used to set the viewport rect. But with an offscreen canvas it never resizes :upside_down_face: (I think).

fastplotlib/graphics/_axes.py
```diff
@@ -243,7 +243,7 @@ class Axes:
     def __init__(
             self,
```
clewis7: would be nice to have a docstring here

clewis7 (dismissed):

fastplotlib/graphics/_axes.py
kushalkolar: once it's more user-facing, we can do a second PR later to refactor axes with a more final version once we have an idea of exactly what we want

kushalkolar:
I think it might actually be taking forever at the scatter_size test example which is really bizarre

kushalkolar:
I removed selectors from testutils for now, let's see

kushalkolar:
ok I think the issue actually is when it goes from test_examples_run to test_examples_screenshots, I guess something about how we now have hte render call in `Figure.show()` for the manual offscreen canvas is causing weird things. But only for tests and not for docs gallery

kushalkolar:
Seems like the `viewport.render()` call is blocking CI but only on github actions and not locally even with lavapipe

kushalkolar:
ok tests should pass now, the issue is something in the render call is blocking test_examples

So I modified `Figure.show()` for offscreen canvas to this:

```python
        elif self.canvas.__class__.__name__ == "WgpuManualOffscreenCanvas":
            # for test and docs gallery screenshots
            for subplot in self:
                subplot.set_viewport_rect()
                subplot.axes.update_using_camera()

                # render call is blocking only on github actions for some reason,
                # but not for rtd build, this is a workaround
                # for CI tests, the render call works if it's in test_examples
                # but it is necessary for the gallery images too so that's why this check is here
                if "RTD_BUILD" in os.environ.keys():
                    if os.environ["RTD_BUILD"] == "1":
                        subplot.viewport.render(subplot.scene, subplot.camera)
```

And in `test_example_screenshots` perform the render & flush

```python
    # import the example module
    example = importlib.import_module(module_name)

    for subplot in example.figure:
        subplot.viewport.render(subplot.scene, subplot.camera)
    example.figure.renderer.flush()

    # render a frame
    img = np.asarray(example.figure.renderer.target.draw())
```

clewis7 (approved):

PR #555, opened 2024-07-15
Figure_shape in imagewidget and related histogram widget bug fix

### LINKED ISSUE #553: Histogram zero values
One bug in histogram widget right now:

Whenever we display data with constant value (say, something that is all zeros or all ones), the histogram computations throw a divide by zero error. 

Concretely: 
data = np.zeros((20, 10, 5))
iw = ImageWidget(data)
iw.show()

will throw an error.

### LINKED ISSUE #554: Figure_shape inconsistency in ImageWidget
In the ImageWidget constructor, there is an option to specify the figure shape via the figure_shape argument. You should also be able to specify a 'shape' field in figure_kwargs (that is the point of figure_kwargs -- to customize the Figure object construction).
However, use figure_kwargs to do this, the constructor will throw an error.  

We need to adjust the logic to 
(1) See if the user specified a figure_shape
(2) If not, see if the user specified a 'shape' variable in figure_kwargs

We then use one set of streamlined statements to construct the Figure object.

apasarkar:
closes #553, closes #554

### DIFF
--- a/fastplotlib/widgets/histogram_lut.py
+++ b/fastplotlib/widgets/histogram_lut.py
@@ -25,8 +25,8 @@ class HistogramLUT(Graphic):
         ----------
         data
         image_graphic
-        nbins
-        flank_divisor: float, default 5.0
+        nbins: int, defaut 100. Total number of bins used in the histogram
+        flank_divisor: float, default 5.0. Fraction of empty histogram bins on the tails of the distribution
             set `np.inf` for no flanks
         kwargs
         """
@@ -161,9 +161,10 @@ class HistogramLUT(Graphic):
             hist, edges = np.histogram(data_ss, bins=self._nbins)
 
         # used if data ptp <= 10 because event things get weird
-        # with tiny world objects due to  floating point error
+        # with tiny world objects due to floating point error
         # so if ptp <= 10, scale up by a factor
-        self._scale_factor: int = max(1, 100 * int(10 / np.ptp(data_ss)))
+        data_interval = edges[-1] - edges[0]
+        self._scale_factor: int = max(1, 100 * int(10 / data_interval))
 
         edges = edges * self._scale_factor
 
@@ -178,7 +179,6 @@ class HistogramLUT(Graphic):
         )
 
         edges_flanked = np.concatenate((flank_left, edges, flank_right))
-        np.unique(np.diff(edges_flanked))
 
         hist_flanked = np.concatenate(
             (np.zeros(flank_nbins), hist, np.zeros(flank_nbins))
@@ -186,7 +186,10 @@ class HistogramLUT(Graphic):
 
         # scale 0-100 to make it easier to see
         # float32 data can produce unnecessarily high values
-        hist_scaled = hist_flanked / (hist_flanked.max() / 100)
+        hist_scale_value = hist_flanked.max()
+        if np.allclose(hist_scale_value, 0):
+            hist_scale_value = 1
+        hist_scaled = hist_flanked / (hist_scale_value / 100)
 
         if edges_flanked.size > hist_scaled.size:
             # we don't care about accuracy here so if it's off by 1-2 bins that's fine
--- a/fastplotlib/widgets/histogram_lut.py
+++ b/fastplotlib/widgets/histogram_lut.py
@@ -25,9 +25,10 @@ class HistogramLUT(Graphic):
         ----------
         data
         image_graphic
-        nbins: int, defaut 100. Total number of bins used in the histogram
-        flank_divisor: float, default 5.0. Fraction of empty histogram bins on the tails of the distribution
-            set `np.inf` for no flanks
+        nbins: int, defaut 100.
+            Total number of bins used in the histogram
+        flank_divisor: float, default 5.0.
+            Fraction of empty histogram bins on the tails of the distribution set `np.inf` for no flanks
         kwargs
         """
         super().__init__(**kwargs)
--- a/fastplotlib/widgets/image.py
+++ b/fastplotlib/widgets/image.py
@@ -368,14 +368,24 @@ class ImageWidget:
             if all([_is_arraylike(d) for d in data]):
                 # Grid computations
                 if figure_shape is None:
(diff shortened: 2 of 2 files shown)


fastplotlib/widgets/histogram_lut.py
```diff
@@ -25,8 +25,8 @@ def __init__(
         ----------
         data
         image_graphic
-        nbins
-        flank_divisor: float, default 5.0
+        nbins: int, defaut 100. Total number of bins used in the histogram
+        flank_divisor: float, default 5.0. Fraction of empty histogram bins on the tails of the distribution
```
clewis7, line 29:
```suggestion
        nbins: int, defaut 100
             Total number of bins used in the histogram
        flank_divisor: float, default 5.0
             Fraction of empty histogram bins on the tails of the distribution
```

apasarkar: Done!

fastplotlib/widgets/histogram_lut.py
```diff
@@ -178,15 +180,17 @@ def _calculate_histogram(self, data):
...
             (np.zeros(flank_nbins), hist, np.zeros(flank_nbins))
         )
 
         # scale 0-100 to make it easier to see
         # float32 data can produce unnecessarily high values
-        hist_scaled = hist_flanked / (hist_flanked.max() / 100)
+        hist_scale_value = hist_flanked.max()
+        if np.allclose(hist_scale_value, 0):
+            hist_scale_value = 1
+        hist_scaled = hist_flanked / (hist_scale_value / 100)
```
kushalkolar, line 193: ah so then we just get a flat histogram, all bins have zero weight?

fastplotlib/widgets/image.py
```diff
@@ -368,13 +368,21 @@ def __init__(
             if all([_is_arraylike(d) for d in data]):
                 # Grid computations
                 if figure_shape is None:
-                    figure_shape = calculate_figure_shape(len(data))
+                    if figure_kwargs is not None:
+                        if "shape" in figure_kwargs:
+                            figure_shape = figure_kwargs["shape"]
+                    else:
+                        figure_shape = calculate_figure_shape(len(data))
```
kushalkolar, line 375: what about initialize `figure_kwargs` before here instead of later (see L511), so then we can just do this and remove the one if block:

```python
if figure_shape is None:
  if "shape" in figure_kwargs.keys()():
    figure_shape = figure_kwargs["shape"]
...
```

kushalkolar (changes requested):

kushalkolar:
Thanks! Is the histogram calc the only thing that was causing issues on your end?

fastplotlib/widgets/histogram_lut.py
```diff
@@ -161,9 +162,10 @@ def _calculate_histogram(self, data):
             hist, edges = np.histogram(data_ss, bins=self._nbins)
 
         # used if data ptp <= 10 because event things get weird
-        # with tiny world objects due to  floating point error
+        # with tiny world objects due to floating point error
         # so if ptp <= 10, scale up by a factor
-        self._scale_factor: int = max(1, 100 * int(10 / np.ptp(data_ss)))
+        data_interval = edges[-1] - edges[0]
+        self._scale_factor: int = max(1, 100 * int(10 / data_interval))
```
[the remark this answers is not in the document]
apasarkar: @kushalkolar True. But I think the bigger reason why this is important is that if you do np.histogram on a dataset with constant value (i.e. np.zeros((Dim1, Dim2)), np.ptp will return 0, leading to divide by 0. But the edges[-1] - edges[0] avoids this issue since np.histogram computes a reasonable interval.

fastplotlib/widgets/histogram_lut.py
apasarkar: @kushalkolar Yeah, this is really just an extra check in case an empty array is passed in (in which case the histogram weights will be 0.

fastplotlib/widgets/image.py
apasarkar: Agreed! Just updated this in the PR

apasarkar:
@kushalkolar  I think one thing to note is that the histogram computation right now is nice if the user's data is loaded into RAM. If otoh the user is doing on-the-fly motion correction, where the software is (1) loading data from disk (2) motion correcting it (3) rendering it, the histogram computations will involve a lot of extra work (loading a large chunk of data into memory just to estimate a histogram of the data). Maybe in that case the user should be able to just specify "absolute" vmin and vmax values without a histogram. I find that knowwing the range of the data (via some scale bars) is very useful, even if there is no histogram.

kushalkolar:
> @kushalkolar I think one thing to note is that the histogram computation right now is nice if the user's data is loaded into RAM. If otoh the user is using ImageWidge to visualize on-the-fly motion correction, things get trickier because the image array object is doing the extra work of loading from disk. In this case, the histogram computations will involve a lot of extra work (loading a large chunk of data into memory just to estimate a histogram of the data). Maybe in that case the user should be able to just specify "absolute" vmin and vmax values without a histogram. I find that knowing the range of the data (via some scale bars) is very useful, even if there is no histogram.

Ah so you mean like the lazy arrays you have? What about allowing `HistogramLUTWidget` to take a pre-computed histogram? 

What about data that can be lazy loaded, such as memmaps? We only use a max of 500 x 100 x 100 pixels for computing the histogram, data is subsampled to meet that max number of pixels:

https://github.com/apasarkar/fastplotlib/blob/90af0d4a136d918f99551b3fead28773693e40cf/fastplotlib/widgets/histogram_lut.py#L140-L152

We could benchmark this.

If we add a kwarg to pass in a pre-computed histogram, we can just pass an "empty" histogram with all zero bins, but have a vmin and vmax so that it can be interacted with.

kushalkolar:
is this good to go? LGTM. The other points, such as just operating as a vmin-vmax widget can be done in a later PR

kushalkolar (approved):

apasarkar:
@kushalkolar Missed your comment about the vmin/vmax. Strongly agree we need something like that -- I will submit a pull for it.

PR #567, opened 2024-07-29
change basis of axes

### LINKED ISSUE #561: adding axes to represent a different basis
From my algo dev stuff I realize it would be useful to define an arbitrary basis and create axes based on that. We would basically need set the rotation quaternion.

Maybe an API like:

```python
ruler_other_basis = Ruler(basis_vector=(0.5, 0.5))  # define basis in 
```

For example, to display the axes for the othogonal basis represented by the two white points here:

image

@BalzaniEdoardo I think you might have ideas on how to do this :smile:

kushalkolar:
closes #561 

WIP, seems to be ok in 2D, something wonky in 3D. Those blue axes should be pointing out towards the red points. And one ruler is completely MIA

download

using [`pylinalg.quat_from_vecs`](https://pylinalg.readthedocs.io/en/latest/reference.html#pylinalg.quat_from_vecs) to create the quaternion to rotate the axes world object.

### DIFF
--- a/fastplotlib/graphics/_axes.py
+++ b/fastplotlib/graphics/_axes.py
@@ -1,9 +1,16 @@
 import numpy as np
 
 import pygfx
+from pylinalg import quat_from_vecs, vec_transform_quat
 
 
 GRID_PLANES = ["xy", "xz", "yz"]
+
+CANONICAL_BAIS = np.array([
+    [1., 0., 0.],
+    [0., 1., 0.],
+    [0., 0., 1.]
+])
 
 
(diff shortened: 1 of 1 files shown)


kushalkolar:
it was an [indexing error](https://github.com/fastplotlib/fastplotlib/pull/567/commits/8c46029d9be6873913851cc097ff7af0fda095a9) :upside_down_face: 

This is really nice :smile: 

download

I think once I figure out how to get the new (xmin, ymin), (xmax, ymax) for the rulers to account for rotation good to go.

clewis7 (approved):

Issue #142, opened 2023-03-02
region selectors


kushalkolar:
- [ ] LinearRegionSelector - similar to pyqtgraph LinearRegionItem
- [ ] RectangleRegionSelection
- [ ] PolygonRegionSelection
- [ ] VolumeRegionSelection?

PR #576, opened 2024-08-02
add rectangular region selector

clewis7:
#142 Rectangle selector

- [x] rectangle selector implementation 
- [x] add selector methods for graphics (e.g. `LineGraphic`, `ImageGraphic`, etc.)
- [x] `get_selected_indices()` and `get_selected_data()` methods
     - [x] add a `mode` argument, one of  `"partial", "full", "ignore"` that handles when selector is not covering part of graphic
- [x] add vertices on the corners

### DIFF
--- a/examples/desktop/selectors/rectangle_selector.py
+++ b/examples/desktop/selectors/rectangle_selector.py
@@ -8,42 +8,53 @@ Example showing how to use a `RectangleSelector` with lines, line collections, a
 # test_example = false
 # sphinx_gallery_pygfx_docs = 'screenshot'
 
-import imageio.v3 as iio
 import numpy as np
 import fastplotlib as fpl
+from itertools import product
 
 # create a figure
 figure = fpl.Figure(
-    shape=(2, 1),
     size=(700, 560)
 )
 
+
+# generate some data
+def make_circle(center, radius: float, n_points: int = 75) -> np.ndarray:
+    theta = np.linspace(0, 2 * np.pi, n_points)
+    xs = radius * np.sin(theta)
+    ys = radius * np.cos(theta)
+
+    return np.column_stack([xs, ys]) + center
+
+
+spatial_dims = (50, 50)
+
+circles = list()
+for center in product(range(0, spatial_dims[0], 9), range(0, spatial_dims[1], 9)):
+    circles.append(make_circle(center, 3, n_points=75))
+
+pos_xy = np.vstack(circles)
+
 # add image
-image_graphic = figure[0, 0].add_image(data=iio.imread("imageio:camera.png"))
+line_collection = figure[0, 0].add_line_collection(circles, cmap="jet", thickness=5)
 
 # add rectangle selector to image graphic
-rectangle_selector = image_graphic.add_rectangle_selector()
-
-# add a zoomed plot of the selected data
-zoom_ig = figure[1, 0].add_image(rectangle_selector.get_selected_data())
+rectangle_selector = line_collection.add_rectangle_selector()
 
 
-# add event handler to update the data of the zoomed image as the selection changes
+# add event handler to highlight selected indices
 @rectangle_selector.add_event_handler("selection")
-def update_data(ev):
-    # get the new data
-    new_data = ev.get_selected_data()
+def color_indices(ev):
+    line_collection.cmap = "jet"
+    ixs = ev.get_selected_indices()
 
-    # remove the old zoomed image graphic
-    global zoom_ig
+    # iterate through each of the selected indices, if the array size > 0 that mean it's under the selection
+    selected_line_ixs = [i for i in range(len(ixs)) if ixs[i].size > 0]
+    line_collection[selected_line_ixs].colors = "w"
 
-    figure[1, 0].remove_graphic(zoom_ig)
 
-    # add new zoomed image of new data
-    zoom_ig = figure[1, 0].add_image(data=new_data)
-
-    # autoscale the plot
-    figure[1, 0].auto_scale()
+# manually move selector to make a nice gallery image :D
+rectangle_selector.selection = (15, 30, 15, 30)
 
 
 figure.show()
--- a/examples/desktop/selectors/rectangle_selector.py
+++ b/examples/desktop/selectors/rectangle_selector.py
@@ -46,6 +46,8 @@ def update_data(ev):
     figure[1, 0].auto_scale()
 
 
+figure.show()
+
 # NOTE: `if __name__ == "__main__"` is NOT how to use fastplotlib interactively
 # please see our docs for using fastplotlib interactively in ipython and jupyter
 if __name__ == "__main__":
--- a/examples/desktop/selectors/rectangle_selector.py
+++ b/examples/desktop/selectors/rectangle_selector.py
@@ -2,7 +2,7 @@
 Rectangle Selectors
 ===================
 
-Example showing how to use a `RectangleSelector` with lines, line collections, and images
+Example showing how to use a `RectangleSelector` with line collections
 """
 
 # test_example = false
--- a/examples/desktop/selectors/rectangle_selector.py
+++ b/examples/desktop/selectors/rectangle_selector.py
@@ -1,0 +1,53 @@
+"""
+Rectangle Selectors
+===================
+
+Example showing how to use a `RectangleSelector` with lines, line collections, and images
+"""
+
+# test_example = false
+# sphinx_gallery_pygfx_docs = 'screenshot'
+
+import imageio.v3 as iio
+import numpy as np
+import fastplotlib as fpl
+
+# create a figure
+figure = fpl.Figure(
+    shape=(2, 1),
+    size=(700, 560)
+)
+
+# add image
+image_graphic = figure[0, 0].add_image(data=iio.imread("imageio:camera.png"))
+
+# add rectangle selector to image graphic
+rectangle_selector = image_graphic.add_rectangle_selector()
+
+# add a zoomed plot of the selected data
+zoom_ig = figure[1, 0].add_image(rectangle_selector.get_selected_data())
+
+
+# add event handler to update the data of the zoomed image as the selection changes
+@rectangle_selector.add_event_handler("selection")
+def update_data(ev):
+    # get the new data
+    new_data = ev.get_selected_data()
+
+    # remove the old zoomed image graphic
+    global zoom_ig
+
+    figure[1, 0].remove_graphic(zoom_ig)
+
+    # add new zoomed image of new data
+    zoom_ig = figure[1, 0].add_image(data=new_data)
+
+    # autoscale the plot
+    figure[1, 0].auto_scale()
+
+
+# NOTE: `if __name__ == "__main__"` is NOT how to use fastplotlib interactively
+# please see our docs for using fastplotlib interactively in ipython and jupyter
+if __name__ == "__main__":
+    print(__doc__)
+    fpl.run()
--- a/fastplotlib/graphics/_features/_selection_features.py
+++ b/fastplotlib/graphics/_features/_selection_features.py
@@ -192,7 +192,7 @@ class LinearRegionSelectionFeature(GraphicFeature):
         #  this is probably a good idea so that the data isn't sliced until it's actually necessary
 
 
-class RectangleRegionSelectionFeature(GraphicFeature):
+class RectangleSelectionFeature(GraphicFeature):
     """
     **additional event attributes:**
 
--- a/fastplotlib/graphics/_features/_selection_features.py
+++ b/fastplotlib/graphics/_features/_selection_features.py
@@ -213,11 +213,12 @@ class RectangleSelectionFeature(GraphicFeature):
     +----------+------------+-------------------------------------------+
 
     """
+
     def __init__(
-            self,
-            value: tuple[float, float, float, float],
-            axis: str | None,
-            limits: tuple[float, float, float, float]
+        self,
+        value: tuple[float, float, float, float],
+        axis: str | None,
+        limits: tuple[float, float, float, float],
     ):
         super().__init__()
 
@@ -317,7 +318,6 @@ class RectangleSelectionFeature(GraphicFeature):
         selector.edges[3].geometry.positions.data[:] = np.array(
             [[xmin, ymax, z], [xmax, ymax, z]]
         )
-
 
         # change the vertice positions
 
--- a/fastplotlib/graphics/_features/_selection_features.py
+++ b/fastplotlib/graphics/_features/_selection_features.py
@@ -206,26 +206,29 @@ class RectangleSelectionFeature(GraphicFeature):
 
     **info dict:**
 
-    +----------+------------+-----------------------------+
-    | dict key | value type | value description           |
-    +==========+============+=============================+
-    | value    | np.ndarray | new [min, max] of selection |
-    +----------+------------+-----------------------------+
+    +----------+------------+-------------------------------------------+
+    | dict key | value type | value description                         |
+    +==========+============+===========================================+
+    | value    | np.ndarray | new [xmin, xmax, ymin, ymax] of selection |
+    +----------+------------+-------------------------------------------+
 
     """
     def __init__(
-        self, value: Tuple[float, float, float, float], axis: str | None, limits: Tuple[int, int]
+            self,
+            value: tuple[float, float, float, float],
+            axis: str | None,
+            limits: tuple[float, float, float, float]
     ):
         super().__init__()
 
         self._axis = axis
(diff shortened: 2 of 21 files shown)


clewis7:
https://github.com/user-attachments/assets/5568a327-345e-468f-948c-a93ddbbd9b04

fastplotlib/graphics/selectors/_rectangle_region.py
```diff
@@ -188,91 +89,105 @@ def __init__(
...
-        # position data for the left edge line
+        group.add(self.fill)
+
+        #position data for the left edge line
         left_line_data = np.array(
             [
-                [origin[0], (-height / 2) + origin[1], 0.5],
-                [origin[0], (height / 2) + origin[1], 0.5],
+                [0, -width / 2, 0],
+                [0, width / 2, 0],
```
kushalkolar, line 142: can this now just be `[xmin, ymin, 0], [xmin, ymax, 0]` ? And similarly for the other lines

clewis7:
https://github.com/user-attachments/assets/bd583135-2628-4a77-87ff-4ee567ae7874

clewis7:
**Rectangle Selector**

- [ ] rectangle selector implementation 
- [ ] add selector methods for graphics (e.g. `LineGraphic`, `ImageGraphic`, etc.)
- [ ] `get_selected_indices()` and `get_selected_data()` methods
     - [ ] add a `mode` argument, one of  `"partial", "full", "ignore"` that handles when selector is not covering part of graphic

fastplotlib/graphics/image.py
```diff
@@ -393,3 +393,45 @@ def add_linear_region_selector(
...
+        selection: (float, float, float, float), optional
+            initial (xmin, xmax, ymin, ymax) of the selection
+        axis: str, default None
+            Optional string to restrict the movement of the selector along one axis. If passed, should
+            be one of "x" or "y".
+        fill_color: (float, float, float), optional
+            The fill color of the selector.
+        """
+
+        # default selection is 25% of the image
```
kushalkolar, line 419: 25% along the diagonal might be better if the array is very tall or wide?

kushalkolar:
I wonder if a method that returns a mask and then using the mask for `get_selected_data()` could be useful, since it is about 20% faster:

```python
620 μs ± 7.79 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [21]: %%timeit
    ...: a[np.where((a > 50_000) & (a < 100_000))]
    ...: 
    ...: 
747 μs ± 31.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [22]: %%timeit
    ...: a[np.where((a > 50_000) & (a < 100_000))]
    ...: 
    ...: 
```

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -199,6 +199,43 @@ def __init__(
...
 
+        # vertices
+        top_left_vertex_data = (xmin, ymax, 1)
+        top_right_vertex_data = (xmax, ymax, 1)
+        bottom_left_vertex_data = (xmin, ymin, 1)
+        bottom_right_vertex_data = (xmax, ymin, 1)
+
+        top_left_vertex = pygfx.Points(
+            pygfx.Geometry(positions=[top_left_vertex_data], sizes=[edge_thickness]),
+            pygfx.PointsMaterial(size=edge_thickness, color=edge_color, size_space="world", size_mode="vertex"),
```
kushalkolar: let's do square points for resize-handlers, since that's commonly used in things like photoshop etc. circles tend to be rotate handles, but let's not worry about rotations here.

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -199,6 +199,43 @@ def __init__(
...
+
+        self.vertices: Tuple[pygfx.Points, pygfx.Points, pygfx.Points, pygfx.Points] = (
+            bottom_left_vertex,
+            bottom_right_vertex,
+            top_left_vertex,
+            top_right_vertex,
+        )
+
+        for vertex in self.vertices:
+            vertex.world.z = -0.25
```
kushalkolar, line 236: why -0.25 :thinking:

fastplotlib/graphics/selectors/_rectangle.py
kushalkolar: and a different color for the vertices

fastplotlib/graphics/selectors/_rectangle.py
clewis7: the edges get set to -0.5, so if the vertices are -0.25 then they will be on top of the edges

fastplotlib/graphics/_features/_selection_features.py
```diff
@@ -190,3 +190,168 @@ def set_value(self, selector, value: Sequence[float]):
...
+    def value(self) -> np.ndarray[float]:
+        """
+        (xmin, xmax, ymin, ymax) of the selection, in data space
+        """
+        return self._value
+
+    @property
+    def axis(self) -> str:
+        """one of "x" | "y" """
+        return self._axis
```
kushalkolar, line 239: what does this do for a rectangle selector?

fastplotlib/graphics/_features/_selection_features.py
```diff
@@ -190,3 +190,168 @@ def set_value(self, selector, value: Sequence[float]):
...
+        selector.edges[2].geometry.positions.data[:] = np.array(
+            [[xmin, ymin, z], [xmax, ymin, z]]
+        )
+
+        # top line
+        selector.edges[3].geometry.positions.data[:] = np.array(
+            [[xmin, ymax, z], [xmax, ymax, z]]
+        )
+
+        # change the vertice positions
```
kushalkolar, line 322: vertex

fastplotlib/graphics/image.py
```diff
@@ -393,3 +394,48 @@ def add_linear_region_selector(
         selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
 
         return selector
+
+    def add_rectangle_selector(
+        self,
+        selection: tuple[float, float, float, float] = None,
+        axis: str = None,
+        fill_color=(0, 0, 0.35, 0.2),
```
kushalkolar, line 402: let's just let them pass kwargs like the other selectors

fastplotlib/graphics/image.py
```diff
@@ -393,3 +394,48 @@ def add_linear_region_selector(
...
+
+        Parameters
+        ----------
+        selection: (float, float, float, float), optional
+            initial (xmin, xmax, ymin, ymax) of the selection
+        axis: str, default None
+            Optional string to restrict the movement of the selector along one axis. If passed, should
+            be one of "x" or "y".
+        fill_color: (float, float, float), optional
+            The fill color of the selector.
```
kushalkolar, line 417: just **kwargs

fastplotlib/graphics/line.py
```diff
@@ -216,6 +216,60 @@ def add_linear_region_selector(
         # so we should only work with a proxy on the user-end
         return selector
 
+    def add_rectangle_selector(
+        self,
+        selection: tuple[float, float, float, float] = None,
+        axis: str = None,
+        fill_color=(0, 0, 0.35, 0.2),
```
kushalkolar, line 223:
```suggestion
```

just keep ***kwargs

fastplotlib/graphics/line_collection.py
```diff
@@ -447,6 +447,56 @@ def add_linear_region_selector(
         # so we should only work with a proxy on the user-end
         return selector
 
+    def add_rectangle_selector(
+        self,
+        selection: tuple[float, float, float, float] = None,
+        axis: str = None,
+        fill_color=(0, 0, 0.35, 0.2),
```
kushalkolar, line 454:
```suggestion
```

just keep ***kwargs

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -0,0 +1,511 @@
...
+        selection: (float, float, float, float)
+            the initial selection of the rectangle, ``(x_min, x_max, y_min, y_max)``
+
+        limits: (float, float, float, float)
+            limits of the selector, ``(x_min, x_max, y_min, y_max)``
+
+        axis: str, default ``None``
+            Restrict the selector to the "x" or "y" axis.
+            If the selector is restricted to an axis you cannot change the selection along the other axis. For example,
+            if you set ``axis="x"``, then the ``y_min``, ``y_max`` values of the selection will stay constant.
```
kushalkolar, line 81: does this make sense here? I would imagine a `LinearRegionSelector` should be used if it needs to be restricted to an axis

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -0,0 +1,511 @@
...
+        ----------
+        graphic: Graphic, optional, default ``None``
+            if provided, returns the data selection from this graphic instead of the graphic set as ``parent``
+        mode: str, default 'full'
+            One of 'full', 'partial', or 'ignore'. Indicates how selected data should be returned based on the
+            selectors position over the graphic. Only used for ``LineGraphic``, ``LineCollection``, and ``LineStack``
+            If 'full', will return all data bounded by the x and y limits of the selector even if partial indices
+            alone one axis are not fully covered by the selector.
+            If 'partial' will return only the data that is bounded within the limits.
+            If 'ignore', will not return data if selector is not fully covering the graphic along the axes bounds.
```
kushalkolar, line 317: line separaters so it doesn't display as one paragraph in the docs, example:

```python
mode: str, default "full"
    | If ....
    | If another...
    | If bah
```

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -0,0 +1,511 @@
...
+
+        # selector (xmin, xmax, ymin, ymax) values
+        bounds = self.selection
+
+        # image data does not need to check for mode because the selector is always bounded
+        # to the image
+        if "Image" in source.__class__.__name__:
+            ys = np.arange(bounds[0], bounds[1], dtype=int)
+            xs = np.arange(bounds[2], bounds[3], dtype=int)
+            return [xs, ys]
```
kushalkolar, line 441:
```suggestion
            xs = np.arange(bounds[0], bounds[1], dtype=int)
            ys = np.arange(bounds[2], bounds[3], dtype=int)
            return [xs, ys]
```

Image data is [row, cols], same as [y-indices, x-indices]

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -0,0 +1,511 @@
...
+
+        Parameters
+        ----------
+        graphic: Graphic, default ``None``
+            If provided, returns the selection indices from this graphic instrad of the graphic set as ``parent``
+
+        Returns
+        -------
+        Union[np.ndarray, List[np.ndarray]]
+            data indicies of the selection, list of np.ndarray if the graphic is a collection
```
kushalkolar, line 428: For image is it `list[x_indices_array, y_indices_array]` and for line/line collection what is the shape =S ? Can specify here

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -0,0 +1,511 @@
...
+        source = self._get_source(graphic)
+        ixs = self.get_selected_indices(source)
+
+        # do not need to check for mode for images, because the selector is bounded by the image shape
+        # will always be `full`
+        if "Image" in source.__class__.__name__:
+            s_x = slice(ixs[0][0], ixs[0][-1] + 1)
+            s_y = slice(ixs[1][0], ixs[1][-1] + 1)
+
+            return source.data[s_x, s_y]
```
kushalkolar, line 333:
```suggestion
            col_ixs = slice(ixs[0][0], ixs[0][-1] + 1)
            row_ixs = slice(ixs[1][0], ixs[1][-1] + 1)

            return source.data[row_ixs, col_ixs]
```

Depends on a suggested change below

kushalkolar (commented):
Wondering if you tried the selector with different offsets for the parent graphic, does it return the correct `selection`, and correct arrays for `get_selected_indices()` and `get_selected_data()`.

kushalkolar:
For collection a helper method that gets indices or returns graphics under the selection could be useful. Not hard to implement currently, we can think about and add more helper functions later :smile: 

```python
line_collection = figure[0, 0].add_line_collection(circles, cmap="jet", thickness=5)

rs = line_collection.add_rectangle_selector()

@rs.add_event_handler("selection")
def bah(ev):
    line_collection.cmap = "jet"
    ixs = ev.get_selected_indices()

    # iterate through each of the selected indices, if the array size > 0 that mean it's under the selection
    selected_line_ixs = [i for i in range(len(ixs)) if ixs[i].size > 0]
    line_collection[selected_line_ixs].colors = "w"
```

https://github.com/user-attachments/assets/0aefa776-1ee5-4150-b449-423b386246c2

fastplotlib/graphics/selectors/_rectangle.py
clewis7: what do you mean?

fastplotlib/graphics/selectors/_rectangle.py
clewis7: I agree, but I think you had it left over from the first implementation you did awhile ago...I will remove

examples/desktop/selectors/rectangle_selector.py
kushalkolar, line 1: Keep both examples, the image one is also great and very useful!!!

fastplotlib/graphics/selectors/_rectangle.py
kushalkolar: Sometimes world objects will have weird overlapping when the z-planes are close with very high or very low zoom, I just tried it and it seems fine within reasonable levels.

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -311,26 +302,27 @@ def get_selected_data(
...
         source = self._get_source(graphic)
         ixs = self.get_selected_indices(source)
 
         # do not need to check for mode for images, because the selector is bounded by the image shape
         # will always be `full`
         if "Image" in source.__class__.__name__:
-            s_x = slice(ixs[0][0], ixs[0][-1] + 1)
-            s_y = slice(ixs[1][0], ixs[1][-1] + 1)
+            row_ixs = slice(ixs[0][0], ixs[0][-1] + 1)
+            col_ixs = slice(ixs[1][0], ixs[1][-1] + 1)
```
kushalkolar, line 323: other way around, xs are columns, ys are rows

kushalkolar (changes requested):

fastplotlib/graphics/image.py
```diff
@@ -393,3 +394,44 @@ def add_linear_region_selector(
         selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
 
         return selector
+
+    def add_rectangle_selector(
+        self,
+        selection: tuple[float, float, float, float] = None,
+        fill_color=(0, 0, 0.35, 0.2),
```
kushalkolar, line 401:
```suggestion
```

fastplotlib/graphics/image.py
```diff
@@ -393,3 +394,44 @@ def add_linear_region_selector(
...
+
+            selection = (0, int(diagonal / 4), 0, int(diagonal / 4))
+
+        # min/max limits are image shape
+        limits = (0, self._data.value.shape[0], 0, self._data.value.shape[1])
+
+        selector = RectangleSelector(
+            selection=selection,
+            limits=limits,
+            fill_color=fill_color,
```
kushalkolar, line 427:
```suggestion
```

kushalkolar (changes requested):

hmaarrfk:
Chiming in here because Kushal showed this to me IRL yesterday. Pretty cool!!!!

I had two comments:

1. We once had pretty bad performance from `pygfx_object.world.matrix` so we ended up caching it ourselves.... maybe i should open an issue to pygfx on that. It made iterating through the list of "selectable objects" we had kinda slow.
2. Bonus points for thinking about how to do this in 3D. We didn't get to that part either....
3. Bonus points for doing this all on the GPU in WGSL and just make "highlighting" a property of your shaders ;)

Keep it going. I love to see these demos!

fastplotlib/graphics/image.py
clewis7: I think for images we want to keep this kwarg because the alpha value needs to be 0.2 which is different than the default for adding a rectangle selector for lines or line collections

kushalkolar (approved):

PR #595, opened 2024-08-26
add lorenz attractor example

clewis7:
Had way too much fun doing this :D 

@kushalkolar just a fun example for the gallery

https://github.com/user-attachments/assets/b37da96b-3615-402f-86cb-4036e141f7a3

### DIFF
--- a/examples/desktop/misc/lorenz_animation.py
+++ b/examples/desktop/misc/lorenz_animation.py
@@ -2,7 +2,7 @@
 Lorenz System Animation
 =======================
 
-Example the Lorenz attractor.
+Example of the Lorenz attractor.
 """
 
 # test_example = false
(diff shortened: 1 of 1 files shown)


kushalkolar:
@jpughesanford your fav thing to plot 😁

examples/desktop/misc/lorenz_animation.py
```diff
@@ -0,0 +1,92 @@
+"""
+Lorenz System Animation
+=======================
+
+Example the Lorenz attractor.
```
kushalkolar, line 5: of

examples/desktop/misc/lorenz_animation.py
```diff
@@ -0,0 +1,92 @@
...
+    x_dot = s * (y - x)
+    y_dot = r * x - y - x * z
+    z_dot = x * y - b * z
+    return np.array([x_dot, y_dot, z_dot])
+
+
+dt = 0.01
+num_steps = 3_000
+
+lorenz_data = list()
```
kushalkolar, line 40: preallocate an array to account for the 3,000 timesteps instead of appending to a list?

examples/desktop/misc/lorenz_animation.py
```diff
@@ -0,0 +1,92 @@
...
+    cameras="3d",
+    controller_types="fly"
+)
+
+lorenz_line = figure[0, 0].add_line_collection(data=lorenz_data, thickness=.1, cmap="tab10")
+
+scatter_markers = list()
+
+for graphic in lorenz_line:
+    marker = figure[0, 0].add_scatter(graphic.data.value[0], sizes=8)
```
kushalkolar, line 62: use color from the corresponding line?

kushalkolar (changes requested):

kushalkolar:
mostly LGTM! Few minor comments

I think it would be nice to also have another example that visualizes it in this way: https://en.wikipedia.org/wiki/File:A_Lorenz_system.ogv 

Maybe using a rolling buffer to show the most recent n_timepoints?

clewis7:
>I think it would be nice to also have another example that visualizes it in this way: https://en.wikipedia.org/wiki/File:A_Lorenz_system.ogv

@kushalkolar this would be cool, I can work on some time this coming week

kushalkolar (approved):

PR #608, opened 2024-09-17
Cursor tool

kushalkolar:
Make it possible to easily use the cursor to point at the same positions w.r.t. world space across subplots

Idea: when added to a subplot, a new world object is created and added to that subplot. All the world objects across the various subplots use the same buffer.

WIP

### DIFF
--- a/fastplotlib/tools/_crosshair.py
+++ b/fastplotlib/tools/_crosshair.py
@@ -1,0 +1,97 @@
+from functools import partial
+
+import numpy as np
+
+import pygfx
+
+
+from ..graphics._features import VertexPositions, UniformColor, Thickness
+from ..layouts._subplot import Subplot
+
+
+class CrosshairTool:
+    def __init__(
+            self,
+            subplots: Subplot | list[Subplot],
+            thickness: float = 1.0,
+            style: str = "+",
+            color: str | tuple | np.ndarray = "w",
+            alpha: float = 0.5,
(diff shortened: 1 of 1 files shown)


kushalkolar:
It would be useful to add crosshairs to multiple graphics in a subplot, in this case it can use the bbox of a graphic as the reference space that it lives in. This is particularly useful when adding multiple images to a subplot in a grid.

kushalkolar:
I think I have some better ideas on this after playing with using a scatter dot.

1. Allow adding a `Cursor` to subplots or `Graphics`
2. The cursor's current "source" position is where the pointer currently is. Either the subplot or the graphic the pointer is on.
3. When added to graphics, the cursor's current "source" position is based on the graphic the mouse pointer is currently on. When added to a graphic the cursor's data is always in the same space as the parent graphic's data, i.e. `Cursor.data` maps directly to `Graphic.data`. The offset of an individual cursor matches the offset of the graphic is is on top of.
4. Allow a cursor to be a scatter point of any shape, one line, or two lines. If a line, then the endpoints are based on the bbox of the graphic or the subplot's extents.

kushalkolar:
@hmaarrfk, thought you might be interested too, outline of what I'm going to implement is above

hmaarrfk:
interesting work!

kushalkolar:
superseeded by #662

Issue #626, opened 2024-09-28
sphinx gallery has a memory leak


kushalkolar:
It executes examples and things stick around in RAM, so if an example loads a 1GB video then that sticks around. Can figure this out later, for now we can display heavily subsampled things etc. in the galleries.

This might really become an issue when we want to make example galleries to include computer vision and other scientific examples. Maybe we can find a reliable lazy loader. 

@almarklein wondering if you guys ever ran into this issue?

almarklein:
mmm I'm not sure. We don't have examples that are *that* large. IIUC what happens if you have two examples that load 1GB of data, you'd expect that after running the two examples, the allocated ram  by the process is at most 1GB, but it's now 2GB, right?

BTW some time ago we ran all examples in a new subprocess, which avoids these kinds of errors. But the overhead is quite substantial, which is why we moved to a single-process solution, making tests and gallery generation *a lot* faster. On the other hand, using subprocesses would make it possible to parallelize ... mmm 🤔

kushalkolar:
> IIUC what happens if you have two examples that load 1GB of data, you'd expect that after running the two examples, the allocated ram by the process is at most 1GB, but it's now 2GB, right?

Yup it seems like this is what's happening when I build it locally, video examples for instance: https://fastplotlib.readthedocs.io/en/latest/_gallery/image_widget/image_widget_single_video.html#sphx-glr-gallery-image-widget-image-widget-single-video-py 

Right now we're subsampling but still coming up against a limit, maybe we'll move to some github actions based doc build system and use github pages. We eventually want to have larger examples. The build takes 1.5GB on my computer, but I guess there's overhead so it's maxing out the 7GB limit on rtd.

almarklein:
Let's compare the different implementations:

* https://github.com/fastplotlib/fastplotlib/blob/main/examples/tests/test_examples.py
* https://github.com/pygfx/wgpu-py/blob/main/examples/tests/test_examples.py
* https://github.com/pygfx/pygfx/blob/main/examples/tests/test_examples.py

In wgpu-py we use `request.addfinalizer(unload_module)` to clear the module after the test. The pygfx version was updated some time ago, it uses a custom `import_from_path()` function that does not add the module to `sys.modules`, so stuff gets cleaned up automatically as it leaves scope. In fastplotlib,`importlib.import_module()` is used, but I see no logic that removes the module from `sys.modules`, so its indeed expected that everything stays alive :)

[...]

PR #627, opened 2024-09-30
Reorganize examples

### LINKED ISSUE #564: Reorganize examples
I think we should:

- [ ] remove all nb examples except for quickstart, test_gc, and for now imagewidget. Once we have imgui integration we can remove imagewidget as well and just have the quickstart notebook
- [ ] move `examples/desktop` to just `examples/` and have these as the main examples for everything, docs, tests, users, etc. Move any examples that are currently only in the notebook to here.

 @clewis7 thoughts? I think this will make everything easier to maintain, and easier for users too since there would only be one set of examples.

### LINKED ISSUE #624: qt in examples gallery
would be nice to display the qt window screenshots, not sure if it's possible though with the offscreen rendering, but we can at least have the examples in the docs without the screenshot

kushalkolar:
- [x] remove all nb examples except for: test_gc, image_widget, quickstart, and zmq
- [x] move relevant nb examples to the new flat universal examples dir. This included selector tools and line cmap stuff, and a scatter sizes example
- [x] selector tools and line cmap stuff in gallery, made a new scatter sizes animation example which is simpler
- [x] fixes a bug with selector tools where the offset on both dimensions were not being used from the parent graphic 
- [x] move all examples from `examples/desktop` to just `examples/`
- [x] fix a bug where Qt-related wgpu stuff wasn't being properly imported and selected, we simply must make sure to import the wgpu gui stuff before anything else in fpl
- [x] qt code-only examples in the examples gallery so that people don't have to dig through the fastplotlib source tree on github to get any examples
- [ ] update main README.md with changes
- [ ] test pass and docs build

closes #624, closes #564


kushalkolar:
running out of RAM with rtd builds again, seriously what is up with rtd :man_facepalming: 

will figure out later, but @clewis7 you can review before I play wack-a-mole again to find the RAM culprit :weary:

kushalkolar:
Idea: until sphinx can give us more resources or we think of an elegant way to generate docs using github actions build just the gallery locally when changing examples and upload that, images will go into git lfs anyways

I tried [this](https://github.com/dfm/rtds-action) and it's very messy and not maintainable

Last thing I'll try before we restort to that is this: https://stackoverflow.com/questions/72089650/how-to-host-multiple-version-of-a-sphinx-based-documentation-on-github 

The yaml is here: https://github.com/brechtm/rinohtype/blob/master/.github/workflows/tests.yml

kushalkolar:
ok I used Almar's suggestion from #626 which gc after each example is run so the RAM usage is reduced significantly, when I build locally it doesn't exceed ~1.25GB which is basically nothing. If this still overruns the RAM on rtd build servers then we need to do something else :man_shrugging: . If RTD has 7GB I don't understand how there can be 5.5GB of overhead?!

kushalkolar:
ok I give up, we need to move to something else or see if RTD will increase our RAM limit.

kushalkolar:
so builds successfully on github actions, ideas on how to deploy versioned gh pages: https://stackoverflow.com/questions/47643881/github-pages-maintaining-multiple-versions

kushalkolar:
This seems like the most viable way to get versioning: https://github.com/sphinx-contrib/multiversion 

their workflow is insightful since it uses github pages

kushalkolar:
This is all becoming a mess, let's switch to the `pydata-sphinx` theme which has a dropdown menu version switcher built-in: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html

kushalkolar:
ok let's get our own webhost and just copy artifacts via ssh: https://github.com/appleboy/scp-action

PR #636, opened 2024-10-06
see if docs get deployed

kushalkolar:


kushalkolar:
ok so this works, just need to organize stuff a bit, do it with the pydata theme so we can use the versions dropdown menu and some if statements in the build so it puts stuff into a specific dir based on if it's from a PR, main, or a release.

PR #638, opened 2024-10-07
fix auto canvas detection

kushalkolar:
the auto gui selection code must be the first thing that runs before we run any other WGPU code for auto-canvas detection to happen correctly


kushalkolar:
failure seems to be with downloading imageio test image lol, re-running

kushalkolar:
@clewis7 this should be gtg!

clewis7 (approved):

PR #639, opened 2024-10-07
docs reorganization take 3

### LINKED ISSUE #564: Reorganize examples

### LINKED ISSUE #624: qt in examples gallery

kushalkolar:
Changes to the actual docs:

- [x] remove all nb examples except for: test_gc, image_widget, quickstart, and zmq
- [x] move relevant nb examples to the new flat universal examples dir. This included selector tools and line cmap stuff, and a scatter sizes example
- [x] selector tools and line cmap stuff in gallery, made a new scatter sizes animation example which is simpler
- [x] fixes a bug with selector tools where the offset on both dimensions were not being used from the parent graphic 
- [x] move all examples from `examples/desktop` to just `examples/`
- [x] qt code-only examples in the examples gallery so that people don't have to dig through the fastplotlib source tree on github to get any examples
- [x] update main README.md with changes, still need to wait for next push to main and release so that all the links can be updated. We can consolidate more stuff from the readme into the docs later.
- [x] test pass and docs build

closes #624, closes #564

Changes to CI/CD:

We're now hosting docs on our own. They are built with github actions since RTD has a very low limit of 7GB which we are hitting despite sphinx-build only taking a max of ~1.25GB when I build locally :man_shrugging: . 

So the docs are built on github actions and then uploaded to our file host via FTP. This is the action we're using to upload: https://github.com/SamKirkland/FTP-Deploy-Action

Switched to pydata-sphinx-theme since it has built-in support for versioning: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html . The MNE Project is a good resource to see how to implement it: https://github.com/mne-tools/mne-python/tree/main/doc 

Basically we can push to different subdirs on our host and the dropdown switcher can point to those various subdirs. I think we can test it properly and make any changes once we do a release. The action we're using supports uploading the docs to specific subdirs. There's a `switcher.json` file which we has to be populated with the names and links to each version, but we can automate generating that after the next release.

For our own use, the same action will also upload to a subdir based on the PR name that we can use to see what the built docs look like online. The PR builds won't be available in the dropdown menu but we can access them like this: `fastplotlib.org/ver/pr-name` . I also made it so a github actions will add a comment to the PR with a link to the docs built from the PR after they're built and uploaded :D. You can see the workflow yaml, it's pretty simple.


kushalkolar:
ok great the iris file has stayed on lfs, no nonsense conflicts, will pick the rest of the files from the other branch :melting_face:

kushalkolar:
@clewis7 ready for review!

clewis7 (approved):
Docs look amazing :DDDD

PR #643, opened 2024-10-08
remove ipywidget code from selectors

kushalkolar:
removes ipywidget code from selectors since we've now moved to imgui and shouldn't have any GUI-framework specific code in core fastplotlib

superseeds #631 since there was an issue with an lfs file


kushalkolar:
ok I was being lazy and just checked out the file from the other branch so it doesn't have the offsets fix, will do tomorrow

PR #644, opened 2024-10-08
remove ipywidget code from selectors and cleanup docstrings

kushalkolar:
since we've not switched to imgui we shouldn't have any GUI specific code in the core library, this really cleans up the selectors :smile:


kushalkolar:
Forgot to regen API docs, I really need to use precommit

kushalkolar:
@clewis7 gtg after review, I removed ipywidget code within the selectors but I also cleaned up the docstrings, there were some remnants stating "slider" from over a year ago.

kushalkolar:
Let's do that separately

PR #657, opened 2024-10-17
colorbar ruler

kushalkolar:
Just a starting point. We need to figure out how to include text in screen space when calculating bboxes for `PlotArea.auto_scale()` so that the tick labels show up properly.


kushalkolar:
closing since this is now in #868

PR #660, opened 2024-10-18
try adding mac to ci

kushalkolar:
macos-latest is now M1 https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


kushalkolar:
It's trying to throw a WGPU error with `test_positions_graphic_vertex_colors()`, will check later

kushalkolar:
superseeded by #707

PR #662, opened 2024-10-18
Cursor tool

### LINKED ISSUE #380: Hiding and using our own mouse cursors
It seems like hiding the cursor is going to be GUI framework specific, i.e. different implementation for Qt, glfw, jupyter_rfb. If we can remove the system cursor, we can use things like our own crosshairs and pointers that are custom synchronized across subplots.

For Qt it's quite simple, I can't seem to get the `BlankCursor` working though, but `CrossCursor` works:

```python
from PyQt6.QtCore import Qt

iw.gridplot.canvas.setCursor(Qt.CursorShape.BlankCursor)
iw.gridplot.canvas.setCursor(Qt.CursorShape.CrossCursor)
```

kushalkolar:
closes #380

WIP

https://github.com/user-attachments/assets/1bcf3ffc-0d7c-4885-ad30-1b8f1f4081d3

- [ ] Allow adding a `Cursor` to:
  - [ ] subplots 
  - [x] `Graphics`
- [x] The cursor's current "source" position is where the pointer currently is. Either the subplot or the graphic the pointer is on.
- [x] When added to graphics, the cursor's current "source" position is based on the graphic the mouse pointer is currently on. When added to a graphic the cursor's data is always in the same space as the parent graphic's data, i.e. `Cursor.data` maps directly to `Graphic.data`. The offset of an individual cursor matches the offset of the graphic is is on top of. 

- [ ] Allow setting transforms from the source graphic to all over children, just do linear transform, I think this will work:

```python
# w_source, h_source and width and height of source graphic bbox
# w_target, h_target are width and height of target graphic bbox

A = np.array([
    [w_target / w_source, 0],
    [0, h_target / h_source],
])

u = [source_x, source_y]

target_pos = A @ u
```

- [ ] Allowing an arbitrary non-linear transform could also be useful for some visualizations, so allow setting an arbitrary function as a transform. If the Cursor has a permanent set parent graphic, an inverse transform could also be useful and settable!

- [x] Allow a cursor to be a scatter point of any shape
- [ ] Line cursors one line, or two lines. If a line, then the endpoints are based on the bbox of the graphic or the subplot's extents.

- [ ] When calling target graphics, know what the source graphic was to prevent the source graphic from also becoming a target during that "event call cycle". Could happen if bbox of graphics overlap.

- [ ] When a graphic is deleted, remove it from the cursor list.

- [ ] Have modes, for example to allow double click to "add a marker" that sticks in that location.

- [ ] Let the source graphic either be dynamic (i.e. based on current mouse pointer position), or fixed so it only responds to a fixed source graphic.

- [ ] Display text tooltips

### DIFF
--- a/fastplotlib/graphics/_features/_common.py
+++ b/fastplotlib/graphics/_features/_common.py
@@ -18,8 +18,8 @@ class Name(GraphicFeature):
         if not isinstance(value, str):
             raise TypeError("`Graphic` name must be of type <str>")
 
-        if graphic._plot_area is not None:
-            graphic._plot_area._check_graphic_name_exists(value)
+        if graphic._fpl_plot_area is not None:
+            graphic._fpl_plot_area._check_graphic_name_exists(value)
 
         self._value = value
 


kushalkolar:
@clewis7 I think this can be greatly simplified after #744

kushalkolar:
superseeded by #947

PR #663, opened 2024-10-18
flat graphic features

### LINKED ISSUE #629: Flat architecture for all graphic features
An idea for simplifying and solidifying the graphic features API. All graphic features should be flat and have no user-changeable or user accessible attributes, except for `.value` for features that subclass `BufferManager`. 

Reasoning: currently `VertexCmap` allows being set like this:

```python3
line.cmap = "viridis"`
line.cmap.transform = array_1d
line.cmap.alpha = 0.5
```

I think this is confusing since it is inconsistent compared to all other graphic features, and it can also make events confusing. For example, what if you want an event for the transform or alpha but no the cmap? 

Specific to cmaps for `PositionsGraphics` we can have an independent `cmap_transform` feature that only manages the cmap transform and doesn't touch the cmap or alpha. If an alpha is set on the graphic then it will read the alpha value set on the graphic to set that dimension in the final RGBA array to set the colors buffer. Similarly, if a user decides to set `VertexColors` it will read the alpha feature of that graphic. Perhaps we can allow `alpha` to be set to `None` or a float between [0.0, 1.0], can decide this detail later.

Anyways, basic idea is that from our experience I think we can make a formal structure that all `GraphicFeatures` should follow. Pretty much all of them already follow this except for `VertexCmap`:

1. `GraphicFeatures` should not have any user-end attributes or methods. Except for `BufferManagers` which have `.value`.
2. They must only have one type of event which corresponds to their single settable attribute value (ex. cmap, cmap_transform, alpha, hickness, etc.) and nothing else. 
3. A `GraphicFeature` is allowed to access the other features of a graphic when necessary, for example `VertexCmap` already does this since it's basically a wrapper for `VertexColors`. But we can go further with this by for example adding `VertexAlpha` to set the color alpha of a graphic. I think this makes it really easy to set things on the user-end in an independent way, and to also get precise events.

Implemeting this is trivial, I just have to split up `VertexCmap`. And we can have this is dev docs somewhere that this is the structure of graphic features.

kushalkolar:
closes #629 

Final tweak to graphic features, last breaking change that I foresee to this part of the API :smile: 

WIP

### DIFF
--- a/fastplotlib/graphics/_collection_base.py
+++ b/fastplotlib/graphics/_collection_base.py
@@ -1,9 +1,9 @@
-from contextlib import suppress
 from typing import Any
 
 import numpy as np
 
 from ._base import Graphic
+from ._features import GraphicFeature
 
 
 class CollectionProperties:
@@ -249,6 +249,7 @@ class GraphicCollection(Graphic, CollectionProperties):
     def add_event_handler(self, *args):
         """
         Register an event handler.
+        Just adds the event handler to each individual graphic in the collection.
 
         Parameters
         ----------
--- a/fastplotlib/graphics/_features/__init__.py
+++ b/fastplotlib/graphics/_features/__init__.py
@@ -1,11 +1,14 @@
 from ._positions_graphics import (
     VertexColors,
     UniformColor,
+    UniformAlpha,
     UniformSize,
     Thickness,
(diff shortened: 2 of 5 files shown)


kushalkolar:
superseeded by #990 1072

PR #676, opened 2024-11-16
start video points example

kushalkolar:
WIP

image


kushalkolar:
this is all ndwidget stuff now

PR #689, opened 2025-01-07
add `size_space` to `pygfx.PointsMaterial` scatter kwargs

FlynnOConnell:
The `size_space` property of [PointsMaterial](https://docs.pygfx.org/stable/_autosummary/materials/pygfx.materials.PointsMaterial.html#pygfx.materials.PointsMaterial) controls which coordinate space to use for the sizing of each point. Default is `screen`, which using logical pixel values mapped to the screen. Zooming into a plot causes the thickness of circles to remain the same thickness. [see here](https://github.com/MillerBrainObservatory/LBM-CaImAn-Python/pull/57#issuecomment-2574232113)

This aims to allow the other options of [CoordSpace Enum](https://docs.pygfx.org/stable/_autosummary/materials/pygfx.materials.PointsMaterial.html#pygfx.materials.PointsMaterial) to be used instead (world, model, screen).


FlynnOConnell:
I'm not immediately seeing the need for a `GraphicFeature` here, but I could be wrong?

kushalkolar:
All aspects of a graphic are managed by a `GraphicFeature` class and exposed to the user as properties: https://www.fastplotlib.org/ver/dev/developer_notes/graphics.html#graphic-properties 

Lines also have a [`thickness_space`](https://docs.pygfx.org/stable/_autosummary/materials/pygfx.materials.LineMaterial.html#pygfx.materials.LineMaterial) parameter so this feature can actually be settable with a property of the `PositionsGraphic` base class.

kushalkolar (changes requested):
A good start!

This doesn't need to be specific to scatters, pygfx line material also uses a `thickness_space` property that takes the same args and does an analogous thing so this can be in the `PositionsGraphic` base class.

fastplotlib/graphics/scatter.py
```diff
@@ -97,13 +103,22 @@ def __init__(
...
         self._set_world_object(world_object)
 
+    @property
+    def size_space(self) -> str:
+        return self._size_space.value
+
+    @size_space.setter
+    def size_space(self, value: str):
+        self._size_space.set_value(self, value)
+
```
kushalkolar, line 121: this can all just be in the base class

fastplotlib/graphics/scatter.py
```diff
@@ -73,13 +77,15 @@ def __init__(
...
             isolated_buffer=isolated_buffer,
+            size_space=size_space,
             **kwargs,
         )
 
         n_datapoints = self.data.value.shape[0]
 
         geo_kwargs = {"positions": self._data.buffer}
         material_kwargs = {"pick_write": True}
+        self._size_space = SizeSpace(size_space)
```
kushalkolar, line 88: create this in the base class

fastplotlib/graphics/line.py
```diff
@@ -96,6 +106,14 @@ def __init__(
...
         self._set_world_object(world_object)
 
+    @property
+    def size_space(self) -> str:
+        return self._size_space.value
+
+    @size_space.setter
+    def size_space(self, value: str):
+        self._size_space.set_value(self, value)
+
```
kushalkolar, line 116: only required in base class

fastplotlib/graphics/line.py
```diff
@@ -66,10 +70,12 @@ def __init__(
             cmap=cmap,
             cmap_transform=cmap_transform,
             isolated_buffer=isolated_buffer,
+            size_space=size_space,
             **kwargs,
         )
 
         self._thickness = Thickness(thickness)
+        self._size_space = SizeSpace(size_space)
```
kushalkolar, line 78: create in base class

kushalkolar (changes requested):
Will also need a simple test when you're done, see this for an outline:

https://github.com/fastplotlib/fastplotlib/blob/f204f554fa99ba9c706af2f969c6988b74ae2bcd/tests/test_positions_graphics.py#L419-L445

fastplotlib/graphics/_features/_positions_graphics.py
```diff
@@ -182,6 +182,24 @@ def set_value(self, graphic, value: float | int):
...
+    def __init__(self, value: str):
+        self._value = str(value)
+        super().__init__()
+
+    @property
+    def value(self) -> str:
+        return self._value
+
+    def set_value(self, graphic, value: str):
+        graphic.world_object.material.size_space = str(value)
```
kushalkolar, line 196: `thickness_space` if the material is for a line, can have a conditional like `if "Line" in material.__class__.__name__`

kushalkolar: also don't try to coerce `value` to a string, user should give a string

fastplotlib/graphics/_positions_base.py
```diff
@@ -54,6 +55,19 @@ def cmap(self, name: str):
...
         self._cmap[:] = name
 
+    @property
+    def size_space(self):
+        """
+        The coordinate space in which the size is expressed (‘screen’, ‘world’, ‘model’)
+        
+        See https://docs.pygfx.org/stable/_autosummary/utils/utils/enums/pygfx.utils.enums.CoordSpace.html#pygfx.utils.enums.CoordSpace for available options.
+        """
+        return self._size_space
```
kushalkolar, line 65: this will return the `GraphicFeature` instance to the user so you must use `.value` to return the `str` value to the user

fastplotlib/graphics/_positions_base.py
```diff
@@ -54,6 +55,19 @@ def cmap(self, name: str):
...
+
+    @size_space.setter
+    def size_space(self, value: str):
+        self._size_space = value
```
kushalkolar, line 69: use `set_value` method

kushalkolar (changes requested):

fastplotlib/graphics/_positions_base.py
FlynnOConnell: [8a0b9be](https://github.com/fastplotlib/fastplotlib/pull/689/commits/8a0b9be3b9ba9ee778c7827186438faaa3876f0e)

fastplotlib/graphics/_positions_base.py
FlynnOConnell: [8a0b9be](https://github.com/fastplotlib/fastplotlib/pull/689/commits/8a0b9be3b9ba9ee778c7827186438faaa3876f0e)

FlynnOConnell:
Eye test looks like everything checks out. will work on a test case next

tests/test_positions_graphics.py
```diff
@@ -443,3 +443,26 @@ def test_thickness(thickness):
...
+
+    if size_space is None:
+        size_space = "screen"  # default space
+
+    if graphic_type == "line":
+        graphic = fig[0, 0].add_line(data=data, **kwargs)
+        assert graphic.world_object.material.thickness_space == size_space
+    elif graphic_type == "scatter":
+        graphic = fig[0, 0].add_scatter(data=data, **kwargs)
+        assert graphic.world_object.material.size_space == size_space
```
kushalkolar, line 468: also need:

```python
assert graphic.size_space == size_space
```

and test changes works
```python
graphic.size_space = "world"
assert graphic.size_space == "world"
assert graphic.material.... == "world"
```

kushalkolar (changes requested):

tests/test_positions_graphics.py
```diff
@@ -460,9 +460,27 @@ def test_size_space(graphic_type, size_space):
...
+    # size_space is really an alias for pygfx.utils.enums.CoordSpace
     if graphic_type == "line":
         graphic = fig[0, 0].add_line(data=data, **kwargs)
+
+        # test getter
         assert graphic.world_object.material.thickness_space == size_space
+        assert graphic.size_space == size_space
+
+        # test setter
+        graphic.size_space = 'world'
```
kushalkolar, line 472: double quotes, black formatting

kushalkolar (changes requested):

tests/test_positions_graphics.py
FlynnOConnell: this is my worst habit

kushalkolar:
Since you've added/changed the API you will need to regenerate the API docs:

in `docs/source`:

delete the `api` dir

with the env activated, run:

`python generate_api.py`

add and commit `docs/api`

kushalkolar:
black is still unhappy, just do `black fastplotlib/`

only black format the `fastplotlib` source dir, not the examples or tests etc.

FlynnOConnell:
plz tell black to stop complaining 

image

FlynnOConnell:
@kushalkolar I don't even have a guess for why the docs failed

`Run SamKirkland/FTP-Deploy-Action@v4.3.5 Error: Error: Input required and not supplied: server`

kushalkolar:
I think it's trying to access the credentials which are repo secrets and it won't since you're on a fork, we'll figure out later how to make docs upload from forks. Anyways the docs built without issues so that's fine and everything else LGTM.

PR #690, opened 2025-01-14
fix `ImageWidget.reset_vmin_vmax()`

### LINKED ISSUE #673: weird histogram LUT tool behavior in imagewidget
Doesn't set the vmin vmax correctly based on the entire data array, it sets the histogram limits based on the entire data array but then sets the vmin vmax only based on the current frame but it's supposed to set the vmin vmax based on the entire array instead.

https://github.com/user-attachments/assets/adf336d6-8a7c-4d75-96d4-8362d2664d4d

I think the histogram tool needs a general refactor anyways.

kushalkolar:
fixes #673

This has really been annoying me lately :laughing:


kushalkolar:
@clewis7 ready for review!

clewis7 (approved):
LGTM

PR #691, opened 2025-01-16
small wgpu changes

kushalkolar:
- fix deprecated wgpu calls
- show a warning instead of raising an exception upon import if no wgpu adapters are found. I think it's probably not the best behavior to raise exception on import for something like this, a warning is more appropriate


kushalkolar:
@clewis7 ready for review

clewis7 (approved):

PR #695, opened 2025-01-17
Fix imgui popup menus

### LINKED ISSUE #694: right click menu is broken
Getting WGPU draw errors. Not sure when this happened, we should add a test for this, should be easy.

histogram colorbar right click is also giving the same draw errors

kushalkolar:
fixes #694 

The kwargs for the imgui popup menu for right clicks got more strict in the v1.6.0 release. Was annoying to troubleshoot since I was first getting errors stating that the previous imgui frame had not been ended and to call `imgui.end_frame()`. So then I added an extra `imgui.end_frame()` within the right click menu and it produced the actual error which was that imgui wasn't happy with the types of the kwargs :man_shrugging: 

So before we had things like this for example:

```python
imgui.menu_item("item name", None, val > 0)

# signature
menu_item(str, str, bool)
```
args are: (item name, keyboard shortcut, selected)

So before it worked if we gave `None` for the second arg which is keyboard shortcut, now it has to explicitly be `""`. And it's not taking expressions that results in a `bool` for whatever reason anymore.

So in summary the above needs to become the following for imgui v1.6.0 to be happy:

```python
menu_item("item name", "", bool(val > 0))
```

### DIFF
--- a/fastplotlib/ui/_base.py
+++ b/fastplotlib/ui/_base.py
@@ -248,8 +248,6 @@ class Popup(BaseGUI):
         self._figure = figure
         self._fa_icons = self._figure._fa_icons
 
-        self._event_filter_names = set()
-
         self.is_open = False
 
     def open(self, pos: tuple[int, int], *args, **kwargs):
--- a/fastplotlib/ui/right_click_menus/_colormap_picker.py
+++ b/fastplotlib/ui/right_click_menus/_colormap_picker.py
@@ -153,7 +153,7 @@ class ColormapPicker(Popup):
                 * imgui.get_font().font_size
             ) - 2
 
-            if imgui.menu_item("Reset vmin-vmax", None, False)[0]:
+            if imgui.menu_item("Reset vmin-vmax", "", False)[0]:
                 self._lut_tool.image_graphic.reset_vmin_vmax()
 
             # add all the cmap options
--- a/fastplotlib/ui/right_click_menus/_standard_menu.py
+++ b/fastplotlib/ui/right_click_menus/_standard_menu.py
@@ -82,14 +82,14 @@ class StandardRightClickMenu(Popup):
             imgui.separator()
 
             # autoscale, center, maintain aspect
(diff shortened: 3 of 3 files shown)


kushalkolar:
@clewis7 ready for review!

kushalkolar:
oh wait let me add a test

clewis7:
I'll wait for test before doing a review :D

kushalkolar:
so there is `canvas.submit_event()` which we can use to offscreen test the right click menu :D 

https://github.com/fastplotlib/fastplotlib/issues/705#issuecomment-2611888070

kushalkolar:
> so there is `canvas.submit_event()` which we can use to offscreen test the right click menu :D
> 
> [#705 (comment)](https://github.com/fastplotlib/fastplotlib/issues/705#issuecomment-2611888070)

I just tried this, works for qt canvas but not offscreen :man_shrugging: . Documenting here for now, let's just merge this and figure it out later.

```python
"""
Test right click menu
=====================

Screenshot test for the right click menu
"""

# test_example = true
# sphinx_gallery_pygfx_docs = 'hidden'

import fastplotlib as fpl
import numpy as np

figure = fpl.Figure(size=(700, 560))

# for some reason right click menu doesn't show if the scene is completely empty
figure[0, 0].add_image(np.zeros((2, 2)), cmap="gray")


figure[0, 0].axes.visible = False

figure.show()


data = {"x": 10, "y": 10, "modifiers": (), "n_touches": 0, "touches": {}}

move = {
    "event_type": "pointer_move",
    "button": 0,
    "buttons": (),
    **data,
}
down = {
    "event_type": "pointer_down",
    "button": 2,
    "buttons": (2,),
    **data,
}
up = {
    "event_type": "pointer_up",
    "button": 2,
    "buttons": (),
    **data,
}


figure.canvas.submit_event(move)
figure.canvas.submit_event(down)
figure.canvas.submit_event(up)


if __name__ == "__main__":
    print(__doc__)
    fpl.loop.run()
```

clewis7 (approved):

PR #696, opened 2025-01-17
remove glfw import that leaked in

kushalkolar:
IDK where this leaked in from

I was getting wayland errors so I was wondering why wayland doesn't like image widgets specifically and then I realized the glfw import :laughing:


kushalkolar:
@clewis7 gtg!

clewis7 (approved):

PR #698, opened 2025-01-17
simplify CI and install extras

### LINKED ISSUE #678: Add test screenshots without imgui
Let's do this before next release. Will doubles the number of test screenshots but it's better than  checking manually.

And will need to check that importing ImageWidget in this condition raises an import error.

kushalkolar:
A _successfull_ attempt to do the following: :smile: 

- [x] extras deps in setup.py are simplified, was kinda entangled before
- [x] use matrix in CI for notebook vs. no notebook installs
- [x] introduce non-imgui tests in the matrix to make sure fpl works without imgui

Also updated to use uplaod-artifacts v4, but for reasons I do not understand it merges the imgui and non-imgui images within the same uploaded zip file even though I specify two separate filenames [as specified by their docs](https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact). One of the artifacts contains both the imgui and non-imgui screnshots (wtf?) and the other one has only the imgui versions which makes no sense (wtf 2) :man_shrugging: . Anyways this is fine for us for now so I don't think we should hold back for this minor quirk of upload-artifacts (they might fix it in a future version sooner than us figuring it out)

closes #678


kushalkolar:
oh shit this is actually working :exploding_head:

kushalkolar:
wow after lots of git actions bash wrangling and sed garbling my idea actually worked :astonished: !

@clewis7 this will require a more thorough review XD

.github/workflows/ci.yml
```diff
@@ -60,69 +61,22 @@ jobs:
...
-        pip install git+https://github.com/pygfx/pygfx.git@main
-        pip install -e ".["tests-desktop"]"
-    - name: Show wgpu backend
-      run:
-        python -c "from examples.tests.testutils import wgpu_backend; print(wgpu_backend)"
-    - name: fetch git lfs files
-      run: |
-        git lfs fetch --all
-        git lfs pull
-    - name: Test examples
```
kushalkolar, line 114: git is stupid, this entire block is actually the previous non-notebooks test pipeline that came after afterwards

.github/workflows/ci.yml
```diff
@@ -15,20 +15,16 @@ on:
 
 jobs:
   test-build-full:
-    name: Test Linux, notebook + offscreen
+    name: Test Linux
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 10
```
kushalkolar, line 20: we can go back to 10 mins now that we don't have python3.10 which required more time due to no imgui wheels existing for it.

.github/workflows/ci.yml
```diff
@@ -47,7 +43,12 @@ jobs:
         # remove pygfx from install_requires, we install using pygfx@main
         sed -i "/pygfx/d" ./setup.py
         pip install git+https://github.com/pygfx/pygfx.git@main
-        pip install -e ".["tests"]"
+    - name: Install fastplotlib
+      run: |
+        # create string with one of: tests,imgui,notebook; test,imgui; test,notebook ; tests
+        # sed removes trailing comma
+        # install fastplotlib with given extras options from above
+        pip install -e ".[$(echo "tests,${{ matrix.imgui_dep }},${{ matrix.notebook_dep }}" | sed -e "s/,\+/,/g" -e "s/,$//")]"
```
kushalkolar, line 51: I copy pasted the sed from stackoverflow

kushalkolar:
Oh wait I can use bash to exclude the ImageWidget nb for the pytest call when imgui doesn't exists

setup.py
```diff
@@ -16,45 +16,28 @@
...
-        "sidecar",
         "imageio[ffmpeg]",
         "matplotlib",
         "scikit-learn",
-        "imgui-bundle",
     ],
     "notebook": [
         "jupyterlab",
-        "jupyter-rfb>=0.4.1",
+        "jupyter-rfb>=0.5.1",
```
clewis7, line 29: does this solve the rehandle issue?

clewis7 (approved):
LGTM, very cool stuff! I like the pattern matching with sed

setup.py
kushalkolar: Yup see jupyter_rfb repo

kushalkolar:
> LGTM, very cool stuff! I like the pattern matching with sed 

I wouldn't say I like the sed, it's an ugly garble but useful 😂

PR #700, opened 2025-01-21
update screenshot

kushalkolar:
my hunch is this is due to https://github.com/pygfx/pygfx/pull/945 

when `set_data()` is used, it wouldn't immediately update the camera related stuff which is why the axes don't appear on the initial draw after `set_data()`, as a user we don't notice it because that draw only lasts for a few milliseconds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
   test-build-full:
     name: Test Linux
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 25
     if: ${{ !github.event.pull_request.draft }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


kushalkolar:
@clewis7 this should fix the current CI failure

kushalkolar:
I also changed the CI timeout from 10 mins to 15, for whatever reason it took llvmpipe etc. 7 mins to install. Maybe something changed on github actions :man_shrugging: . Anyways jobs usually finish in 5 mins, maybe stuff gets cached so if it hits a new machine or something it takes longer.

kushalkolar:
idk what's taking so long with apt install on github actions lately, relaxed the timeout time to 25 mins :man_shrugging:

kushalkolar:
wtf, so now all the jobs finished within 5mins?? Must be some caching things or apt servers were slow for some reason during those earlier runs

clewis7 (approved):

PR #701, opened 2025-01-21
Update ci step

kushalkolar:
Some small cleanup and organization. Previously it would test the components (i.e. `pytest tests/`) and test examples in the same step, now separated into individual steps. Makes no difference but it makes the github actions logs easier to read and go through when things break.


kushalkolar:
once #700 is merged I can rebase on main and should be gtg then

kushalkolar:
@clewis7 ready for review

clewis7 (approved):

PR #706, opened 2025-01-25
add job that uses release version of pygfx

### LINKED ISSUE #702: Also test using latest `pygfx` on pypi
We should have `pygfx@main` and latest `pygfx` on pypi in our CI matrix. But if the job using `pygfx` from pypi fails we want it to produce some sort of "warning" instead of failing the entire action. Maybe we can do https://github.com/burningmantech/ranger-ims-server/pull/1347 

This will be useful for us to better automate releases, we will know whether our current main branch works with the latest `pygfx` before releasing without us trying them locally in a new env which is what we do now. 

@clewis7 can you take a look and see if there are better ways to do this? 

I think this is the last main thing we should do before the next release!

kushalkolar:
closes #702 

see this comment https://github.com/fastplotlib/fastplotlib/pull/706#issuecomment-2613724223

~Instead of adding it to the matrix I just made a new identical job so we can use the branch protection rules instead of the more complicated check and message system linked in that issue~

~Will see if this works, I've set "Test Linux" (which uses `pygfx@main`) as a required status check that must pass for a PR to be merged and I wonder if that will cover the entire matrix for that job. "Test Linux pygfx release" is however not required to pass.~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,70 @@ jobs:
           examples/diffs
           examples/notebooks/diffs
 
+jobs:
+  test-using-pygfx-release:
+    name: Test Linux pygfx release
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    if: ${{ !github.event.pull_request.draft }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["3.11", "3.12", "3.13"]
+        imgui_dep: ["imgui", ""]
+        notebook_dep: ["notebook", ""]
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        lfs: true
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install llvmpipe and lavapipe for offscreen canvas
+      run: |
+        sudo apt-get update -y -qq
+        sudo apt-get install --no-install-recommends -y ffmpeg libegl1-mesa-dev libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers xorg-dev
+    - name: Install fastplotlib
+      run: |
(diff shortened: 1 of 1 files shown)


kushalkolar:
`ImageWidget.set_data()` screenshot test is failing due to https://github.com/fastplotlib/fastplotlib/pull/700 , this isn't a release blocker.

kushalkolar:
ok this was getting too complicated, let's just add pygfx-release/main to the matrix and during PRs we just manually look at the job statuses and determine if we want to merge or not based on any pygfx-release failures. This is only really necessary until pygfx stabilizes so we don't want to spend too much effort making a complicated thing that's only for this temporary phase of developement

kushalkolar:
@clewis7 ready for review!

clewis7:
> ok this was getting too complicated, let's just add pygfx-release/main to the matrix and during PRs we just manually look at the job statuses and determine if we want to merge or not based on any pygfx-release failures. This is only really necessary until pygfx stabilizes so we don't want to spend too much effort making a complicated thing that's only for this temporary phase of developement
> 
> @clewis7 what do you think?

But will having the CI badge as failing look bad? I feel like that's something I look at when I look at a repo :flushed:

clewis7 (approved):

kushalkolar:
> But will having the CI badge as failing look bad? I feel like that's something I look at when I look at a repo 😳

True, we can remove the CI badge for now?

clewis7:
> > But will having the CI badge as failing look bad? I feel like that's something I look at when I look at a repo 😳
> 
> True, we can remove the CI badge for now?

But does that also look bad? I can't decide what the best optics are lol

kushalkolar:
> > > But will having the CI badge as failing look bad? I feel like that's something I look at when I look at a repo 😳
> > 
> > 
> > True, we can remove the CI badge for now?
> 
> But does that also look bad? I can't decide what the best optics are lol

pygfx seems like it'll stabilize over the coming months, it's fine for now I think.

PR #707, opened 2025-01-25
add macos-latest to CI matrix

kushalkolar:
let's see if this works

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
     - name: Install llvmpipe and lavapipe for offscreen canvas
+      if: ${{ matrix.os == "ubuntu-latest" }}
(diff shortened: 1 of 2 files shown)


kushalkolar:
mac test failing because there is a slight diff for the lines-data test, will increase tolerance

no-imgui-nb-diff-lines-data-rgb

kushalkolar:
ok so basically everything that isn't pygfx-latest is passing! :partying_face:

kushalkolar:
This mean we'll have over 50 tests with all the variants, but once pygfx becomes more stable it'll go down by half which seems normal for projects of this scope (I looked at matplotlib etc.)

clewis7 (approved):

kushalkolar:
@almarklein thought you might be interested to know that tests now work with the github mac runners :smile:

almarklein:
O, that's really cool! 🚀

Issue #711, opened 2025-01-27
ImageWidget computation issue


apasarkar:
There is some code [here](https://github.com/fastplotlib/fastplotlib/blob/d4af1a900a8d7c89bcc25601badf1bf1a9a92ff2/fastplotlib/utils/functions.py#L292) that uses the "size" attribute. ImageWidget uses this code to do some data sampling to generate a histogram/colorbar for each image stack being displayed. Previously, imageWidget did not require arrays to implement size; instead shape was sufficient. 

We should change data.size to np.prod(data.shape) to get the total number of elements.

kushalkolar:
done in #712

PR #712, opened 2025-01-27
removes size attribute use for histogram component of imagewidget

apasarkar:
Addresses issue #711, verified that .size is no longer a "required" property for lazy arrays used in ImageWidget.


kushalkolar:
LGTM but tests are failing due to this: https://github.com/fastplotlib/fastplotlib/pull/713#issuecomment-2635650155 

will try to figure out in a few days

kushalkolar (approved):

PR #718, opened 2025-02-05
add differential eq phase portrait example

kushalkolar:
WIP, will add interactive one later which allows changing the variables with sliders

image


kushalkolar:
closing in favor of #1085

PR #713, opened 2025-01-28
bump version to 0.4.0

kushalkolar:

PR #726, opened 2025-02-13
figure out what is wrong with the texture clamp on github actions ubuntu runners

kushalkolar:
figuring out where [this](https://github.com/fastplotlib/fastplotlib/pull/713#issuecomment-2635650155) is coming from, trying older pygfx commits in case it's from there. I cannot reproduce this locally on any of my computers.


kushalkolar:
ok so it's not in this commit: https://github.com/pygfx/pygfx/commit/396186a861b0b161c4dc0afcb7719f089be50abb

I had a hunch this was the cause: https://github.com/pygfx/pygfx/pull/957

kushalkolar:
yup the gamma PR introduced this issue, weird that it's only on lavapipe

PR #730, opened 2025-02-17
CNN weights example

clewis7:
Started playing around with a simple CNN in pytorch to visualize the weights as the model is training

The weights are definitely being updated every epoch, but the plot will not update until the last training epoch is done...not sure why

Another thing of note, pytorch tensors cannot directly be used with fpl right now because 

```
File [~/repos/fastplotlib/fastplotlib/graphics/_features/_image.py:101](http://localhost:8888/lab/tree/examples/fastplotlib/graphics/_features/_image.py#line=100), in TextureArray._fix_data(self, data)
     95     raise ValueError(
     96         "image data must be 2D with or without an RGB(A) dimension, i.e. "
     97         "it must be of shape [rows, cols], [rows, cols, 3] or [rows, cols, 4]"
     98     )
    100 # let's just cast to float32 always
--> 101 return data.astype(np.float32)

AttributeError: 'Tensor' object has no attribute 'astype'
```

### DIFF
--- a/examples/machine_learning/neural_net.ipynb
+++ b/examples/machine_learning/neural_net.ipynb
@@ -1,0 +1,647 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "8a7e6b7d-5d6e-4ffc-ae58-6dd8857af672",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.\n",
+      "Detected skylake derivative running on mesa i915. Clears to srgb textures will use manual shader clears.\n",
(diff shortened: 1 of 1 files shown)


kushalkolar:
Are you trying to update a graphic in a for loop? You can't do this right now, see the discussions post. For now we can use zmq or see rendercanvas and try to expose the ondemand rendering mode in fpl

Torch tensors can't be used directly but if you do np.asarray it will not make a copy

clewis7:
>Torch tensors can't be used directly but if you do np.asarray it will not make a copy

Even when I do np.asarray I still get issues that the arrays are on different devices

```
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

kushalkolar:
Yea you need to send to cpu

clewis7:
Okay, still a bit messy and inelegant...but progress

PR #738, opened 2025-02-24
helper method to add text to the corner or center of a graphic

kushalkolar:
very simple, API subject to change later

image

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -361,6 +361,72 @@ class Graphic:
                 feature = getattr(self, f"_{t}")
                 feature.remove_event_handler(wrapper)
 
+    def add_text(self, text: str, location: str = "center", location_z: str = "front", **kwargs):
+        """
+        Add a TextGraphic at one of the corners or center of this graphic.
+
+        Note 1:
+        Note 2: Uses axis aligned bounding box to get corners, does not account for any rotation set on the graphic.
+
+        Parameters
+        ----------
+        text: str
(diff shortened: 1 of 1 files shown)


clewis7:
I think this could be super useful. Even in the kmeans example, I was thinking about how nice it would be to have a label for each centroid point. Although, that would be giving a label to each point in a scatter, as opposed to a single label for a graphic. Should we support both?

The only potential issue that I see is if the graphics were super close to each other and the labels overlapped. How to reconcile potential readability issues. I guess making it easy to change what corner or position on the graphic the text is would be easiest.

kushalkolar:
> I think this could be super useful. Even in the kmeans example, I was thinking about how nice it would be to have a label for each centroid point. Although, that would be giving a label to each point in a scatter, as opposed to a single label for a graphic. Should we support both?

In that case a user can easily create `TextGraphic` with the offset set near the scatter point. Maybe for scatters we override `add_text()` and have the option to place it w.r.t. a bbox corner or w.r.t. certain scatter indices? 

> The only potential issue that I see is if the graphics were super close to each other and the labels overlapped. How to reconcile potential readability issues. I guess making it easy to change what corner or position on the graphic the text is would be easiest.

This can come later.

kushalkolar:
Let's wait for this PR which adds lots of MultiText functionality! https://github.com/pygfx/pygfx/pull/834

@clewis7 for example: https://pygfx--834.org.readthedocs.build/834/_gallery/feature_demo/text_multiple_labels.html#sphx-glr-gallery-feature-demo-text-multiple-labels-py

PR #740, opened 2025-02-27
Rectangle/bbox layouts in a Figure

### LINKED ISSUE #315: Customizable Figure layouts
One easy thing we can start with is allowing subplots to be created with bounding boxes that described 0-100 normalized positions and widths. We can also think about more "automated" ways to define the subplots, because it can get quite tedious to write out bboxes for every subplot. The matplotlib gridspec API is kind of tedioius, we can look into other libs for inspiration as well. https://matplotlib.org/stable/gallery/lines_bars_and_markers/scatter_hist.html#defining-the-axes-positions-using-a-gridspec.

```python3
subplot1 = [bbox]
subplot2 = [bbox2]
...
gridplot = GridPlot(subplots=[subplot1, subplot2])
```

image

@clewis7 @ArjunPutcha

### LINKED ISSUE #616: imgui subplot toolbar should go below subplot "bottom" dock

kushalkolar:
Flexible rectangle/bbox layouts with resizable and draggable subplots!

Summary of this PR:

1. Adds `layouts._engine`, which is a module for layout engines.
    - `FlexLayout` is a flexible layout
    - `GridLayout` is a subclass of `FlexLayout` using fixed extents.
2. Adds a `Frame` class. This contains the `plane` Mesh, `resize_handle` Point object, and the subplot title `TextGraphic`. It also manages the `Subplot` viewport rect and the dock viewport rects.
3. Adds a `RectManager` which manages the `Frame` rect, makes it easy to manage rect, fractional rects, extent, etc. in one class. This allows `Frame` to use a `rect` or `extent` in absolute screen space or as fraction of the canvas.
4. A Layout engines manages the subplot frames.

Users can specify rects or extents, in absolute pixels or as fractions of the canvas. 

Future PRs will allow adding/removing a subplot and insets.

closes #315 closes #616


kushalkolar:
One possible way to do automatic rectangle packing: https://github.com/laserson/squarify 

@almarklein have you done something like this before? Allow bboxes for subplot viewport rects to be defined, and then when a viewport rect is resized all the other rects should adjust accordinlgy.

almarklein:
One observation is that sometimes you want to position something in absolute pixels, and sometimes as a fraction of the parent's space. In [visvis](https://github.com/pygfx/visvis2/blob/main/visvis2/_view.py) I therefore positioned widget-objects like this (taken from the visvis docs):

> Each element (x,y,w,h) can be either: 1) The integer amount of pixels relative to the parent's position; 2) The fractional amount (float value between 0.0 and 1.0) of the parent's width or height. Each value can be negative. For x and y this simply means a negative offset from the parent's left and top. For the width and height the difference from the parent's full width/height is taken.

But this is a bit weird/confusing. Plus very often you want a relative position with some offset.

So in visvis2 (wip) I borrowed an approach from CSS, each item in the rect is a string e.g. "120px", "50%" or "50% - 20px". See https://github.com/pygfx/visvis2/blob/3e3e9d785c34d3170433c78595390f2304185df3/visvis2/_view.py#L33-L39 I've implemented something similar in https://github.com/pygfx/pygfx/pull/954 (not sure if that will stay).

almarklein:
If you want more elaborate layout you should probably implement container objects that position their children in a specific way, similar to CSS flexbox and the like.

kushalkolar:
Thanks! I think we'll just have a free-form drag & resize subplots on the canvas and block overlaps, and have another mode that uses a simple tiling algorithm maybe from squarify. 

This answer on stackoverflow is very good: https://stackoverflow.com/a/306332 and this demo :smile: https://silentmatt.com/rectangle-intersection/

kushalkolar:
Idea:

Instead of putting all the layout management logic inside the Figure class we keep it as separate classes. `Figure.layout_manager` is an instance of a layout manager class. Right now we can have 3 different layout managers, `Grid`, freeform `Flex` (the implementation in the video above), and `Squarify` for the treemap once I figure out how to implement that. 

Merge the implementation of the subplot Frame stuff into the `Subplot` class, I think this makes sense since it will then give all subplots the underlay plane, title, and resize handler. The `rect` and `extent` properties can be made read-only since we don't want users to manually change a subplot rect by directly addresses the subplot. Users can "request" changing a subplot rect via `Figure.layout_manager` if the type of `LayoutManager` supports it. For example, do not allow users to change the rect if the layout manager is a `Grid` since the rects are done automatically, but if it is a `Flex` layout we allow the user to request changing the `rect`. The request to change the rect is "denied" if the user passes in an invalid rect or a rect that would overlap with another subplot or cause other issues.

Basically to summarize:

We make:

`Figure.layout_manager` which is an instance of `GridLayout`, `FlexLayout`, or `SquarifyLayout`.

The `LayoutManager` manages pointer events which allow subplots to move or resize (if a flex layout). In a `GridLayout` it just auto-sets the subplot rects.

For `FlexLayout` we have `FlexLayout.set_rect(subplot, rect: <x, y, w, h>)` which is a user-facing function that allows them to manually define a rect. This function will raise if the user passes an invalid rect. 
Similarly we have functions `FlexLayout.set_extent(subplot, extent: <x0, x1, y0, y1>)` for extent.

To the current `Subplot` class we add the stuff from `Frame` in the PR as it looks right now, to give it the plane Mesh, resize handler, subplot title text. Properties like `rect`, `extent` are read-only for the user. The `LayoutManager` however can set the subplot rect.

clewis7:
I think layout manager sounds like a nice way to abstract, and then users can specify the layout they want via string when they instantiate a plot 

```python
fig = fpl.Figure(layout="grid")
```

I think by default, we would want to maintain the grid layout we have now.

clewis7:
Also, would squarify be for making rectangular subplots of different sizes and tiling them all together? 

If that's the case, could we not also use squarify for making our grids too? Fixed rectangles that get tiles together in a particular shape. Or maybe I am not understanding what the the squarify layout would be for

kushalkolar:
Yes squarify is for tiling rects of diff sizes. I think it's overkill for a simple grid plot layout though so we should keep them separate.

kushalkolar:
In the flex and squarify layouts we can also have a swap method to swap subplot rects

kushalkolar:
> keep the OS's cursor and stop it from moving.

This is what I meant, I'll do another search and see if there is such a thing. Otherwise once we have our own cursors.

I just realized a grid layout can just be managed by a flex layout with fractional extents, this will make the implementations easier.

kushalkolar:
We can allow one inset plot area within a subplot, arbitrary overlaps would be a mess with the screenspace camera and if a user wants many overlapping plots they should probably make them separate anyways

kushalkolar:
Let's do insets and add/remove subplots and row/columms in another PR.

fastplotlib/layouts/_frame.py
```diff
@@ -0,0 +1,371 @@
...
+
+from ._rect import RectManager
+from ._utils import IMGUI, IMGUI_TOOLBAR_HEIGHT
+from ..utils._types import SelectorColorStates
+from ..graphics import TextGraphic
+
+
+"""
+Each Subplot is framed by a 2D plane mesh, a rectangle.
+The rectangles are viewed using the UnderlayCamera  where (0, 0) is the top left corner.
```
clewis7, line 12: top left or top right? see comment from before, there you say top right

clewis7 (changes requested):

fastplotlib/graphics/_base.py
```diff
@@ -365,7 +365,7 @@ def _fpl_add_plot_area_hook(self, plot_area):
         self._plot_area = plot_area
 
     def __repr__(self):
-        rval = f"{self.__class__.__name__} @ {hex(id(self))}"
+        rval = f"{self.__class__.__name__}"
```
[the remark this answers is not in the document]
kushalkolar: knowing the memory address isn't useful for the average user, they're useful for devs but devs will know how to get them anyways

kushalkolar:
@almarklein would be great to get your thoughts on this PR! :smile:

fastplotlib/layouts/_engine.py
```diff
@@ -0,0 +1,384 @@
...
+    def __init__(
+        self,
+        renderer: pygfx.WgpuRenderer,
+        subplots: np.ndarray[Subplot],
+        canvas_rect: tuple[float, float],
+        moveable: bool,
+        resizeable: bool,
+    ):
+        """
+        Base layout engine, subclass to create a useable layout engine.
```
almarklein, line 39:
```suggestion
        Base layout engine, subclass to create a usable layout engine.
```

fastplotlib/layouts/_engine.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        Set the extent of a Subplot
+
+        Parameters
+        ----------
+        subplot: Subplot
+            the subplot to set the extent of
+
+        extent: (xmin, xmax, ymin, ymax)
+            as absolute pixels or fractional
```
almarklein, line 311: Questions that pop up: When is the value considered fractional, based on the type (float vs int)? Can you mix fractional with absolute values in one rect/extent?  Can values also be negative, i.e. going from the end of the parent rectangle?

fastplotlib/layouts/_engine.py
```diff
@@ -0,0 +1,384 @@
...
+        self,
+        renderer,
+        subplots: np.ndarray[Subplot],
+        canvas_rect: tuple,
+        moveable=True,
+        resizeable=True,
+    ):
+        """
+        Flexible layout engine that allows freely moving and resizing subplots.
+        Subplots are not allowed to overlap.
```
almarklein, line 140: The name `FlexLayout` makes me think immediately about CSS flexbox, which is a common layout method. But this seems to do something else (allowing users to freely move the contents). You could consider a different name to avoid confusion.

fastplotlib/layouts/_engine.py
```diff
@@ -0,0 +1,384 @@
...
+        extent = new_rect.extent
+        # check for overlaps
+        for s in self._subplots:
+            if s is subplot:
+                continue
+
+            if s.frame.rect_manager.overlaps(extent):
+                raise ValueError(f"Given rect: {rect} overlaps with another subplot.")
+
+    def set_extent(self, subplot: Subplot, extent: tuple | list | np.ndarray):
```
almarklein, line 301: Interesting idea to have these two flavors to define a rectangular region! ❤️

clewis7:
with glfw, canvas size not the same

```python
  File "/home/clewis/repos/fastplotlib/examples/flex_layouts/extent_layout.py", line 42, in <module>
    figure = fpl.Figure(
        extents=extents,
        names=["astronaut image", "wikkie image", "astronaut histogram", "wikkie histogram"],
        size=size
    )
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_imgui_figure.py", line 55, in __init__
    super().__init__(
    ~~~~~~~~~~~~~~~~^
        shape=shape,
        ^^^^^^^^^^^^
    ...<9 lines>...
        names=names,
        ^^^^^^^^^^^^
    )
    ^
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_figure.py", line 362, in __init__
    subplot = Subplot(
        parent=self,
    ...<7 lines>...
        resizeable=resizeable,
    )
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_subplot.py", line 89, in __init__
    self._frame = Frame(
                  ~~~~~^
        viewport=self.viewport,
        ^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
        canvas_rect=parent.get_pygfx_render_area(),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_frame.py", line 162, in __init__
    self._rect_manager = RectManager.from_extent(extent, canvas_rect)
                         ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 125, in from_extent
    rect = cls.extent_to_rect(extent, canvas_rect)
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 143, in extent_to_rect
    RectManager.validate_extent(extent, canvas_rect)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 193, in validate_extent
    raise ValueError(
        f"extent: {extent} y-range is beyond the bounds of the canvas: {canvas_extent}"
    )
ValueError: extent: [  0 200 280 560] y-range is beyond the bounds of the canvas: [  0. 640.   0. 480.]
```

fastplotlib/layouts/_engine.py
kushalkolar: For rect, if width and height are both less than or equal to 1 we assume it is fractional:

https://github.com/fastplotlib/fastplotlib/blob/91e0a912423c46cbd12560e10635e2767e98052c/fastplotlib/layouts/_rect.py#L29-L30

A width, height of 1 in absolute pixels isn't useful so I think this assumption is fine.

If both width, heigh > 1 then we assume it's in absolute pixels:

https://github.com/fastplotlib/fastplotlib/blob/91e0a912423c46cbd12560e10635e2767e98052c/fastplotlib/layouts/_rect.py#L32-L33

Similarly for extents, if the x, y of the bottom right corner (the bigger x, y) are <= 1 we assume it's fractional:

https://github.com/fastplotlib/fastplotlib/blob/91e0a912423c46cbd12560e10635e2767e98052c/fastplotlib/layouts/_rect.py#L161-L166

Negatives aren't allowed, it checks that `w, h <= 0` for both rects and extents.

fastplotlib/layouts/_engine.py
kushalkolar: Yes! I find both useful in different cases.

fastplotlib/layouts/_engine.py
kushalkolar: changed to `WindowLayout`

kushalkolar:
> with glfw, canvas size not the same
> 
> ```python
> Traceback (most recent call last):
>   File "/home/clewis/repos/fastplotlib/examples/flex_layouts/extent_layout.py", line 42, in <module>
>     figure = fpl.Figure(
>         extents=extents,
>         names=["astronaut image", "wikkie image", "astronaut histogram", "wikkie histogram"],
>         size=size
>     )
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_imgui_figure.py", line 55, in __init__
>     super().__init__(
>     ~~~~~~~~~~~~~~~~^
>         shape=shape,
>         ^^^^^^^^^^^^
>     ...<9 lines>...
>         names=names,
>         ^^^^^^^^^^^^
>     )
>     ^
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_figure.py", line 362, in __init__
>     subplot = Subplot(
>         parent=self,
>     ...<7 lines>...
>         resizeable=resizeable,
>     )
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_subplot.py", line 89, in __init__
>     self._frame = Frame(
>                   ~~~~~^
>         viewport=self.viewport,
>         ^^^^^^^^^^^^^^^^^^^^^^^
>     ...<6 lines>...
>         canvas_rect=parent.get_pygfx_render_area(),
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     )
>     ^
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_frame.py", line 162, in __init__
>     self._rect_manager = RectManager.from_extent(extent, canvas_rect)
>                          ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 125, in from_extent
>     rect = cls.extent_to_rect(extent, canvas_rect)
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 143, in extent_to_rect
>     RectManager.validate_extent(extent, canvas_rect)
>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
>   File "/home/clewis/repos/fastplotlib/fastplotlib/layouts/_rect.py", line 193, in validate_extent
>     raise ValueError(
>         f"extent: {extent} y-range is beyond the bounds of the canvas: {canvas_extent}"
>     )
> ValueError: extent: [  0 200 280 560] y-range is beyond the bounds of the canvas: [  0. 640.   0. 480.]
> ```

I found the cause:

https://github.com/pygfx/rendercanvas/blob/12e1919d1cbf1702a862a1ae908781f81521e6f3/rendercanvas/glfw.py#L184-L185

kushalkolar:
@clewis7 should be gtg! :partying_face:

kushalkolar:
All tests using pygfx@main passing, docs built!

clewis7 (approved):
:partying_face: :partying_face: :partying_face:

Issue #741, opened 2025-03-03
Universal store model with bidirectional publisher-subscriber


kushalkolar:
I just got this idea which might allow us to create a universal model of the store model @clewis7 wrote. Not fully thought out but a starting point.

1. Every publisher can be a subscriber, and every subscriber can be a publisher. 
2. The `StoreModel` receives or emits values of a certain type, call it the "store-data type".
3. Publishers register not only a callback and the event type, but also a function that maps the callback output to the store-data type.
4. Subscribers register a callback function which takes the subscriber and the store-data type.

Examples:

Sync multiple selectors:

```python
def selection_changed(ev) -> float:
  # returns data type used by subscriber function set_selection
  return ev.selection

def set_selection(selector, new_value: float):
  # new_value is returned from selection_changed and used here
  selector.selection = new_value

for selector in [sel1, sel2, sel3]:
  # the first arg must be an object with an `add_event_handler` method
  store_model.add_publisher(selector, selection_changed, "selection")
  store_model.subscribe(selector, set_selection)
```

How this works internally:

```python
# `StoreModel.add_publisher(obj, func_map_to_store, event_types)`
obj.add_event_handler(partial(StoreModel.notify_subscribers, obj, func_map_to_store), *event_types)

# `StoreModel.add_subscriber(obj, func_map_from_store)`
self._subscribers.add((obj, func_map_from_store))

# called by graphic/renderer when event happens
# `StoreModel.notify_subscribers(event_source, func_map_to_store, ev)`
new_value = func_map_to_store(ev)
# block events for all graphics
graphics = (sub[0] for sub in self.subscribers)
# call all subscribers
with pause_events(*graphics):
  for subscriber in self.subscribers:
    # unpack tuple
    obj, func = subscriber
    # skip if event source is this subscriber object
    if obj is event_source:
      continue
    # call subscriber func
    func(obj, new_value)
```

This idea can also be extended to have complex bidirectional events across event types since the functions allow to map to a central store-data-type and then inverse map to the subscriber function.

Decorator syntax would be nice here 😄 

```python
@store_model.add_publisher([sel1, sel2, sel3], "selection")
def selection_changed(ev) -> float:
  # returns data type used by subscriber function set_selection
  return ev.selection

@store_model.subscribe([sel1, sel2, sel3])
def set_selection(selector, new_value: float):
  # new_value is returned from selection_changed and used here
  selector.selection = new_value
```

[...]

PR #744, opened 2025-03-03
implemenet `@block_reentrance` decorator

kushalkolar:
This simple decorator applied on all the `GraphicFeature.set_value()` methods should allow us to create complex event graphs with cycles. Users don't have to think about managing bidirectional events :smile: 

related #741 

Example:

```python
import fastplotlib as fpl
import numpy as np

fig = fpl.Figure((3, 1))
sels = list()

for subplot in fig:
    lg = subplot.add_line(np.sin(np.linspace(0, np.pi * 10, 100)))
    sel = lg.add_linear_selector()
    sels.append(sel)


# create event handler and just add it to all graphics
# don't need to worry worry when event graphs have cycles
def set_selection(ev):
    for sel in sels:
        sel.selection = ev.info["value"]

for sel in sels:
    sel.add_event_handler(set_selection, "selection")

```

Need to test with more graphics and also test when exceptions are raised in `GraphicFeature.set_value()`

@apasarkar @FlynnOConnell you guys will love this!

### DIFF
--- a/examples/selection_tools/unit_circle.py
+++ b/examples/selection_tools/unit_circle.py
@@ -1,0 +1,114 @@
+"""
+Unit circle
+===========
+
+Example with linear selectors on a sine and cosine function that demonstrates the unit circle.
+
+This shows how fastplotlib supports bidirectional events, drag the linear selector on the sine
+or cosine function and they will both move together.
+
+Click on the sine or cosine function to set the colormap transform to illustrate the sine or
+cosine function output values on the unit circle.
+"""
+
+# test_example = false
+# sphinx_gallery_pygfx_docs = 'screenshot'
+
+
+import numpy as np
+import fastplotlib as fpl
+
+
+# helper function to make a cirlce
+def make_circle(center, radius: float, n_points: int) -> np.ndarray:
+    theta = np.linspace(0, 2 * np.pi, n_points)
+    xs = radius * np.cos(theta)
+    ys = radius * np.sin(theta)
+
+    return np.column_stack([xs, ys]) + center
+
+
+# create a figure with 3 subplots
+figure = fpl.Figure((3, 1), names=["unit circle", "sin(x)", "cos(x)"], size=(700, 1024))
+
+# set the axes to intersect at (0, 0, 0) to better illustrate the unit circle
+for subplot in figure:
+    subplot.axes.intersection = (0, 0, 0)
+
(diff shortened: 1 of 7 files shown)


kushalkolar:
added a simple example with trivially "synced" selectors:

https://github.com/user-attachments/assets/21c1c202-23b0-4561-b739-7748d7e8aeec

kushalkolar:
The decorator is working as expected and exceptions are also raised correctly, @clewis7 if you can test on your end too it should be gtg! This is probably my favorite PR, small few lines decorator that makes life so much easier :partying_face:

examples/selection_tools/unit_circle.py
```diff
@@ -0,0 +1,114 @@
...
+
+This shows how fastplotlib supports bidirectional events, drag the linear selector on the sine
+or cosine function and they will both move together.
+
+Click on the sine or cosine function to set the colormap transform to illustrate the sine or
+cosine function output values on the unit circle.
+"""
+
+# test_example = false
+# sphinx_gallery_pygfx_docs = 'screenshot'
```
clewis7, line 15: could make this animate if you do something like the cov matrix example where you have a function to +1

examples/selection_tools/unit_circle.py
```diff
@@ -0,0 +1,114 @@
...
+def make_circle(center, radius: float, n_points: int) -> np.ndarray:
+    theta = np.linspace(0, 2 * np.pi, n_points)
+    xs = radius * np.cos(theta)
+    ys = radius * np.sin(theta)
+
+    return np.column_stack([xs, ys]) + center
+
+
+# create a figure with 3 subplots
+figure = fpl.Figure((3, 1), names=["unit circle", "sin(x)", "cos(x)"], size=(700, 1024))
```
clewis7, line 32: does the gallery like this figure size? I thought it had to be something else to appear correctly in the docs?

examples/selection_tools/unit_circle.py
kushalkolar: This is very short, the user should just try it out

examples/selection_tools/unit_circle.py
kushalkolar: Extra height upto 1024 is fine (texture size limit we imposed). Width other than 700 causes weirdness on docs

kushalkolar:
No API changes here.

clewis7 (approved):

PR #745, opened 2025-03-04
calcium imaging toy example

kushalkolar:
An example that generates some calcium imaging-like toy data, performs decomposition, and visualizes. Should put all the generation and decomposition functions in some other file though.

also contains some fixes for the functions that auto-get the linear selector ranges and `get_nearest_graphics()`. We should ideally just get a line shader for a infinite line.

Depends on #744


kushalkolar:
I should rebase this on main and merge, has some important fixes that I thought I had fixed but wondered where they went.

kushalkolar:
this is ndwidget stuff now, the data gen might be useful though for toy examples

PR #773, opened 2025-03-19
more event docs, make some things public

kushalkolar:
WIP

- [x] cleanup docs on events.
- [x] better event tables, list of dicts to encode event table info, more maintainable, parse for docs
- [x] More event examples
    - [x] image data
        - [x] click and move pointer around image to change the data, like pyqtgraph ImageItem draw example
    - [x] cmap between image, scatter, line
    - [x] image click
    - [x] key events
    - [x] line data and thickness event
    - [x] "pointer_move" combined with  `get_nearest_graphics()` with circles
    - [x] scatter click, scatter "hover", i.e. pointer_move event
    - [x] scatter "hover" to show same datapoint across data transformation, based on sklearn example
    - [x] drag points
    - [x] example similar to pyqtgraph Text Item example
        - [x] have a sinc wave, scatter point at the peak with text indicating some text and the position
        - [x] move a scatter point with a text label along the line on each draw using animation function
- [x] Controller examples, ids, sync, etc. all combinations
- [x] add docs on how to integrate with UI libs
    - [x] ipywidget examples
- [x] docs on `utils.plot_helpers`
- [x] make `graphics._features` public under `fpl.graphics`, this allows the various features to be used by users for instance checks and similar purposes.
- [x] rename `FeatureEvent` -> `GraphicFeatureEvent`, make it a top level public import.
- [x] make `graphics.Graphic` accessible publicly, also useful for instance checks etc.
- [x] make `Subplot` accessible under `fpl.Layouts`


kushalkolar:
@clewis7 replacing all hard-coded rst event info tables with list of dicts, more maintainable, easy to parse it into a table for docs.

```python
class VertexPositions(BufferManager):
    property_name = "data"
    event_info_spec = [
        {
            "dict key": "key",
            "type": "slice, index (int) or numpy-like fancy index",
            "description": "key at which vertex positions data were indexed/sliced"
        },
        {
            "dict key": "value",
            "type": "int | float | array-like",
            "description": "new data values for points that were changed"
        },
    ]
```

kushalkolar:
Thoughts for making classes under graphics.features public:

* Keep the name `GraphicFeature`, no renaming to `GraphicProperty` because then that confuses it with the actual properties on the graphics. A `GraphicFeature` is different from the actual property attributes on a `Graphic` which are the user-facing properties to modify a `GraphicFeature`.
* Rename `FeatureEvent` to `GraphicFeatureEvent` ? More explicit.
* Expose `GraphicFeatureEvent` at top level, i.e. `fpl.GraphicFeatureEvent`. Very useful for type annotations of event handlers.
* Users will not need to use `GraphicFeature` classes, except for advanced users who want to make their own `Graphic` classes etc. I cannot think of a case where a standard user would need access to a `GraphicFeature` class. So, although we make graphic feature classes part of the public API it will need to be accessed explicitly, ex: `from fastplotlib.graphics.features import VertexColors`

@clewis7

clewis7:
> Thoughts for making classes under graphics.features public:
> 
> * Keep the name `GraphicFeature`, no renaming to `GraphicProperty` because then that confuses it with the actual properties on the graphics. A `GraphicFeature` is different from the actual property attributes on a `Graphic` which are the user-facing properties to modify a `GraphicFeature`.
> * Rename `FeatureEvent` to `GraphicFeatureEvent` ? More explicit.
> * Expose `GraphicFeatureEvent` at top level, i.e. `fpl.GraphicFeatureEvent`. Very useful for type annotations of event handlers.
> * Users will not need to use `GraphicFeature` classes, except for advanced users who want to make their own `Graphic` classes etc. I cannot think of a case where a standard user would need access to a `GraphicFeature` class. So, although we make graphic feature classes part of the public API it will need to be accessed explicitly, ex: `from fastplotlib.graphics.features import VertexColors`
> 
> @clewis7

I like keeping the name the same. I also like `GraphicFeatureEvent` because it is more explicit. 

Aside from type annotations, when would a user need `fpl.GraphicFeatureEvent`? If it is just for that, I am not sure we should expose it...maybe better to just have really good docs. 

I like the explicit access of a `GraphicFeature` just because I agree no one would ever need it unless they were doing something custom.

kushalkolar:
> Aside from type annotations, when would a user need `fpl.GraphicFeatureEvent`? If it is just for that, I am not sure we should expose it...maybe better to just have really good docs.

Earlier we kinda came to the conclusion that anything which is use ~80% of the time should be a top level import, and event it's used for type annotations it's still very useful and events are used very often so I've made it a top level import. It's a pretty clean import so there aren't issues with doing this.

PR #774, opened 2025-03-22
update gov and CoC about LLM spam

kushalkolar:


kushalkolar:
on another note, this is one of the rare moments where tests are passing on both main and release; we should do a release :joy:

clewis7 (approved):
LGTM

PR #779, opened 2025-03-27
Use pyproject.toml

almarklein:
In future PRs I can do:
* use same `_version.py` mechanism as pygfx/wgpu, that e.g. includes nr of commits since last version-tag.
* Formatting and linting with ruff, perhaps?


almarklein:
Grrr, I thought I had done `git pull` but that failed, so this is a bit behind :)

pyproject.toml
```diff
@@ -1,3 +1,78 @@
...
+    "science",
+    "interactive",
+    "pygfx",
+    "webgpu",
+    "wgpu",
+    "vulkan",
+    "gpu",
+]
+requires-python = ">= 3.10"
+dependencies = ["numpy>=1.23.0", "pygfx~=0.9.0", "wgpu>=0.20.0", "cmap>=0.1.3"]
```
kushalkolar, line 21: I think the failure is because this is one line, if each dep is on its own line then the `sed` command in the workflow that uses `pygfx@main` will remove the pygfx line from the file.

kushalkolar: Also I just closed #776 so we can put the lower version pin of `pygfx>=0.9.0` right here instead.

kushalkolar:
> * use same `_version.py` mechanism as pygfx/wgpu, that e.g. includes nr of commits since last version-tag.
> * Formatting and linting with ruff, perhaps?

Sure! I need to go through the codebase and my pycharm and decide how we want to configure ruff, I don't think there's a way to download the rules I'm currently ignoring in pycharm. flake8 is too strict sometimes so we ignore a bunch of things.

fastplotlib/layouts/_graphic_methods_mixin.py
kushalkolar, line 1: I wonder if black changed trailing commas recently, the autogen script blacks this file after creating it:

https://github.com/fastplotlib/fastplotlib/blob/88359d32be8a68892a7279db1b665750d74b8eae/scripts/generate_add_graphic_methods.py#L85-L87

pyproject.toml
```diff
@@ -1,3 +1,84 @@
...
+[project.urls]
+Homepage = "https://www.fastplotlib.org/"
+Documentation = "https://www.fastplotlib.org/"
+Repository = "https://github.com/fastplotlib/fastplotlib"
+
+# ===== Building
+
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["flit_core >=3.2,<4"]
```
kushalkolar, line 69: I just look into flit, it seems nice! Do we need to add an `[include]` section to replace the MANIFEST.in?

kushalkolar:
So the deploy docs failure is because the secrets are only in the org. If you set your remote origin to `https://github.com/fastplotlib/fastplotlib.git` and then push and make a new PR it should work.

pyproject.toml
almarklein: Nope, Flit by default includes everything under the `fastplotlib` dir. I double-checked that the file in `assets` are in the distribution. The directory `fastplotlib/utils/colormap` does not exist, so I think this line in the manifest is outdated.

fastplotlib/layouts/_graphic_methods_mixin.py
almarklein: The weird thing is that if I run `black` using the latest version, it changes *a lot* of files. Not sure what's going on there.

almarklein:
> If you set your remote origin to https://github.com/fastplotlib/fastplotlib.git and then push and make a new PR it should work.

I'm not sure I understand. Does this mean only PR's from devs with write-access to the main repo can be merged?

kushalkolar:
> > If you set your remote origin to https://github.com/fastplotlib/fastplotlib.git and then push and make a new PR it should work.
> 
> I'm not sure I understand. Does this mean only PR's from devs with write-access to the main repo can be merged?

We can still see if the docs build step succeeded, but uploading to the server requires the secrets so we skip it for external contributors for now until we figure out a better way.

fastplotlib/layouts/_graphic_methods_mixin.py
kushalkolar: We only black the source files, not the examples

pyproject.toml
kushalkolar: Ah thanks!

almarklein:
Superseded by #782

PR #784, opened 2025-03-31
Use _version.py that includes details using git

almarklein:
Using the same module as pygfx and wgpu, which includes more detailed version info when you're inside a git repo.

### DIFF
--- a/fastplotlib/_version.py
+++ b/fastplotlib/_version.py
@@ -1,4 +1,112 @@
+"""
+Versioning: we use a hard-coded version number, because it's simple and always
+works. For dev installs we add extra version info from Git.
+"""
+
+import logging
+import subprocess
+from pathlib import Path
+
+
+# This is the reference version number, to be bumped before each release.
+# The build system detects this definition when building a distribution.
 __version__ = "0.5.0"
+
+# Allow using nearly the same code in different projects
+project_name = "fastplotlib"
+
+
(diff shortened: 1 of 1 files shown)


kushalkolar:
So with this when it's a git version number, it's interpreated as version x.y.z + git-commit? So from now on bump right before release? I would usually bump after a release.

kushalkolar (approved):

almarklein:
> So with this when it's a git version number, it's interpreated as version x.y.z + git-commit? 

Yes, e.g. my current pygfx is `0.7.0.post56+g092bc843.dirty`, which is 56 commits after the last git vx.y.z tag, the git commit hash is also present. And the `.dirty` means that I have uncommitted changes :)

> So from now on bump right before release? I would usually bump after a release.

Ah, so I guess that in your previous system the `main` is a bit like a preview (wip) of the upcoming release. 

An advantage of bumping right before a release is that you can pick what number to bump, which usually depends a bit on how much (the API) has changed since the previous release.

I don't want to force either system on you. Actually, you could still bump after a release, and we could change the `post` to `pre` 🤷

kushalkolar:
> Ah, so I guess that in your previous system the `main` is a bit like a preview (wip) of the upcoming release.

yup, let's just follow the same system for fpl after the next release, easier for everyone!

PR #785, opened 2025-03-31
Ruff

almarklein:
* [x] Update `pyproject.toml`
* [x] Update CI job.
* [x] Ran `ruff .` and `ruff check --fix`.
* [ ] Fix more errors / add to ignore list.  

Just a quick draft for now. Let's first see if this works for you. @kushalkolar you mentioned flake8 rules in pycharm. I would expect pycharm to pick up the ruff rules from the pyproject, but I'm not sure.


kushalkolar:
Thanks, it will take a while to go through this

almarklein:
> Thanks, it will take a while to go through this

Not sure if you mean fixing the errors that are still reported. There are quite a few, but many are similar. So with a bit of 💪 its should not take *that* long, in my experience.

PR #791, opened 2025-04-12
add `ImageVolumeGraphic`

kushalkolar:
actually going to merge it this time

- [ ] `TextureArrays` supports 3D textures
- [ ] implement simple `ImageVolumeGraphic`
- [ ] implement graphic features for all the different volume materials
- [ ] examples
    - [ ] basic
    - [ ] 3D video, maybe generated calcium toy data?
    - [ ] 2D video over time, maybe from the calcium example I added to imageio
    - [ ] other toy data, does sklearn have any relevant ones?
- [ ] fix some typos in `graphics.image`
- [ ] `scripts/generate_add_graphic_methods.py` is better, removed `Graphic.type` attribute which was only used in the generate mixin script and use regex instead to create the `add_<graphic>` method names.
- [ ] tests, screenshots, API, new graphic, TextureArray for 2d and 3d, and any new graphic features.
- [ ] improve docstring for quick_min_max()

@almarklein are RGB volumes supported? Is it even a thing? 

selector tools for slicing, cube slicing, etc., non-orthogonal slicing to come later

### DIFF
--- a/fastplotlib/graphics/image_volume.py
+++ b/fastplotlib/graphics/image_volume.py
@@ -1,0 +1,229 @@
+from typing import *
+
+import pygfx
+
+from ..utils import quick_min_max
+from ._base import Graphic
+from .features import (
+    TextureArray,
+    ImageCmap,
+    ImageVmin,
+    ImageVmax,
+    ImageInterpolation,
+    ImageCmapInterpolation,
+)
+
+
+class _VolumeTile(pygfx.Volume):
+    """
+    Similar to pygfx.Volume, only difference is that it modifies the pick_info
+    by adding the data row start indices that correspond to this chunk of the big Volume
+    """
+
+    def __init__(
+        self,
+        geometry,
+        material,
+        data_slice: tuple[slice, slice, slice],
+        chunk_index: tuple[int, int, int],
+        **kwargs,
+    ):
+        super().__init__(geometry, material, **kwargs)
+
+        self._data_slice = data_slice
+        self._chunk_index = chunk_index
+
+    def _wgpu_get_pick_info(self, pick_value):
+        pick_info = super()._wgpu_get_pick_info(pick_value)
+
+        data_row_start, data_col_start, data_z_start = (
+            self.data_slice[0].start,
+            self.data_slice[1].start,
+            self.data_slice[2].start,
+        )
+
+        # add the actual data row and col start indices
+        x, y, z = pick_info["index"]
+        x += data_col_start
+        y += data_row_start
+        z += data_z_start
+        pick_info["index"] = (x, y, z)
+
+        xp, yp, zp = pick_info["voxel_coord"]
+        xp += data_col_start
+        yp += data_row_start
+        zp += data_z_start
+        pick_info["voxel_coord"] = (xp, yp, zp)
+
+        # add row chunk and col chunk index to pick_info dict
+        return {
+            **pick_info,
+            "data_slice": self.data_slice,
+            "chunk_index": self.chunk_index,
+        }
+
+    @property
+    def data_slice(self) -> tuple[slice, slice, slice]:
+        return self._data_slice
+
+    @property
+    def chunk_index(self) -> tuple[int, int, int]:
+        return self._chunk_index
+
+
+class ImageVolumeGraphic(Graphic):
+    _features = {
+        "data": TextureArray,
+        "cmap": ImageCmap,
+        "vmin": ImageVmin,
+        "vmax": ImageVmax,
+        "interpolation": ImageInterpolation,
+        "cmap_interpolation": ImageCmapInterpolation,
+    }
+
+    def __init__(
+            self,
+            data: Any,
+            mode: str = "ray",
+            vmin: int = None,
+            vmax: int = None,
+            cmap: str = "plasma",
+            interpolation: str = "nearest",
+            cmap_interpolation: str = "linear",
+            isolated_buffer: bool = True,
+            **kwargs,
+    ):
+        valid_modes = ["basic", "ray", "slice", "iso", "mip", "minip"]
+        if mode not in valid_modes:
+            raise ValueError(f"invalid mode specified: {mode}, valid modes are: {valid_modes}")
+
+        super().__init__(**kwargs)
+
+        world_object = pygfx.Group()
+
+        # texture array that manages the textures on the GPU that represent this image volume
+        self._data = TextureArray(data, dim=3, isolated_buffer=isolated_buffer)
+
+        if (vmin is None) or (vmax is None):
+            vmin, vmax = quick_min_max(data)
+
+        # other graphic features
+        self._vmin = ImageVmin(vmin)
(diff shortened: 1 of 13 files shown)


kushalkolar:
took some 3d movie gen code from caiman to generate a 3d movie:

@FlynnOConnell 

https://github.com/user-attachments/assets/a26cd878-95ea-4339-8c58-3209bbbfc2c1

FlynnOConnell:
``` python
File [~\repos\fastplotlib\fastplotlib\graphics\features\_image.py:147](http://localhost:8888/lab/tree/demos/notebooks/~/repos/fastplotlib/fastplotlib/graphics/features/_image.py#line=146), in TextureArray._fix_data(self, data)
    145 def _fix_data(self, data):
    146     if data.ndim not in (2, 3):
--> 147         raise ValueError(
    148             "image data must be 2D with or without an RGB(A) dimension, i.e. "
    149             "it must be of shape [rows, cols], [rows, cols, 3] or [rows, cols, 4]"
    150         )
    152     # let's just cast to float32 always
    153     return data.astype(np.float32)

ValueError: image data must be 2D with or without an RGB(A) dimension, i.e. it must be of shape [rows, cols], [rows, cols, 3] or [rows, cols, 4]
```

``` python
data = np.random.randn(50, 2, 448, 448)
data.shape
>> (50, 2, 448, 448)
```

The ability to view a time-series with a 4th dimension would be nice here

FlynnOConnell:
Also how difficult is something like `scale=[1, 4, 1]` to account for anisotropy in XY and Z? Here I have a volume of `[14, 448, 448] ZXY`, but from the side its a flat plane. This dataset is `[2um x 2um x 17um]`

image

FlynnOConnell:
> Also how difficult is something like `scale=[1, 4, 1]` to account for anisotropy in XY and Z? Here I have a volume of `[14, 448, 448] ZXY`, but from the side its a flat plane. This dataset is `[2um x 2um x 17um]`
> 
> image

nevermind :) 

``` python
volume.world_object.world.scale_z = 15.0
```

Some other helpful tidbits:

``` python
hlut = fpl.HistogramLUTTool(volume.data.value, volume)
volume.interpolation = "linear"
volume.cmap = "gnuplot2"

fig[0, 0].docks["right"].add_graphic(hlut)
fig[0, 0].docks["right"].size = 80
```

``` python
fig[0, 0].docks["right"].controller.enabled = False
fig[0, 0].docks["right"].camera.maintain_aspect = False
fig[0, 0].docks["right"].auto_scale(maintain_aspect=False)
```

almarklein:
> @almarklein are RGB volumes supported? Is it even a thing?

Yes, looks like it is; it uses the same logic (in the shader) as the image. I have never encountered it in the wild though.

kushalkolar:
ok all the basics are down, switchable materials, can change material props:

https://github.com/user-attachments/assets/767767df-ac8c-46d5-83f8-e9203405dae0

Need to figure out why regex is turning `VolumeSliceMaterial` into `slic` instead of `slice`

kushalkolar:
In other news, the latest pycharm is full of bugs, wtf is this. Also it wanted to auto insert the old rst structured docstring template even though I configured it to do numpydoc format :vomiting_face: 

image

kushalkolar:
All that's left is tests, 2D over time example, and RGB(A) examples. I also think I want to merge #849 and then rebase this so the examples will be much simpler.

Also pygfx has non-orthogonal slicing:

https://github.com/user-attachments/assets/eba558bb-704a-4e34-852f-8519cf9a25c6

kushalkolar:
ok there's a bug in the `TextureArray` or tiling somewhere, because when a dim is greater than the limit it is tiling along the wrong dim :laughing: 

image

kushalkolar:
The rendering at the edges is weird if one of the dims is larger than the 3d texture size limit so let's just not support this, I think the uses are very much rare edge cases. Also if the 3d texture size limit is 16,383, then 16384x16384x16384 with float32 is 17 TB which is more than any GPU that I know of can handle :joy: .

kushalkolar:
> ok there's a bug in the `TextureArray` or tiling somewhere, because when a dim is greater than the limit it is tiling along the wrong dim 😆
> 
> image

This is fixed now, wrong indexing (z is first index, not last; and need to have both `chunk_index` and `data_slice` as `tuple` **NOT** `list` in `__next__`). Was easier to just have a seprate `TextureArrayVolume` for volumes, restored the original Image graphic's `TextureArray` for 2D images.

kushalkolar:
More tests for the TextureArray for volumes would be nice, but can do that later. I think tiling of volume is a rare edge case so if there is a bug it will affect very rare usecases not already covered by the tests.

examples/image_volume/image_volume_mip.py
```diff
@@ -0,0 +1,47 @@
+"""
+Volume Mip mode
+===============
+
+View a volume using MIP rendering.
```
clewis7: I would maybe extend this docstring and explain what MIP rendering is. Or, we should have somewhere in the docs where the different rendering modes are explained

fastplotlib/graphics/features/_volume.py
```diff
@@ -0,0 +1,449 @@
...
+        self._iter = None
+
+        # iterate through each chunk of passed `data`
+        # create a pygfx.Texture from this chunk
+        for _, buffer_index, data_slice in self:
+            texture = pygfx.Texture(self.value[data_slice], dim=3)
+
+            self.buffer[buffer_index] = texture
+
+        self._shared: int = 0
```
clewis7: code comment here

fastplotlib/graphics/features/_volume.py
```diff
@@ -0,0 +1,449 @@
...
+    @property
+    def zdim_indices(self) -> np.ndarray:
+        """
+        z dimension indices that are used to chunk the big data array
+        into individual Textures on the GPU
+        """
+        return self._zdim_indices
+
+    @property
+    def shared(self) -> int:
```
clewis7: doc string

fastplotlib/graphics/image_volume.py
```diff
@@ -0,0 +1,420 @@
...
+    ):
+        """
+
+        Parameters
+        ----------
+        data: array-like
+            array-like, usually numpy.ndarray, must support ``memoryview()``.
+            Shape must be [n_planes, n_rows, n_cols] for grayscale, or [n_planes, n_rows, n_cols, 3 | 4] for RGB(A)
+
+        mode: str, default "ray"
```
clewis7: mismatch between default above and here

fastplotlib/graphics/features/_volume.py
kushalkolar: ah this didn't get removed in #842

FlynnOConnell:
Amazing! @kushalkolar I'll test this out!

kushalkolar:
@clewis7 gtg! :partying_face: and then we can merge #904 once I make new ground truth screenshots for that and then #883

Issue #781, opened 2025-03-29
Improvements to selector tools


kushalkolar:
Ideas after meeting with @almarklein 

- [x] Infinite line in pygfx
- ~[ ] See if we can just use geo bbox for RectangleSelector~
- ~[ ] Maybe draw these in an overlay render pass? Would allow 2D selector tools to be used in perspective projections. Rectangle selectors, polygon, circle etc. will have to scale w.r.t. the scene.~ Almar has now done this with render queues
- [x] Polygon selector
- [ ] lasso selector
- [ ] allow setting a max range, Useful for LinearRegionSelector and any future resizable rectangle or polygon selectors. If they are tied to displaying a large sub-selection of data, it can be useful to prevent the selection from ever becoming too large.
- [ ] step size, useful for block based image algorithms and other use cases

kushalkolar:
@almarklein I think the Polygon selector can be turned into a "lasso" (pointer down and move to draw) by some tweaks to events right? I can do this.

almarklein:
I'm cleaning up my inbox, and realize I had missed this earlier. Anyway, yeah, I think I'd add logic to the pointer-move handler, and if the pointer has moved more than say 10 logical pixels, introduce a new vertex.

PR #799, opened 2025-04-15
Improve hover color behavior for selectors

almarklein:
Ref #781

I noticed that the color of the selector was flickering during motion, as the mouse enters and exits the object. This PR freezes the color while `._moving`, keeping track of color changes in a dict, and then applying that dict when the moving is done.

Notice how the corner vertex becomes highlighted as the mouse is released after dragging the edge:

https://github.com/user-attachments/assets/5f5db53c-cccf-43f5-8859-f1dc1589f8cc


kushalkolar (approved):

kushalkolar:
Thanks! I also want to refactor it to use the `SelectorColorStates` later

PR #800, opened 2025-04-15
Refactor selector drag behavior

almarklein:
Ref #781

Refactor to move the selectors based on their original value, instead of using incremental deltas. This makes it much easier to implement natural behavior when the movement is clipped by the limits.

One question: What's up with the `PolygonSelector`, I see no example of it, and it also does not implement `_move_graphic`.

### DIFF
--- a/fastplotlib/graphics/selectors/_base_selector.py
+++ b/fastplotlib/graphics/selectors/_base_selector.py
@@ -277,7 +277,12 @@ class BaseSelector(Graphic):
         """
         position = self._plot_area.map_screen_to_world(ev)
 
-        self._move_info = MoveInfo(start_selection=None, start_position=position, delta=np.zeros_like(position), source=event_source)
+        self._move_info = MoveInfo(
+            start_selection=None,
+            start_position=position,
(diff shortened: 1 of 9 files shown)


kushalkolar:
The polygon selector module in there was a prototype, it probably has to be rewritten.

fastplotlib/graphics/selectors/_base_selector.py
```diff
@@ -16,9 +16,14 @@ class MoveInfo:
...
     """
 
-    # last position for an edge, fill, or vertex in world coordinates
-    # can be None, such as key events
-    last_position: Union[np.ndarray, None]
+    # The initial selection. Differs per type of selector
+    start_selection: Any
+
+    # The initial world position of the cursor
+    start_position: Union[np.ndarray, None]
```
kushalkolar, line 23: The `Union` is lingering from python 3.9 before we had `|`

almarklein: fixed!

kushalkolar (approved):
this is great thanks!

PR #803, opened 2025-04-18
Fixes array issue that was causing incompatibility in the subsample a…

### LINKED ISSUE #804: Unexpected subsample_array behavior
Right now, [this line](https://github.com/fastplotlib/fastplotlib/blob/c95ed34f6e0832f80479d03c28e76cfdac559a97/fastplotlib/utils/functions.py#L459) leads to errors if "arr" is not a np.ndarray. Need to do "arr[:]" to avoid unexpected behavior.

apasarkar:
Small change, allows lazy arrays to pass as expected through the subsample (and the vmin/vmax quick computation) functions. Fixes #804

### DIFF
--- a/fastplotlib/utils/functions.py
+++ b/fastplotlib/utils/functions.py
@@ -456,7 +456,7 @@ def subsample_array(
         subsample of the input array
     """
     if np.prod(arr.shape) <= max_size:
-        return arr  # no need to subsample if already below the threshold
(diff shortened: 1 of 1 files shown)


kushalkolar:
Ah this is because whatever `np.nanmin` calls isn't implemented for your lazy arrays. If you have a way to pre-compute the min and max for your lazy arrays that would be fastest.

kushalkolar (approved):

PR #806, opened 2025-04-25
coerce dtype=np.int64 for np.prod calls in subsample

FlynnOConnell:
Some of my datasets are overflowing during histogram calculation. 

```python
raw_scan.shape
Out[1]: (5632, 14, 448, 448)
np.prod(raw_scan.shape)
Out[2]: -1354760192
np.prod(raw_scan.shape, dtype=np.int64)
Out[3]: 15825108992
```

Also, `ignore_dims` won't work due to tuple item assignment. Changed that to a list

### DIFF
--- a/fastplotlib/utils/functions.py
+++ b/fastplotlib/utils/functions.py
@@ -455,11 +455,11 @@ def subsample_array(
     np.ndarray
         subsample of the input array
     """
-    if np.prod(arr.shape) <= max_size:
+    if np.prod(arr.shape, dtype=np.int64) <= max_size:
         return arr[:]  # no need to subsample if already below the threshold
 
     # get factor by which to divide all dims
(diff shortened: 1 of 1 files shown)


fastplotlib/utils/functions.py
```diff
@@ -455,17 +455,17 @@ def subsample_array(
     np.ndarray
         subsample of the input array
     """
-    if np.prod(arr.shape) <= max_size:
+    if np.prod(arr.shape, dtype=np.int64) <= max_size:
```
kushalkolar, line 458: I think a better way is:

```
full_shape = np.array(arr.shape, dtype=np.uint64)
```

and then use this throughout

also note the uint64 instead of int64, has a higher bound

kushalkolar (changes requested):

kushalkolar (approved):

Issue #772, opened 2025-03-18
Superimposing Reference Frames


kushalkolar:
I think we can easily have multiple different reference spaces in one PlotArea using the Scene's world affine transform's scale, i.e. `Scene.world.scale`. Would be the way we do things like matplotlib's `twinx` functionality, but we can do it much more elegantly.

Also automatic linking with selectors to update the axes. Allow update the "reference space" ranges dynamically using a selector. For example, to show a zoomed in range of a large range in another subplot, update the axes dymically.

nxnen8282:
Our system attempted to address this issue:

Added model space transforms to allow different coordinate systems in one PlotArea. This enables functionality similar to matplotlib's twinx() but more flexible.

Key changes:
- Added ModelSpace class to manage coordinate transformations
- Added model_space property to PlotArea class
- Added methods to create and manage model spaces
- Updated Graphic class to support model space transforms
- Added helper methods for common transformations

You can examine the suggested modifications here: https://github.com/nxnen8282/fastplotlib-fastplotlib-772/commit/4b00e32b2d2ea190ba9f87fdcaa2f29fd9ec1fe3.


> [!CAUTION]
> Advisory: This solution was generated using automated tools, and we strongly recommend reviewing the code thoroughly before implementation. The solution may be partial - please consider it as inspiration for your own implementation.

---

http://missionfixinternet.com

kushalkolar:
Played with things a bit, I think a better way is to scale the `camera` for a given model space and not the scene!

A model space can maybe be some simple abstraction of the `PlotArea` class, i.e. mulitple plot areas overlaid to create a subplot. Need to see how controllers and stuff works with this!

kushalkolar:
`PlotArea` should have a primary reference frame which is accessed using the current API, i.e. `PlotArea.scene`, `PlotArea.camera` etc. all point to the primary `ReferenceFrame`. All additional reference frames must be accessed _explicitly_.

Not sure if `PlotArea.auto_scale()` and other similar function should only autoscale the primary reference frame or all reference frames? What should the methods for affecting all reference frames be? Such as for centering all frames so that their origins overlap?

kushalkolar:
It would be very useful to be able to have cameras or scenes (depending on how we implement this) at different scales, and then be able to pan and zoom in the main reference frame and have another controller that pans to the same worldspace position in another reference frame.

[...]

Issue #729, opened 2025-02-16
`pint` integration


kushalkolar:
https://github.com/hgrecco/pint

kushalkolar:
We might be able to create a general store model that uses pint to convert units and automatically spits things out if they make sense. Example here for time, but a user could define the units for each axis of a data array and as long as the units make sense a very general simple store model can sync them up.

```python
In [1]: import pint

In [2]: ureg = pint.UnitRegistry()

# define some sampling rates
In [3]: rate1 = 10 * ureg.Hz

In [4]: rate2 = 30 * ureg.Hz

# some timepoint
In [5]: timepoint = 1.5 * ureg.second

# define index which is really just dimensionless
In [11]: ureg.define("index = second * hertz")

# get index of that timepoint for rate2
In [12]: (timepoint * rate2).to("index")
Out[12]: <Quantity(45.0, 'index')>

# other way around, index to timepoint
In [15]: (75 * ureg("index") / rate1).to("second")
Out[15]: <Quantity(7.5, 'second')>

In [16]: (75 * ureg("index") / rate2).to("second")
Out[16]: <Quantity(2.5, 'second')>
```

@clewis7

clewis7:
I think it would be cool for now to add an example using pint, but we should really consider whether it is something we would want to commit to long-term. 

I just am not sure how it feels from a dependency standpoint, even though the project seems active and well-supported. 

Also, would we want users directly interacting with pint to have to use fpl or would we want them to specify their units and then we set up the pint part under the hood

Needs more brainstorming I think

kushalkolar:
> I just am not sure how it feels from a dependency standpoint, even though the project seems active and well-supported.

I see it as a potential optional integration, like imgui

> Also, would we want users directly interacting with pint to have to use fpl or would we want them to specify their units and then we set up the pint part under the hood

If a user has used pint with their array, then have fpl be aware of it, such as setting axes labels for example.

> Needs more brainstorming I think

Yes definitely needs brainstorming, just posted this as a thought to get started :D

PR #810, opened 2025-05-06
Remove threshold from SubPlot.auto_scale

cassiersg:
From what I could see, the threshold was arbitrary. It is also not clear that any non-zero value should be a threshold, since all computations should work over the full range of a 32-bit float.

### DIFF
--- a/fastplotlib/layouts/_plot_area.py
+++ b/fastplotlib/layouts/_plot_area.py
@@ -597,9 +597,9 @@ class PlotArea(GraphicMethodsMixin):
             width, height, depth = (1, 1, 1)
 
         # make sure width and height are non-zero
-        if width < 0.01:
+        if width <= np.finfo(np.float32).smallest_normal:
             width = 1
-        if height < 0.01:
+        if height <= np.finfo(np.float32).smallest_normal:
             height = 1
 
         # scale all cameras associated with this controller else it looks wonky


kushalkolar:
I'm pretty sure this was due to floating point error where the width and height can be very close to zero after extensive panning and zooming.

cassiersg:
That seems likely indeed, but it seems like a quick fix that does not work in general, since it is reasonable for an axis to have very small values when showing scientific data (a nice plot would often benefit from proper units like `ns` instead of showing `1e-9` values, but in practice the latter is very common).

If this is an error caused by ill-conditioning after zooming and panning, I think that it should be possible to find better fixes. I also encountered some issues with extensive zooming (see https://github.com/pygfx/pygfx/issues/1081), so I'm interested in finding a proper solution to that problem. Do you have a testcase for the bug that motivated the introduction of this threshold?

kushalkolar:
> very small values

Ah good point, we'll have to think of a better fix

> a nice plot would often benefit from proper units like `ns` instead of showing `1e-9` values, but in practice the latter is very common.

pint would be a good way to do this, see #729 , setting units on a reference frame (see #772) would probably be a good way to do this.

cassiersg:
When comparing to `0.0`, that would amount to `abs(x) < atol`, and the question becomes what `atol` should be. numpy's default is close to the machine epsilon for 32-bit floats, which is reasonable in some cases, but not applicable in others. It's hard to say what the comparison should be without knowing what it expected. I don't know what are the edge cases to consider. One example would be a scatter plot with a single point: what is the expected behavior in that case?

kushalkolar:
Perhaps anything that is 1 bit above `float32` 0 could work as a threshold? The issue arises when the scene bbox [w, h] = 0, which results in a camera matrix which is unusable.

cassiersg:
By "1 bit above 0", if you mean the smallest non-zero positive float32, this is about `1.175494e-38`, and I believe that the check "> 0.0" or ">= 1.175494e-38` are equivalent (ignoring denormals and such).

I can update the PR to go from `!= 0.0` to `> 0.0` (that should be more robust w.r.t. pathological cases like negative values and NaNs).

kushalkolar:
Sounds like that should work, hardcoding the binary value should work, `0b0000...1` or something I think, do you want to test it or should I?

kushalkolar:
BTW looking at your actual use-case repo, I think your `SplitXYController` should be possible with the existing `pygfx` controllers?

And from this example it looks like you're just trying to create a `LineCollection`, but in a very roudabout way?

https://github.com/cassiersg/scaviz/blob/main/examples/traces_plot.ipynb

If you describe what you're trying to do we can help better.

cassiersg:
> Sounds like that should work, hardcoding the binary value should work, `0b0000...1` or something I think, do you want to test it or should I?

I switched to `np.finfo(np.float32).smallest_normal` (which is approx `np.float32(1.1754944e-38)`), I think that is it a reasonable value (at that order of magnitude, we keep full precision when computing on GPU).

> If you describe what you're trying to do we can help better.

Thanks for the suggestion. I started a discussion [here](https://github.com/fastplotlib/fastplotlib/discussions/811) to not derail this thread. (This PR is not discussed there, I noticed this issue when my code was buggy, I don't have a plot with scales below 0.01 anymore - but I expect to have it in the future.)

PR #830, opened 2025-05-20
Tooltips

### LINKED ISSUE #753: tooltips
Would be easy to create different kinds of tooltips:

1. simple text with a plane mesh behind it
2. pops up a `PlotArea` in an overlay render pass with its own camera to display any arbitrary scene with graphics

kushalkolar:
closes #753

prototype works, need to decide what the API should look like. Right now there is a `register_graphic()` method:

```python
    def register_graphic(
        self,
        graphic,
        event_type: str = "pointer_move",
        display_property: str = "data",
        formatter: callable = None,
    ):
        graphic.add_event_handler(partial(self._event_handler, display_property), event_type)
        graphic.add_event_handler(self._clear, "pointer_leave")
```

Should also have the option to add a custom event handler. It takes the event, and returns the text to display? @clewis7 thoughts?

```python
import numpy as np
import fastplotlib as fpl

fig = fpl.Figure(size=(1200, 600), shape=(1, 2))

sg = fig[0, 0].add_scatter(np.random.rand(1_000, 3), sizes=10)
ig = fig[0, 1].add_image(np.random.rand(100, 100))

tooltip = fpl.Tooltip()
fig._overlay_scene.add(tooltip._world_object)  # this will of course not have to be done

tooltip.register_graphic(sg)
tooltip.register_graphic(ig)

fig.show()

fpl.loop.run()
```

https://github.com/user-attachments/assets/1bcf35fd-224c-4e87-a453-c82b47ee9cec

### DIFF
--- a/fastplotlib/layouts/_engine.py
+++ b/fastplotlib/layouts/_engine.py
@@ -7,7 +7,7 @@ from ._subplot import Subplot
 from ._rect import RectManager
 
 
-class UnderlayCamera(pygfx.Camera):
+class ScreenSpaceCamera(pygfx.Camera):
     """
     Same as pygfx.ScreenCoordsCamera but y-axis is inverted.
 
--- a/fastplotlib/layouts/_figure.py
+++ b/fastplotlib/layouts/_figure.py
@@ -19,7 +19,7 @@ from ._utils import (
 )
 from ._utils import controller_types as valid_controller_types
 from ._subplot import Subplot
-from ._engine import GridLayout, WindowLayout, UnderlayCamera
+from ._engine import GridLayout, WindowLayout, ScreenSpaceCamera
 from .. import ImageGraphic
 
 
@@ -409,9 +409,13 @@ class Figure:
                 canvas_rect=self.get_pygfx_render_area(),
             )
 
-        self._underlay_camera = UnderlayCamera()
-
+        # underlay render pass
+        self._underlay_camera = ScreenSpaceCamera()
         self._underlay_scene = pygfx.Scene()
+
+        # overlay render pass
+        self._overlay_camera = ScreenSpaceCamera()
+        self._overlay_scene = pygfx.Scene()
 
         for subplot in self._subplots.ravel():
             self._underlay_scene.add(subplot.frame._world_object)
@@ -491,6 +495,9 @@ class Figure:
         self._call_animate_functions(self._animate_funcs_pre)
         for subplot in self:
             subplot._render()
+
+        # overlay render pass
+        self.renderer.render(self._overlay_scene, self._overlay_camera, flush=False)
 
         self.renderer.flush()
 
--- a/fastplotlib/tools/_tooltip.py
+++ b/fastplotlib/tools/_tooltip.py
@@ -1,0 +1,177 @@
+from functools import partial
+
+import numpy as np
+import pygfx
+from .. import ScatterGraphic
+
+from ..graphics import LineGraphic, ImageGraphic, TextGraphic
+
+
+class MeshMasks:
+    """Used set the x0, x1, y0, y1 positions of the plane mesh"""
+
+    x0 = np.array(
+        [
+            [False, False, False],
+            [True, False, False],
+            [False, False, False],
+            [True, False, False],
+        ]
+    )
+
+    x1 = np.array(
+        [
+            [True, False, False],
+            [False, False, False],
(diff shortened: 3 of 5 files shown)


fastplotlib/layouts/_engine.py
```diff
@@ -7,7 +7,7 @@
 from ._rect import RectManager
 
 
-class UnderlayCamera(pygfx.Camera):
+class ScreenSpaceCamera(pygfx.Camera):
     """
     Same as pygfx.ScreenCoordsCamera but y-axis is inverted.
```
kushalkolar, line 12: @almarklein would be useful to have a pygfx Camera that is in screenspace with inverted y, pointer events use (0, 0) as the top left corner so it would be useful to have the option.

almarklein: Something like this? https://github.com/pygfx/pygfx/pull/1092

kushalkolar:
@clewis7 ready for review

Screenshot_20250520_164705

Screenshot_20250520_164713

Screenshot_20250520_164720

Screenshot_20250520_164730

Custom tooltips by providing a function that takes the pointer event and outputs a `str`

Screenshot_20250520_164751

fastplotlib/layouts/_figure.py
```diff
@@ -492,6 +530,9 @@ def _render(self, draw=True):
         for subplot in self:
             subplot._render()
 
+        # overlay render pass
+        self.renderer.render(self._overlay_scene, self._overlay_camera, flush=False)
```
kushalkolar, line 534: @almarklein the overlay render pass without the depth buffer doesn't exist yet right? we'll just keep this here for now and update later I guess.

examples/line_collection/line_stack.py
```diff
@@ -28,6 +31,26 @@
     separation=1,  # spacing between lines along the separation axis, default separation along "y" axis
 )
 
+
+def tooltip_info(ev):
+    """a custom function to display the index of the graphic within the collection"""
```
clewis7, line 36:
```suggestion
    """A custom function to display the index of the graphic within the collection."""
```

examples/line_collection/line_stack.py
clewis7, line 1: Seems like this should be listed as something like `custom_tooltip.py`. I think that it would make it easier for users when looking in the gallery to find this behavior.

clewis7: Now I see the file below...this still feels like beyond normal tooltip behavior

fastplotlib/tools/_tooltip.py
clewis7, line 1: I think you should be consistent with your docstring, you have them for some setters and then not others

fastplotlib/tools/_tooltip.py
```diff
@@ -0,0 +1,298 @@
...
+        self._plane.geometry.positions.update_range()
+
+        # line points
+        pts = [[x0, y0], [x0, y1], [x1, y1], [x1, y0], [x0, y0]]
+
+        self._line.geometry.positions.data[:, :2] = pts
+        self._line.geometry.positions.update_range()
+
+    def _event_handler(self, custom_tooltip: callable, ev: pygfx.PointerEvent):
+        """handles the tooltip appear event, determines the text to be set in the tooltip"""
```
clewis7, line 193:
```suggestion
        """Handles the tooltip appear event, determines the text to be set in the tooltip"""
```
Also, always capitalizing doc strings

examples/line_collection/line_stack.py
kushalkolar: we can do that in a later docs revamp

fastplotlib/tools/_tooltip.py
kushalkolar: if the docstring is on the property it shows up on the setter as well, I'll remove them from the setters

fastplotlib/tools/_tooltip.py
```diff
@@ -0,0 +1,298 @@
...
+        self._plane.material.color = color
+
+    @property
+    def outline_color(self):
+        """Get or set outline color using a str or RGB(A) array"""
+        return self._line.material.color
+
+    @outline_color.setter
+    def outline_color(self, color: str | tuple | list | np.ndarray):
+        """Get or set outline color using a str or RGB(A) array"""
```
kushalkolar, line 140:
```suggestion
```

clewis7 (approved):

PR #837, opened 2025-05-26
Implement PolygonSelector

almarklein:
* [x] Get it to draw a polygon.
* [x] Triangulation.
* [x] Selection.
* [x] Decide on interaction model.

image


almarklein:
> add morphing

That's worthy of a separate lib :)

ivoflipse:
You already have gfxmorph 🤣

almarklein:
This is nearly done, but the ux needs some thought and work. I'm thinking something like this:

* User clicks points to create the polygon. Optionally start with existing polygon.
* Once created, vertices can be dragged.
* Clicking on an edge will split it in two, creating an extra vertex.
* Double-clicking on inside the polygon will delete and let the user create a new polygon.

kushalkolar:
> This is nearly done, but the ux needs some thought and work. I'm thinking something like this:
> 
> * User clicks points to create the polygon. Optionally start with existing polygon.
> * Once created, vertices can be dragged.
> * Clicking on an edge will split it in two, creating an extra vertex.

All this sounds good to me! 

> * Double-clicking on inside the polygon will delete and let the user create a new polygon.

I think this should be explicitly called, or via imgui right click menu. Or clicking on the fill area selects the polygon and presing the "Del" key deletes it.

almarklein:
Interaction done! Slightly different than I said earlier:
* Can initialize the tool with an initial selection.
* If the initial selection is empty, enter polygon creation mode.
* Can also call `start_new_polygon()` to clear the current polygon and draw a new one.
* When creating a new polygon, clicking adds a vertex. Clicking on the first vertex again will finish the polygon.
* When created, the vertices can be dragged. When dragged on top of a neighbouring polygon, it will be deleted.
* When clicking/dragging on a line, a new vertex is inserted.

*edit: in the above, a snapping visualization is applied to help the user understand that the vertex will be "merged" with the vertex under the mouse.*

almarklein:
I just learned about [bermuda](https://github.com/napari/bermuda), a triangulation lib implemented in Rust, a spin-off of the Napari project. We could try and see if it helps here. It would mean an extra dependency, so maybe not use it unless we need it. But we can at least do some quick benchmarks and add a note, so that in case we do need more speed, we know how to get it.

fastplotlib/graphics/features/_selection_features.py
```diff
@@ -340,3 +342,109 @@ def set_value(self, selector, value: Sequence[float]):
...
+
+        self._value = value
+
+        if len(value) >= 3:
+            indices = triangulate(value)
+        else:
+            indices = np.zeros((0, 3), np.int32)
+
+        # TODO: Update the fill mesh
+        # selector.fill.geometry.positions = ...
```
kushalkolar: seems like the fill mesh is updated here below?

fastplotlib/graphics/image.py
```diff
@@ -437,3 +441,41 @@ def add_rectangle_selector(
...
+        """
+        Add a :class:`.PolygonSelector`.
+
+        Selectors are just ``Graphic`` objects, so you can manage, remove, or delete them
+        from a plot area just like any other ``Graphic``.
+
+        Parameters
+        ----------
+        selection: List of positions, optional
+            initial points for the polygon
```
kushalkolar: add to docstring: If `None`, start clicking to create the selection

fastplotlib/graphics/line.py
```diff
@@ -288,6 +293,46 @@ def add_rectangle_selector(
...
+        """
+        Add a :class:`.PolygonSelector`.
+
+        Selectors are just ``Graphic`` objects, so you can manage, remove, or delete them from a
+        plot area just like any other ``Graphic``.
+
+        Parameters
+        ----------
+        selection: (float, float, float, float), optional
+            initial (xmin, xmax, ymin, ymax) of the selection
```
kushalkolar: can update by copying pasting the changes from the image docstring

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+from ._base_selector import BaseSelector
+
+
+@dataclass
+class MoveInfo:
+    """Movement info specific to the polygon selector."""
+
+    mode: str
+    index: int
+    snap_index: int
```
kushalkolar, line 27: comment for what this is

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+
+    @property
+    def parent(self) -> Graphic | None:
+        """Graphic that selector is associated with."""
+        return self._parent
+
+    @property
+    def selection(self) -> np.ndarray[float]:
+        """
+        The polygon as an array of 3D points.
```
kushalkolar: add to docstring: `shape is [n_points, 2]` (I'm assuming)

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+        """
+        return self._selection.value.copy()
+
+    @selection.setter
+    def selection(self, selection: np.ndarray[float]):
+        # set (xmin, xmax, ymin, ymax) of the selector in data space
```
kushalkolar:
```suggestion
```

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
-
-        self.edge_color = edge_color
-        self.edge_width = edge_width
-
-        self._move_info: MoveInfo = None
-
-        self._current_mode = None
+        self.geometry = pygfx.Geometry(
+            positions=np.zeros((8, 3), np.float32),
+            indices=np.zeros((8, 3), np.int32),
```
kushalkolar, line 97: can you comment why this initialization shape

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+        )
+        self.geometry.positions.draw_range = 0, 0
+        self.geometry.indices.draw_range = 0, 0
```
kushalkolar, line 102: what do the draw range do?

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+            self.geometry,
+            pygfx.LineMaterial(
+                thickness=edge_thickness, color=edge_color, pick_write=True
+            ),
+        )
+        self._points = pygfx.Points(
+            self.geometry,
+            pygfx.PointsMaterial(size=vertex_size, color=vertex_color, pick_write=True),
+        )
+        self._points.local.z = 0.01  # move it slightly towards the camera
```
kushalkolar: I do this often! Is `0.01` the best way to do it? I usually do `0.5`

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
+            # Maybe initiate a drag
+            if ev.target is self._points:
+                index = ev.pick_info["vertex_index"]
+                self._start_move_mode("drag", index)
+            elif ev.target is self._line:
+                index = ev.pick_info["vertex_index"]
+                if ev.pick_info["segment_coord"] > 0:
+                    index += 1
+                self._insert_polygon_vertex(index, world_pos)
+                self._start_move_mode("drag", index)
```
kushalkolar, line 413: if the target is the fill area, move the entire polygon?

fastplotlib/graphics/selectors/_polygon.py
```diff
@@ -1,152 +1,491 @@
...
-        handlers = {
-            self._add_segment: "click",
-            self._move_segment_endpoint: "pointer_move",
-            self._finish_segment: "click",
-            self._finish_polygon: "double_click",
-        }
-
-        for handler, event in handlers.items():
-            self._plot_area.renderer.remove_event_handler(handler, event)
+        # Are we close to a point that we can snap to?
```
kushalkolar, line 426: what exactly is the snapping :thinking:

kushalkolar:
Sorry for the delay in reviewing this, awesome work thanks a lot! :partying_face: :partying_face: . Some minor comments and questions.

fastplotlib/graphics/selectors/_polygon.py
almarklein: The draw range allows us to draw only part of the buffer, i.e. it allows us to oversize our buffers to avoid creating a new one for every added point.

Added as a comment as well.

fastplotlib/graphics/selectors/_polygon.py
almarklein: It depends on the other z-values in your scene, and on the camera depth range settings. From what I understand FPL has layers that are 1.0 apart. I'll change the value above to 0.1; still close to it's original "layer".

fastplotlib/graphics/selectors/_polygon.py
almarklein: done

fastplotlib/graphics/selectors/_polygon.py
almarklein: The concept of snapping does multiple things:
- preventing the user from creating points that are very close to each-other,
- allowing the user to finish the polygon by connecting to the start-point when in 'create' mode.
- allowing the user to merge points by dragging one onto its neighbour.

Also added this as a comment.

almarklein:
I addressed all comments.

almarklein:
Need to tweak `alpha_mode` for the worldObjects used in the selectors, but I propose we first merge this (if you're happy), and then in #873 I make those adjustments.

kushalkolar:
Thanks! I was traveling this weekend so it slipped through.

almarklein:
> Thanks! I was traveling this weekend so it slipped through.

No problem. My comment was mainly to suggest a sensible order of merging the prs.

PR #838, opened 2025-05-27
Tweaks to adjust to new pygfx blending

almarklein:
For upcoming https://github.com/pygfx/pygfx/pull/1002

May not be complete, can leave open until the pygfx-pr is merged.


fastplotlib/layouts/_figure.py
```diff
@@ -560,6 +560,11 @@ def _render(self, draw=True):
         # draw the underlay planes
         self.renderer.render(self._underlay_scene, self._underlay_camera, flush=False)
 
+        # With new pygfx' blending, the depth buffer is only cleared after each flush, we need a manual depth
+        # clear to erase the depth values set by the underlay.
```
kushalkolar, line 564: what about the overlay?

kushalkolar: What are depth tests? Is there a performance difference between this or clearing again? What's your recommendation, I'm not too familiar with how the depth buffer works in this context, only how it's important for knowing what objects are above what in a 3D scene. Eventually the overlay scene will probably have axes when using an orthographic projection, along with legends etc.

kushalkolar:
I think this should be merged soon after we decide how to deal with the depth buffer for the overlay pass?

almarklein:
There are some more upcoming changes. I'll likely make some more changes this week.

fastplotlib/layouts/_figure.py
almarklein: The depth test makes sure that objects are not occluded by objects that are behind it (but are rendered later).

The depth buffer needs to be cleared before using it, otherwise the depth test makes no sense. Since there is an underlay pass and an overlay pass, we need extra manual clears.

I added a clear before the overlay pass. That's definitely the simplest solution. The only caveat is that if users want access to the depth buffer (of the scene) they can't because the depth buffer will represent the content of the overlay. If we ever want to change that, we can look into that later. There are more upcoming changes in pyfx that will likely make that easier.

almarklein:
Let's merge this, so we keep `main` working. I'll just open a new PR if new stuff arises.

kushalkolar:
Failures are due to https://github.com/pygfx/wgpu-py/issues/723 

And I think due to AA being disabled for tests now? Some diffs below:

diff-rect_frac_layout-rgb

diff-line_cmap_more-rgb

almarklein:
The image tests fail because the downsample filter is changed (to go from hirez to target rez). So indeed related to SSAA. See https://github.com/pygfx/pygfx/pull/1121

almarklein:
>  AA being disabled for tests now

The env var that disables PPAA is for a different kind of aa 😉, which has not been merged yet (at the time of writing): https://github.com/pygfx/pygfx/pull/1117  That env var will make sure that the screenshots don't change when that PR is in effect.

kushalkolar:
> The image tests fail because the downsample filter is changed (to go from hirez to target rez). So indeed related to SSAA. See https://github.com/pygfx/pygfx/pull/1121

So if I update the ground truth screenshots good to go!?

almarklein:
> So if I update the ground truth screenshots good to go!?

I'm half wondering if I want to add a similar env var to override the pixel_ratio for screenshot tests, now that we have to update all screenshots ... I'm sorry it takes so long. Is this stopping other prs from progressing?

kushalkolar:
~This is weird, seems like only the regenerated screenshots with imgui in the environment don't have AA, whereas the environments without imgui do:~

EDIT: ok now I understand how github actions pushes generated screenshots artifacts. If an image hasn't been re-generated it still gets sent because it exists in the repo already. I should do separate dirs for imgui and non-imgui screenshots or filter what's upload to the artifact.

fastplotlib/layouts/_figure.py
kushalkolar: > The only caveat is that if users want access to the depth buffer (of the scene) they can't because the depth buffer will represent the content of the overlay. If we ever want to change that, we can look into that later. There are more upcoming changes in pyfx that will likely make that easier.

do you want to make a note of it in the pygfx issue if you don't have this planned already?

PR #844, opened 2025-05-30
retry sklearn datset downloads if 403 error

kushalkolar:
should fix the http 403 erros on CI, saves us from having to manually restart those jobs


kushalkolar:
@clewis7 ready for review

PR #849, opened 2025-06-01
`@Figure.add_gui` and `append_gui` decorators

### LINKED ISSUE #848: Create imgui edge windows without declaring a class
Sometimes I just want to quickly prototype a slider and it's a lot of boilerplate to make an `EdgeGUI` subclass.

Maybe something like this:

```python
fig = fpl.Figure()

@fig.add_gui(location="right", title="window title", size=300)
def gui():
    if imgui.button("bah"):
        ...
    
```

related #847

kushalkolar:
closes #848


examples/guis/imgui_decorator.py
```diff
@@ -0,0 +1,34 @@
...
+from imgui_bundle import imgui
+
+figure = fpl.Figure(size=(700, 560))
+figure[0, 0].add_line(np.random.rand(100))
+
+
+@figure.add_gui(location="right", title="window", size=200)
+def gui(fig_local):  # figure is the only argument, so you can use it within the local scope of the GUI function
+    if imgui.button("reset data"):
+        fig_local[0, 0].graphics[0].data[:, 1] = np.random.rand(100)
```
clewis7: I think this needs to be "new data" instead of "reset data"

clewis7: Would make it more informative for what the button actually does

clewis7:
I have never seen someone love a decorator more than you lol

clewis7:
I think this is pretty cool

One thing, I'm not sure how I feel about the `ChangeFlag` class 

It seems like a lot of scaffolding just to indicate that the user has interacted with the imgui window...unless I am missing something

kushalkolar:
> One thing, I'm not sure how I feel about the `ChangeFlag` class
> 
> It seems like a lot of scaffolding just to indicate that the user has interacted with the imgui window...unless I am missing something

It's useful when you have a lot of UI elements that could have changed and you want to know if any of them have changed. It's cumbersome to write:

```python
changed1, val1 = imgui.slider(...)
changed2, val2 = imgui.slider(...)
...
changed_n, valn = imgui.slider(...)

if any([changed1, changed2, ... change_d]):
  # recompute

# could also do
if changed1 | changed2 | ... changed_n:
  # recompute
```

Basically the following is invalid python syntax otherwise I would just use the `or` operator `|` with tuple unpacking:

```python
changed |, val1 = imgui.slider(...)
changed |, val2 = imgui.slider(...)
...
changed |, val_n = imgui.slider(...)
```

This is valid without tuple unpacking:

```python
changed = func()
changed |= func2()
```

kushalkolar:
Failures are due to an imgui update

kushalkolar:
superseeded by #1066

PR #868, opened 2025-06-29
separate array logic and graphic logic in `ImageWidget`

kushalkolar:
Basically a re-write of `ImageWidget`.

## `NDImageProcessor`

This is a class that manages one n-dimensional array-like object. An array-like object must have at least 2 dimensions, and can have an unlimited number of more dimensions (but practically I doubt there's use cases for beyond ~5 dimensions, but it will work anyways). It can also process window functions and a finalizer function (previously called "frame_apply").

`NDImageProcessor` has the following important settable properties:
data: the n-dimensional array it manages
n_display_dims: one of 2 | 3, the number of "spatial dimensions". This is used to determine if an `ImageGraphic` or an `ImageVolumeGraphic` should be used. 
rgb: whether it is rgb or not
n_slider_dims: auto-computed on demand based on `n_display_dims` and `rgb`.

`NDImageProcessor.get(indices: tuple[int, ...])` is where all the work is done. It retrieves a 2D or 3D image and applies _independent_ window functions on each dimension where a window_func is defined. Functions and window sizes can be unique across all dims!

Once the window functions are applied, a `finalizer_func()` is applied on the remaining 2 or 3 spatial dims if necessary.

The `NDImageProcessor` also handles computing the histogram.

`NDImageProcessor` can be subclassed to create ones that are specific for different types of arrays, such as dask arrays, and also allow for histogram logic to be unique to each.

## `ImageWidget`

ImageWidget no longer does any compute or processing! It just hands each data array to an `NDImageProcessor`, and therefore keeps a list of all the processors that correspond to each array: `ImageWidget._image_processors`.

ImageWidget just handles graphics and sliders. When the slider indices (managed by `ImageWidget.indices` settable property) change, it just calls `NDImageProcessor.get(new_indices)` on each processor and the result is displayed, that's it :smile: .

When data arrays, `n_display_dims`, or `rgb` changes, dimensions are pushed/popped on `ImageWidget.indices`. For example, if a dimension is no longer present on any of the current arrays in the `ImageWidget` then it pops a dimension from `indices` and a slider will disappear. And vice-versa.

`HistogramLUTTool` was also re-written. It now just manages the graphics side of things. A precomputed histogram must be given to it.

### DIFF
--- a/fastplotlib/graphics/image_volume.py
+++ b/fastplotlib/graphics/image_volume.py
@@ -204,23 +204,17 @@ class ImageVolumeGraphic(Graphic):
         self._vmax = ImageVmax(vmax)
 
         self._interpolation = ImageInterpolation(interpolation)
+        self._cmap_interpolation = ImageCmapInterpolation(cmap_interpolation)
 
-        if self._data.value.ndim == 4:
-            # set map to None for RGB image volumes
-            self._cmap = None
-            self._texture_map = None
-            self._cmap_interpolation = None
+        # use TextureMap for grayscale images
+        self._cmap = ImageCmap(cmap)
+        self._texture_map = pygfx.TextureMap(
+            self._cmap.texture,
+            filter=self._cmap_interpolation.value,
+            wrap="clamp-to-edge",
+        )
 
-        elif self._data.value.ndim == 3:
-            # use TextureMap for grayscale images
-            self._cmap = ImageCmap(cmap)
-            self._cmap_interpolation = ImageCmapInterpolation(cmap_interpolation)
-            self._texture_map = pygfx.TextureMap(
-                self._cmap.texture,
-                filter=self._cmap_interpolation.value,
-                wrap="clamp-to-edge",
-            )
-        else:
+        if self._data.value.ndim not in (3, 4):
             raise ValueError(
                 f"ImageVolumeGraphic `data` must have 3 dimensions for grayscale images, "
                 f"or 4 dimensions for RGB(A) images.\n"
@@ -312,10 +306,9 @@ class ImageVolumeGraphic(Graphic):
         self._mode.set_value(self, mode)
 
     @property
-    def cmap(self) -> str | None:
-        """Get or set colormap name"""
-        if self._cmap is not None:
-            return self._cmap.value
+    def cmap(self) -> str:
+        """Get or set colormap name, only used for grayscale images"""
+        return self._cmap.value
 
     @cmap.setter
     def cmap(self, name: str):
@@ -349,10 +342,9 @@ class ImageVolumeGraphic(Graphic):
         self._interpolation.set_value(self, value)
 
     @property
-    def cmap_interpolation(self) -> str | None:
+    def cmap_interpolation(self) -> str:
         """Get or set the cmap interpolation method"""
-        if self._cmap_interpolation is not None:
-            return self._cmap_interpolation.value
+        return self._cmap_interpolation.value
 
     @cmap_interpolation.setter
     def cmap_interpolation(self, value: str):
--- a/fastplotlib/ui/right_click_menus/_standard_menu.py
+++ b/fastplotlib/ui/right_click_menus/_standard_menu.py
@@ -100,12 +100,6 @@ class StandardRightClickMenu(Popup):
             )
             self.get_subplot().camera.maintain_aspect = maintain_aspect
 
-            change, show_tooltips = imgui.menu_item(
-                "Show tooltips", "", self._figure.show_tooltips
-            )
-            if change:
-                self._figure.show_tooltips = show_tooltips
-
             imgui.separator()
 
             # toggles to flip axes cameras


kushalkolar:
If we add the following properties we can make things way more arbitrary:

dim_names: dict that maps numerical dim index to a dimension name, ex: `{0: "t", 1: "z"}`. Only need to name scrollable dims.

index: dict that maps dim name to current index, ex: `{"t": 123, "z": 5}`

image_dims: `tuple[int, int] | tuple[int, int, int]` numerical dim indices that define the image displayed in the `ImageGraphic` or `ImageVolumeGraphic`.

Also, for `window_funcs`, if a window func is provided for only one dim then it should apply the window only along that dim. If `window_funcs` are defined for multiple dimensions, then apply the window for all provided dims. Can also have a way to define the order, ex. do "t" first and "z" etc. 

Histogram calculation can also be done within this widget anytime the window funcs or frame_apply funcs are changed. We can then just allow force-setting the `HistogramLUTGraph`.

kushalkolar:
Make a few ready-made subclasses of `ImageWidgetArray`, for example `IWA_Dask`, `IWA_Lazy` etc.

In `ImageWidget.__init__()`, parse each data array in the list use the appropriate IWA type based on the data array type.

IW also has a new kwarg `iwa_types` which can explicitly define the IWA type which corresponds to the data array.

Make the window functions more flexible. Allow options kwargs that include the full data array, i.e. IWA.data, and the current index. Useful for things like computing a window function and then using that on the original data or current frame. Example, use a rolling filter on the current window, use this to subtract from the original frame.

kushalkolar:
new kwarg on `ImageWidget`, "adaptive_histogram", `List[bool] | bool` which creates an event handler for `current_index` so that any time the slider is moved or the index is changed it sets the histogram based on the current frame/volume.

kushalkolar:
If the data doesn't change when moving through dims, example: if there's a simple 2D image amongst multiple txy videos, then would be good to handle this efficiently so that Texture doesn't isn't uploaded when it doesn't change.

kushalkolar:
ok so another issue is that you can't use a window function when you have a mix of movies and images (no t dim) :disappointed: . So this separation also needs to fix this issue.

Also need to make it possible for an ImageWidget subplot to be blank, set that data index as `None` and just ignore when scrolling.

kushalkolar:
argument to specify a transpose tuple of size 2 or 3 that is applied on the final image/volume after slicing, window functions, etc. could be useful.

kushalkolar:
ok I think I got a good way to apply window funcs, clean, understandable. 

```python
# array with many slider_dims
a = np.random.rand(25, 20, 8, 4, 100, 100)
print("shape of a:\t\t", a.shape)

# final slice is: (i - w, i, i + w)
# specify window size for each dim
windows = (3, None, 2, None)
# order in which window funcs are applied
order = (0, 2)
# window function for each dim
funcs = (np.mean, None, np.std, None)

# current slider indices
indices = (15, 5, 4, 2)

indexer = list()

for i, w in zip(indices, windows):
    if w is not None:
        s = slice(i - w, i + w, 1)  # start, stop, step
    else:
        s = slice(i, i + 1, 1)
    indexer.append(s)

print("indexer:\t\t", indexer)

a_s = a[tuple(indexer)]

# also provide the user the option to use this window-sliced array in one function where they handle everything
print("indexer applied to a:\t", a_s.shape)

for dim in order:
    f = funcs[dim]

    a = f(a, axis=dim, keepdims=True)
    print(f"shape of a after window func: {f} on dim: {dim} is:\t {a.shape}")
```

kushalkolar:
Thoughts on how sliders should map onto dims:

if we have this array, let's say it's `[t, z, m, n]`:
`[1_000, 30, 100, 100]`

then, do we assume the following smaller array is `[t, m, n]` or `[z, m, n]`? 
`[n, 100, 100]`

Another way to think about this:

We want to allow arbitrary numbers of dims, so if we have the following array with ndim 6:
`[a, b, c, d, m, n]`

`[m, n]` are the image dims (it could also be RGB or volumetric, that stuff is easy to deal with, ignore it for now)

Do we assume a smaller array with ndim 4 is:
`[a, b, m, n]` 

or:

`[c, d, m, n]`

Sliders will be made for `a, b, c, d` so this is important to decide.

One way to thing of it is: right-most are image dims, so slider dims accumulate as soon as image dims finish? 
This would mean we interpret the shorter array above as `[c, d, m, n]`

Another way to think of it is, slider dims accumulate from left to right until they reach the image dims.
This would mean we interpret the above array as `[a, b, m, n]`

EDIT: We have chosen to do `[c, d, m, n]`

FlynnOConnell:
I think accumulating the slider dims from left to right makes sense.  

However, depending on the use-case, I would want to be able to choose between `[a b m n]` and `[c d m n]`. For microscopy, `[t z m n]` is the most popular but I've also seen libraries use `[z t m n]` (e.g. [suite3d](https://github.com/alihaydaroglu/suite3d/blob/032ebf0bbadc1eea5ecba5adaa1b1cb84fc3b1de/suite3d/io/lbmio.py#L80)). These two dim orders behave completely differently and take a lot of fine-tuning in dask/numpy depending on which you use. 

What about defaulting to "left to right" mapping with an optional semantic argument, e.g.: 

```python
ImageWidget(
    data=array,
    image_dims=(-2, -1),  # or just always the last two dims
    slider_dims=(0, 1, 2, 3),  # explicit ordering
    dim_names={0: "t", 1: "z", 2: "c"}  # for imgui slider labels
)
```

kushalkolar:
Another important thing: other than slider dims, nothing else has to be the same between iw arrays. For example, we can allow each individual array in the ImageWidget to have their own window function or frame apply function. Not sure what's the best API for this though.

Maybe the constructor can take a window funca arg, which if it's a list it defines the window func per data array. This would of course be symmetric with the properties.

kushalkolar:
`display_dims` should be a mutable property. When changed from 2 or 3, delete the existing graphic and replace it.

kushalkolar:
ok I got the basics down!

I now realize the histogram is not gonna be trivial :smile: 

We can't sub-sample the original array and then pass it through the window funcs because that won't correspond to windows in the original data. 

I think we need to do something like this:
* if a window function is used for a dim, use a small number of indices from this dim depending on the window size, but at least 5 or 10. Apply the window function with these indices.
* if a dim has no window function, use some number of indices from this dim, but can be somewhat generous.
* if using a `finalizer_func` that's applied before displaying the frame or volume, do not subsample these spatial dims, for example a Gaussian kernel would provide very different results on a full spatial sample vs. a subsample.

I think we can use some of the logic in `utils.subsample_array` and do something like;

1. Create a new func `get_subsample_indices` that returns the indices necessary to subsample the array.
2. Determine a number of sample indices, call `get()` with the indices.
3. Concatenate all these arrays and compute the histogram.

@apasarkar @FlynnOConnell

kushalkolar:
> Thoughts on how sliders should map onto dims:
> 
> if we have this array, let's say it's `[t, z, m, n]`: `[1_000, 30, 100, 100]`
> 
> then, do we assume the following smaller array is `[t, m, n]` or `[z, m, n]`? `[n, 100, 100]`
> 
> Another way to think about this:
> 
> We want to allow arbitrary numbers of dims, so if we have the following array with ndim 6: `[a, b, c, d, m, n]`
> 
> `[m, n]` are the image dims (it could also be RGB or volumetric, that stuff is easy to deal with, ignore it for now)
> 
> Do we assume a smaller array with ndim 4 is: `[a, b, m, n]`
> 
> or:
> 
> `[c, d, m, n]`
> 
> Sliders will be made for `a, b, c, d` so this is important to decide.
> 
> One way to thing of it is: right-most are image dims, so slider dims accumulate as soon as image dims finish? This would mean we interpret the shorter array above as `[c, d, m, n]`
> 
> Another way to think of it is, slider dims accumulate from left to right until they reach the image dims. This would mean we interpret the above array as `[a, b, m, n]`
> 
> EDIT: We have chosen to do `[c, d, m, n]`

ok so both are actually easy to do and this is something that `ImageWidget` can handle before calling `get()` on each of the individual `NDImageViews`:

```python
# consider an example index from all 4 sliders
indices = (100, 15, 5, 3)

# if we have 4 example array which have the following number of slider dims
n_slider_dims = [4, 3, 2, 1]
```

**right -> left ordering**

```python
# get the indices we use for each of the 4 arrays with different numbers of dims
for n in n_slider_dims:
    print(f"{n} slider_dims, right -> left indices are: {indices[-n:]}")
```

out:

```
4 slider_dims, right -> left indices are: (100, 15, 5, 3)
3 slider_dims, right -> left indices are: (15, 5, 3)
2 slider_dims, right -> left indices are: (5, 3)
1 slider_dims, right -> left indices are: (3,)
```

**left -> right ordering**

```python
for n in n_slider_dims:
    print(f"{n} slider_dims, left -> right indices are: {indices[:n]}")
```

out:

```
4 slider_dims, left -> right indices are: (100, 15, 5, 3)
3 slider_dims, left -> right indices are: (100, 15, 5)
2 slider_dims, left -> right indices are: (100, 15)
1 slider_dims, left -> right indices are: (100,)
```

@FlynnOConnell

kushalkolar:
my all time favorite commit in fastplotlib :smile: : https://github.com/fastplotlib/fastplotlib/pull/868/commits/7770ee05a71092ca53c4927613f2c58d8c4ef2ce

FlynnOConnell:
image

beauty lol

kushalkolar:
ok the basics work! :partying_face: 

Needs tests, more manual testing, some cleanup and tweaks and will be ready to go! I anticipate there will be a few growing pains, lots of parsing since this has to deal with a diverse range of complex datasets and compute possibilities so it may take a while to iron out all the bugs.

https://github.com/user-attachments/assets/b2ecbe70-9324-4de7-b5fa-c03e84aacfae

https://github.com/user-attachments/assets/5bae7c8a-552e-4bae-b0cd-551ac5ab96f2

https://github.com/user-attachments/assets/c0e7a1e4-d790-4b79-aeeb-804f4c1ed9f2

@apasarkar @FlynnOConnell @clewis7 gonna need a review from all of you :D

kushalkolar:
And here is some craziness with dims and window funcs :joy: 

```python
import fastplotlib as fpl
import numpy as np

# 3d and 3 sliders
a = np.random.rand(100, 15, 10, 5, 20, 20)

# 3d and 2 sliders
b = np.random.rand(10, 3, 5, 20, 20)

# 3d over time
c = np.random.rand(15, 5, 20, 20)

# 2d and 2 sliders
c2 = np.random.rand(15, 5, 20, 20)

# 2d over time
d = np.random.rand(30, 20, 20)

# 3 d static
e = np.random.rand(50, 50, 50)

# 2d static
f = np.random.rand(50, 50)

iw = fpl.ImageWidget(
    data=[a, b, c, c2, d, e, f],
    cmap="viridis",
    n_display_dims=[3, 3, 3, 2, 2, 3, 2],
    window_funcs=[(np.mean, None, np.max), (np.mean, None), None, (np.mean, np.max), None, None, None],
    window_sizes=[(11, None, 5), (5, None), None, (3, 11), None, None, None],
    figure_kwargs={
        "size": (700, 560),
        "controller_ids": None,
        "cameras": ["3d", "3d", "3d", "2d", "2d", "3d", "2d", "2d", "2d"],
        "controller_types": ["orbit", "orbit", "orbit", "panzoom", "panzoom", "orbit", "panzoom", "panzoom", "panzoom"]
    }
)
iw.show()
```

https://github.com/user-attachments/assets/d492e538-4f6d-4d98-842c-574d8ed535f2

kushalkolar:
Left to do:

* allow a data array to be `None` and allow initializing `ImageWidget` with no data arrays
* make all the various `ImageWidget` properties such as `indices`, `n_display_dims`, `data`, as mutable `UserList` subclass so we can override `__setitem__` and have those changes reflected in the `ImageWidget`. 
* Look into whether there's some type of object in `collections` that we can use to manage all this better.

kushalkolar:
ok so things left to do are:

There is a bug when the number of slider dims varies across arrays, and when we change one of them with more dims the others get affected for 1 index (╯°□°)╯︵ ┻━┻. I might just get rid of the `Indices` class and force requiring full indices to always be supplied. Likewise for `data`.

To reproduce just make an imagewidget with these arrays:

```python
a = np.random.rand(50, 60, 50)
b = np.random.rand(50, 70, 50)
c = np.random.rand(100, 5, 50, 50, 50)
```

Also need to figure out how to initialize with `None`, and allow changing data such that the number of slider dims can change. This will need some thought to be done well. I should just focus on this.

Thoughts: 

I think based on slider dims order something like this might work:

If right -> left (usual order).

If the new array has more `n_slider_dims` than the `ImageWidget.n_sliders`, insert new indices right -> left. Something like:

```python
while new_image_array.n_slider_dims > self.n_sliders:
  self.indices.data.insert(0)
```

Similarly, if the new array `n_sliders_dims` means that `ImageWidget.n_sliders` reduces, pop from left -> right, i.e. pop at index 0, same code as above but with `data.pop()`. For example an `ImageWidget` with 1 static 2D image and 1 2D + time image, and the time image get replaced with a static image.

kushalkolar:
I think all the key features work now! :partying_face: 

todo:
- more testing
- test garbage collection
- document all the abstract logic behind everything
- I think I'll try implementing the fine-grained control over individual dims or arrays again, the bug had before was from something else. It was just much easier to debug with less syntactic sugar in the way. It'll be nice to do:
  - iw.indices[d] = i
  - iw.data[array_ix] = new_data
  - iw.data["subplot-name"] = new_data
  - iw.n_display_dims[array_ix] = 2 | 3

Other cosmetic and quality of life improvements, especially for @FlynnOConnell :smile:

kushalkolar:
Should make `ImageWidgegt.histogram_widget` settable, and it should update the `NDImageProcessors` accordingly and reset the histogram graphics.

Need to implement ImageWidget properties and setters for window func stuff and finalizer func. Each of these should call `ImageWidget._reset`

kushalkolar:
OK can get and set `data`, all window function stuff, and `finalizer_func`, per-data array using the index or subplot name. Likewise dimension `indices` can be get or set on a per-dimension basis.

So all this works:

```python
iw.data[2] = new_array_of_any_dims

# can also use the subplot name
iw.data["pmd"] = new_pmd_array

# likewise for other image processor properties
iw.window_funcs[1] = (None, np.mean)  # mean window function only on dim at index 1
iw.window_sizes[1] = (None, 23)

# similar for dimension indices
iw.indices[1] = 7

# if `slider_dim_names` is set, can also use that:
iw.slider_dim_names = ("t", "z")
iw.indices["t"] = 5
```

otherwise you'd have to do this (you can also do this if necessary)
```python
# set all new data arrays
iw.data = [new_a, new_b]

# set only the 2nd one as new
iw.data = [old_a, new_b]

# this is equivalent to
iw.data[1] = new_b

# same for all other window funcs

# likewise for setting dim indices
# if the current indices are already (10, 0)
iw.indices = (10, 3)  # keep first dim same, change second dim
# the above is identical to this:
iw.indices[1] = 3
```

@FlynnOConnell

FlynnOConnell:
There is a new requirement that the data input must implement `astype`, due to `/graphics/features/_image.py` calling `_fix_data()`, which it wasn't before?

If so, might be helpful to add something like

```python
if hasattr(data, "astype"):

else:
("Input data must implement astype") 
...

``` 

to the `_is_arraylike()`?

kushalkolar:
Histogram stuff I tested which all works:

- all same vals, all zeros, nans, infs

Also tested large ranges such as between 0 - 2^16, 0 - 2^32, 0 - 2^64, 
also: 1 / 2^16 - 2^16, -2^16 - 2^16, tested 2^32 as well with all these.
@apasarkar 

Can also have a mode where the same histgram lut tool just shows the colorbar with the scalebar and the linear region selector can change the vmin, vmax, probably in a future PR.

kushalkolar:
I encountered what may be a bug while making the mcorr widget. `NDImageProcessor.get()` was using a window function on the mcorr array which has dims [t, m, n], and then getting an array of shape (170,) which is really weird. I can't seem to easily reproduce this though so not sure how this happened.

kushalkolar:
We should also try to support arrays with _*dynamic shapes*_, i.e. where **the shape of the array is a function of one of the indices of a dimension** :laughing: .

This may sound weird but it has a very clear usecase, multi-session data! @apasarkar @FlynnOConnell 

For example consider a multi session array with the following shape:

`[s, t, z, m, n]: [session, time, z-plane, m, n]`

Each session may have a different number of timepoints or z planes!

I _think_ we can actually implement this very easily in a future `MultiSessionNDImageProcessor` subclass by just making the `shape` property something like this:

```python
@property
def shape(self) -> tuple[int, ...]:
  match self._last_indices  # last indices tuple that was used on a `get()` call
    case 0: # session 0
      return ... # shape of session 0
    ...
```

The issue with this is that we've written `NDImageProcessor` to specifically have no notion of storing a slice index. The `ImageWidget` is responsible for maintaining the current slice indices and the `NDImageProcessor` just processes and gives it to `ImageWidget`. Can think later about the best way to make this work. 

Maybe when `get()` is called, it uses the session dimension to set the current shape? Will need to see if dropping this in works directly.

kushalkolar:
Allow the spatial func to be a compute shader or a python function.

FlynnOConnell:
@kushalkolar I'm confused at where in the code we expect custom lazy arrays to be called on a set of indices. 

This line is causing me problems: 

https://github.com/fastplotlib/fastplotlib/blob/219ea3cc2c45b5dce8ecdfbdb2b39fe2afaa7e82/fastplotlib/graphics/features/_image.py#L44-L47

.. where `data[:]` is causing my arrays [__getitem__](https://github.com/MillerBrainObservatory/mbo_utilities/blob/9c73c0cd58fb4ea61cf048b836f024940ad4b5e4/mbo_utilities/array_types.py#L1139) to load fully into RAM.

These are 4D `TZXY`, so when `texture = pygfx.Texture(self.value[data_slice], dim=2)` is called, my array's 4D structure errors:

```python
  File "C:\Users\RBO\repos\mbo_utilities\.venv\Lib\site-packages\pygfx\resources\_texture.py", line 145, in __init__
    the_size = size_from_array(view, dim)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\RBO\repos\mbo_utilities\.venv\Lib\site-packages\pygfx\resources\_texture.py", line 591, in size_from_array
    raise ValueError(
ValueError: Can't map shape (71, 14, 448, 448) on 2D tex. Maybe also specify size?
```

In the `ImageGraphic.data` docstring , it mentions data must implement `memoryview()`, but the call to `memoryview` doesn't happen until the `Texture()` definition here: 

https://github.com/fastplotlib/fastplotlib/blob/219ea3cc2c45b5dce8ecdfbdb2b39fe2afaa7e82/fastplotlib/graphics/features/_image.py#L72-L76

The `dim=2` here is throwing me off, so we expect this to always be 2D at this point?

**EDIT**
> ValueError: Can't map shape (71, 14, 448, 448) on 2D tex. Maybe also specify size?

This happened because my array.ndim was returning the wrong value.

Everything is working flawlessly so far!

FlynnOConnell:
@kushalkolar YESYESYESYES THANK YOU 

https://github.com/user-attachments/assets/455c4c8a-9bfa-4150-9442-5fc24bb4e1c9

More to test on Sunday!

FlynnOConnell:
```
if issubclass(processors, NDImageProcessor):
    processors = [processors] * len(data)
```

Gives `TypeError: issubclass() arg 1 must be a class` when I pass a list of processors.

Could do: 

```python
if isinstance(processors, type) and issubclass(processors, NDImageProcessor):
    processors = [processors] * len(data)
```

Works with a single processor:
```python
isinstance(processors[0], type)
Out[1]: True
```

kushalkolar:
More ideas:

- two modes for histograms: 1) update on every frame, 2) global histogram that represents full data. A kwargs or something.

FlynnOConnell:
right-click histogram to change the colormap?

kushalkolar:
I just learned python has builtin sigfig formatting! https://docs.cems.umn.edu/intro/Prt_01_Lssn_04_Using_Strings_and_Print%28%29.html#general-format-notation

apasarkar:
@kushalkolar some thoughts re: the multi-session use case.

I think I understand what you're suggesting above re: arrays with dynamical shapes. Conceptually this would allow us to define a (num_sessions, <other_dimensions>) array. As you scroll through the "num_sessions" axis, you display data from a different session, and this movie can have a different spatial FOV (and number of frames). 

I wonder whether it would be more intuitive to allow the "data" parameter in ImageWidget to be multi-dimensional. For e.g.: 

```

group_1 = [session_1_vid1, session_1_vid2, ..._session_1_vidn1]

group_2 = [session_2_vid2, session_2_vid2, ..., session_2_vidn2]

group_M = [...]

data_arr = [group_1, group_2, ..., group_M]

iw = fpl.ImageWidget(data = [data_arr])

iw.show()               

```

Here, imagewidget would recognize the nested list structure and add an extra scrollable dimension to allow the user to move between groups of arrays. 

This probably also relates to the multi-dimensional-everything vis.

apasarkar:
Re: the parameters for the constructor, I wonder whether we can get rid of n_display_dims by adding a "volume" boolean. This boolean is by default False (in the same way that rgb is). 

By inspecting the values for volume and rgb, the imagewidget constructor can track what the number of display dimensions should be automatically. (2, 3, or 4 in the case where rgb and volume are both true). 

This might be easier for the user (since they are specifying parameters directly related to what they expect to see in the data).

kushalkolar:
ok this is ready to merge into the `ndwidget` branch, but I will do that tomorrow to make sure it goes well :partying_face:

Issue #754, opened 2025-03-11
separate `AlphaFeature` for `PositionsGraphic` subclasses?


kushalkolar:
I often want to change a `cmap` or `colors` on a line or scatter without changing the alpha. We could make alpha an independent feature.

We can make this per-vertex (if not using a uniform buffer) or for the entire graphic.

@clewis7 what do you think?

clewis7:
It is annoying sometimes to be trying to change the colors (especially when I am sampling a random color) to then have to append the alpha to the end lol

clewis7:
Other option is to be more flexible in the argument we allow for setting colors, and if the dim is 3 instead of 4 to assume they want the alpha to remain the same

kushalkolar:
done in #873

PR #873, opened 2025-07-08
Refactor handling of alpha

almarklein:
## Context

Previously, Pygfx automatically distinguished between opaque and transparent fragments on a per-fragment level. With the new blending in Pygfx (current `main`), this distinction will have to be made per object. This means that *somewhere* a decision will have to be made whether an object is opaque or transparent. 

The PyGfx `material.alpha_mode` is the easy way to control this behavior. The two values that make the most sense for Fastplotlib are:
* 'auto': applies blending, and writes depth when `material.opacity` is 1.
* 'blend': applies blending but does not write depth: relies on the correct render order.

This PR uses 'blend' for all WorldObjects. We can use `material.render_queue` to control the order of axes, grids, legends etc, with respect to other objects. This means that they do not write to the depth buffer.

Using 'auto' would also work, but self-intersecting objects like noisy lines or scattered points would be prone to show artifacts in semi-transparent regions (e.g. the aa edge).

If we add support for 3D objects (e.g. meshes) these should probably use `alpha_mode='auto' or 'opaque'.

## Places where users can introduce transparency

(Let me know if I missed one!)

* the graphic's `alpha` argument.
* alpha in colormaps.
* alpha in a uniform color
* alpha in per-point colors.
* alpha in image pixels.

## Ideas (at the start of this PR)

Some ideas, some compatible, some mutually exclusive:

* The `alpha` value passed to the graphic could be used to set `world_object.material.opacity`.
* Also allow RGB colors, and maybe even promote that a bit over RGBA.
* Automatically make a graphic transparent when `alpha < 1`. Easy to do either in fpl or pygfx. Covers quite some use-cases.
  * Somewhat intuitive: people who wonder why their transparent image is opaque may set `alpha` and see that it works then. Then they can set it to 0.999 to fix it.
* For other cases (e.g. images or per-point colors), require setting `alpha_mode`. 
  * A bit harder on users, but easy enough to document?
* Or check the alpha values in images and colorlists, and automatically set the `alpha_mode` when values < 1 are present.
  * Makes it easier for users, but adds complexity and possible source of bugs.
* Default to "dither", which always works. Then users can set `alpha_mode="blend"` for smoother results of transparent objects.

cc @korijn because this is is a topic that touches both fpl and pygfx


examples/gridplot/multigraphic_gridplot.py
```diff
@@ -54,12 +55,7 @@ def make_circle(center, radius: float, n_points: int = 75) -> np.ndarray:
     circles.append(make_circle(center, 5, n_points=75))
 
 # things like class labels, cluster labels, etc.
-cmap_transform = [
-    0, 1, 1, 2,
-    0, 0, 1, 1,
-    2, 2, 8, 3,
-    1, 9, 1, 5
-]
+cmap_transform = [0, 1, 1, 2, 0, 0, 1, 1, 2, 2, 8, 3, 1, 9, 1, 5]
```
kushalkolar: We don't black the examples because it's useful to manually set things like this list sometimes so the user has a better visual understanding.

kushalkolar:
Thanks! Will check this in detail later today, but I wanted to mention this: https://github.com/fastplotlib/fastplotlib/issues/754

We want to make alpha a separate graphic feature and thus an entirely independent Graphic property (in general we're trying to make the constructor as symmetric as possible with the settable properties, just like pygfx) . This would probably be a good time to do that?

examples/gridplot/multigraphic_gridplot.py
almarklein: My IDE autoformats the whole file with Ruff. Maybe I can rewrite this as a list of lists to keep it square.

almarklein:
> We want to make alpha a separate graphic feature and thus an entirely independent Graphic property

Good to know this was already on the agenda! Yeah I think it makes sense both from an API perspective, but also technically; changing a color is fine, but changing an alpha value might affect how you want to blend things.

examples/gridplot/multigraphic_gridplot.py
kushalkolar: It's a 1d list that defines per color lines 🤔, anyways not too important for now, can fix later

Korijn:
> Then they can set it to 0.999 to fix it.

This seems a bit awkward. I foresee users would complain about having to do this.

> * Or check the alpha values in images and colorlists, and automatically set the `alpha_mode` when values < 1 are present.

From your list of options, this seems most straight forward.

> * Default to "dither", which always works. Then users can set `alpha_mode="blend"` for smoother results of transparent objects.

After thinking about this for a while, I guess what I would expect from a 2D plotting library as a user:

* All graphics have alpha_mode="blend"
* Graphics are drawn in the order that I add them to the figure, transparent or not, so the depth buffer is kind of ignored I guess

I'm not sure if fastplotlib also does 3D plotting, but in that case I would expect it to behave as pygfx does naturally.

By the way, I appreciate you calling for my thoughts, so I'm sharing them here, but I really have not been keeping up with fastplotlib's development for a while, so take it with a grain of salt.

kushalkolar:
> > * Or check the alpha values in images and colorlists, and automatically set the `alpha_mode` when values < 1 are present.
> 
> From your list of options, this seems most straight forward.

I'm leaning towards this as well.

> After thinking about this for a while, I guess what I would expect from a 2D plotting library as a user:
> 
> * Graphics are drawn in the order that I add them to the figure, transparent or not, so the depth buffer is kind of ignored I guess

When using an orthographic projection graphics are stacked in z.

> I'm not sure if fastplotlib also does 3D plotting, but in that case I would expect it to behave as pygfx does naturally.

Everything is 3D just like pygfx, perspective camera for everything. We just don't have a high level API for meshes yet, I've rarely used meshes myself so I hope someone else would be able to contributre a high level API for that. Or maybe meshes are so specific that the pygfx level of abstraction is the best for it, I don't know yet.

almarklein:
> When using an orthographic projection graphics are stacked in z.

Do you mean that the objects are offset in their `ob.local.z`?

Is it possible to clearly identify whether the current graphic is in a 2D or 3D scene? If that is the case we could indeed default to `blend`.

> Or maybe meshes are so specific

Haha!. Some Pygfx users consider meshes the "normal object" and think all the non-meshes are special 😉

Korijn:
> When using an orthographic projection graphics are stacked in z.

And what determines the stacking order? Insertion order?

kushalkolar:
> > When using an orthographic projection graphics are stacked in z.
> 
> Do you mean that the objects are offset in their `ob.local.z`?

Yup! 

> > When using an orthographic projection graphics are stacked in z.
> 
> And what determines the stacking order? Insertion order? 

Yup. Though the z can be set when added, or later at any time. 

> Is it possible to clearly identify whether the current graphic is in a 2D or 3D scene? If that is the case we could indeed default to `blend`.

If camera.fov == 0

You can switch between them and it's something I do all the time when looking at low dimensional projections with lines and scatters.

examples/gridplot/multigraphic_gridplot.py
almarklein: fixed

almarklein:
Tests pass again! (or am I missing something?)

almarklein:
oh lol, all tests are skipped 😆

Korijn:
How are you feeling about this refactor @almarklein, now that you can experience the revamped transparency system as a user?

almarklein:
> How are you feeling about this refactor @almarklein, now that you can experience the revamped transparency system as a user?

I'm feeling pretty good about it in general.

I felt rather disappointed that we cannot abstract the whole transparency thing away from the user. But after having done quite a bit of research on the topic I'm confident that even with the more advanced methods (which would consume more memory and/or be slower) it'd still not be fool proof. That's what it is.

So the only way to go is to shove the problem on the user 😄 , but as gentle as possible, and providing multiple different options to deal with it. I think we've now now done that pretty well, providing pretty powerful options, but also easy-to-use shorthands with `alpha_mode`.

As for this PR in Fastplotlib, I think it helps a lot if we can determine/assume the scene to be 2D, so we can just use 'blend'. If no object writes depth, Pygfx will not create the depth buffer, so it'd be a bit lighter too.

kushalkolar:
> if we can determine/assume the scene to be 2D, so we can just use 'blend'. If no object writes depth, Pygfx will not create the depth buffer, so it'd be a bit lighter too

3D is used often, easiest way to check is if `camera.fov > 0` 

This will also need some refactoring of the graphic features, will add more thoughts soon

Korijn:
You may want to consider introducing an Enum or a dict with a list of custom render queues, just so they're easier to organize and manage.

almarklein:
Ok, the approach I implemented now:

* Background and overlays use `alpha_mode=='blend'`; we don't really need the depth buffer there.
* Same for annotations like the selector tools, but these are rendered in the same scene so I used `render_queue=3500` so they always render later, and they don't use the depth buffer (no write and no test).
* For these underlays and overlays I used 'ob.render_order` to define the order of rendering within that group, instead of using z.
* Axes are solid, except for the text labels, for which we want aa.
* Lines and Points have `alpha_mode='auto'` by default, but user can change with an arg. The `aa` is turned on if the alpha mode is 'blend'.

The screenshot tests fail. obviously. What's the best way to update these? Same as in Pygfx?

> This will also need some refactoring of the graphic features, will add more thoughts soon.

What kind of changes?

examples/gridplot/multigraphic_gridplot.py
```diff
@@ -70,10 +71,10 @@ def make_circle(center, radius: float, n_points: int = 75) -> np.ndarray:
 figure["circles"].add_line_collection(
     circles,
     cmap="tab10",
-    cmap_transform=cmap_transform,
+    cmap_transform=sum(cmap_transform, []),
```
kushalkolar:
```suggestion
    cmap_transform=np.asarray(cmap_transform).ravel()
```

simpler?

fastplotlib/graphics/selectors/_base_selector.py
```diff
@@ -10,6 +10,11 @@
 from .._base import Graphic
 
 
+# For selectors we use a render_queue of 3500, which is at the end of what is considered the group of transparent objects.
+# So it's rendered later than the normal scene (2000 - 3000), but before overlays like legends and tooltips.
+render_queue = 3500
```
kushalkolar: Do you think it would be more organized to put all the `render_queue` positions in `fastplotlib.utils` so we have one place to see them and get them from?

fastplotlib/graphics/selectors/_rectangle.py
```diff
@@ -203,9 +241,10 @@ def __init__(
         )  # left line, right line, bottom line, top line
 
         # add the edge lines
+
         for edge in self.edges:
-            edge.world.z = -0.5
-            group.add(edge)
+            edge.render_order = 1
```
kushalkolar, line 247: ah so don't need to worry about z for the primitives that are used to build a selector tool?

fastplotlib/layouts/_frame.py
```diff
@@ -171,16 +171,28 @@ def __init__(
...
+            depth_test=False,
+            pick_write=True,
+        )
         self._plane = pygfx.Mesh(geometry, material)
         wobjects.append(self._plane)
 
-        # otherwise text isn't visible
-        self._plane.world.z = 0.5
+        # Plane gets rendered before text and point
+        self._plane.render_order = -1
```
kushalkolar, line 196: nice! The text will look nicer now :smile:

fastplotlib/graphics/selectors/_base_selector.py
kushalkolar: And Korijn's suggestion:

> You may want to consider introducing an Enum or a dict with a list of custom render queues, just so they're easier to organize and manage.

kushalkolar:
>     * Background and overlays use `alpha_mode=='blend'`; we don't really need the depth buffer there.
> 
>     * Same for annotations like the selector tools, but these are rendered in the same scene so I used `render_queue=3500` so they always render later, and they don't use the depth buffer (no write and no test).
> 
>     * For these underlays and overlays I used 'ob.render_order` to define the order of rendering within that group, instead of using z.
> 

Nice this is a much nicer way to handle the ordering!

> The screenshot tests fail. obviously. What's the best way to update these? Same as in Pygfx?

Similar:

```python
WGPU_FORCE_OFFSCREEN=1 REGENERATE_SCREENSHOTS=1 pytest -v examples/
FASTPLOTLIB_NB_TESTS=1 REGENERATE_SCREENSHOTS=1 pytest --nbmake examples/notebooks/
```

> > This will also need some refactoring of the graphic features, will add more thoughts soon.
> 
> What kind of changes?

I've see you've already done most of them in [`_positions_graphics.py`](https://github.com/fastplotlib/fastplotlib/pull/873/files#diff-4bffcf43e8d4c24f1f1581927a835e78e374439d2a2352c82117edde3adc375e), mainly removing the alpha handling from the colors feature.

I think the last things to do are:

* Make a simple [common `GraphicFeature`](https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib/graphics/features/_common.py) so that `alpha` is just like other graphic properties and evented. I think we can set `alpha_mode` automatically when the alpha changes? 
* Move the [docstring for `alpha` and `alpha_mode`](https://github.com/fastplotlib/fastplotlib/pull/873/files#diff-b037d7d63ca4cce71b583b7191a46b184a3cd314587c590fd45e54a4ff8ecd3eR59-R75) to the [Graphic base class](https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib/graphics/_base.py#L75-L99) ?

fastplotlib/graphics/selectors/_base_selector.py
almarklein: Made an enum. I indeed much cleaner.

fastplotlib/graphics/selectors/_rectangle.py
almarklein: Indeed. Using `render_order` is a bit simpler, since it applies only to the same objects in the same render_queue. I.e. we can define the order of objects without iterfering with other objects.

almarklein:
I implemented a `GraphicsFeature` for `alpha` and `alpha_mode`. This indeed helps to make the code cleaner (less duplication). 

However, setting the alpha / alpga_mode in `figure[0, 0].add_scatter(.. , alpha=0.07)` does not work. While it does work for `visible` which is also a graphics feature. Any idea why? I may have missed some init step or something.

almarklein:
> However, setting the alpha / alpga_mode in figure[0, 0].add_scatter(.. , alpha=0.07) does not work. While it does work for visible which is also a graphics feature. Any idea why? I may have missed some init step or something.

Fixed; these were set in `_set_world_object()`

almarklein:
Also updated screenshots. Two examples have a diff above the threshold: image_widget_single_video and image_widget_videos. I tried copying the screenshots from the `screenshots` CI build, but the screenshots are the same (git status is empty). Any ideas?

kushalkolar:
> Also updated screenshots. Two examples have a diff above the threshold: image_widget_single_video and image_widget_videos. I tried copying the screenshots from the `screenshots` CI build, but the screenshots are the same (git status is empty). Any ideas?

`ImageWidget` screenshots aren't being generated, that regeneration action is failing: https://github.com/fastplotlib/fastplotlib/actions/runs/18008358652/job/51234040316?pr=873 

Must be because newer wgpu uses newer imgui, so I'm going to try and put the fastplotlib imgui option to just point to `"wgpu[imgui]"`, will see if that works.

kushalkolar:
updated to latest imgui, uses the new texture registering method, uses a merged font so regular text and font awesome icons can be used together without push/pop font. @almarklein I think one last thing is to set the render order so that axes lines and points are on top when `camera.fov == 0`, but not on top of any overlays:

This is `examples/gridplot/gridplot.py`

image

If it doesn't cause a performance hit you could put a check for `camera.fov` in [`Axes.update`](https://github.com/fastplotlib/fastplotlib/blob/f7634bdbe2d027a447dae8866e3e2339ba2893b3/fastplotlib/graphics/_axes.py#L515):

```python
        if self._plot_area.camera.fov == 0:
            # set render order for ruler line, points, text
        else:
            ...
```

A better way might be if it's possible to add an event handler when camera.fov changes?

kushalkolar:
All tests against pygfx@main are passing so I will merge, can do the axes thing in another PR, #903 

The tests that are failing are all against the latest release of pygfx, which does not work with the latest imgui so it's fine to skip.

PR #876, opened 2025-07-10
Imgui pin to attempt bugfix release

kushalkolar:
I think https://github.com/pygfx/wgpu-py/issues/723 will take a while to resolve and then wait for a wgpu release before we can release fpl v0.6.0. Also once we merge #838 into main then we have to wait for the next release of pygfx before we can release fpl v0.6.0. So let's see if we can do a bugfix release for v0.5 right now to prevent users running into imgui font errors


kushalkolar:
@clewis7 gtg, all release tests passing, will do release once you merge

clewis7 (approved):

PR #896, opened 2025-09-06
add animation func getter and `clear_animations()` method

### LINKED ISSUE #826: public property to return animation functions, method to clear animation functions

clewis7:
closes #826

### DIFF
--- a/fastplotlib/layouts/_figure.py
+++ b/fastplotlib/layouts/_figure.py
@@ -543,6 +543,10 @@ class Figure:
         """show/hide tooltips for all graphics"""
         return self._show_tooltips
 
+    @property
+    def animation_funcs(self) -> list[callable]:
+        return self._animate_funcs_pre + self._animate_funcs_post
+
     @show_tooltips.setter
     def show_tooltips(self, val: bool):
         self._show_tooltips = val
@@ -763,6 +767,14 @@ class Figure:
             self._animate_funcs_pre.remove(func)
 
         if func in self._animate_funcs_post:
+            self._animate_funcs_post.remove(func)
+
+    def clear_animations(self):
+        """Removes all animation functions from both pre and post render."""
+        for func in self._animate_funcs_pre:
+            self._animate_funcs_pre.remove(func)
+
+        for func in self._animate_funcs_post:
             self._animate_funcs_post.remove(func)
 
     def clear(self):
--- a/fastplotlib/layouts/_plot_area.py
+++ b/fastplotlib/layouts/_plot_area.py
@@ -272,6 +272,10 @@ class PlotArea(GraphicMethodsMixin):
         """1, 2, or 4 colors, each color must be acceptable by pygfx.Color"""
         self._background_material.set_colors(*colors)
 
+    @property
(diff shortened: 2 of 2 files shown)


clewis7:
@kushalkolar ready for review

fastplotlib/layouts/_figure.py
```diff
@@ -543,6 +543,10 @@ def show_tooltips(self) -> bool:
         """show/hide tooltips for all graphics"""
         return self._show_tooltips
 
+    @property
+    def animation_funcs(self) -> list[callable]:
```
kushalkolar: might be useful to instead return a dict of lists `{"pre": list[callables], "pos": list[...]}`, or a tuple `(pre_list, post_list)` ? 

and docstring

fastplotlib/layouts/_figure.py
```diff
@@ -765,6 +769,14 @@ def remove_animation(self, func):
         if func in self._animate_funcs_post:
             self._animate_funcs_post.remove(func)
 
+    def clear_animations(self):
```
kushalkolar: takes a string kwarg to remove only "pre" or "post", default is `None` which clears both?

kushalkolar (changes requested):

fastplotlib/layouts/_figure.py
```diff
@@ -765,6 +770,37 @@ def remove_animation(self, func):
         if func in self._animate_funcs_post:
             self._animate_funcs_post.remove(func)
 
+    def clear_animations(self, type: str = None):
```
kushalkolar: Generally don't want to use `type` since it's a built-in function in python, suggest renaming to "kind" or something else.

clewis7:
@kushalkolar ready for another review :D

clewis7:
@kushalkolar I rebased with main, should be good to go now if you want to give it another look

fastplotlib/layouts/_figure.py
```diff
@@ -543,6 +543,11 @@ def show_tooltips(self) -> bool:
         """show/hide tooltips for all graphics"""
         return self._show_tooltips
 
+    @property
+    def animation_funcs(self) -> dict[str, list[callable]]:
```
kushalkolar:
```suggestion
    def animations(self) -> dict[str, list[callable]]:
```

let's just call it `animations` so it's consistent with `add_animations` and `clear_animations`, and the rest of the API is generally like this too

fastplotlib/layouts/_plot_area.py
```diff
@@ -274,6 +274,11 @@ def background_color(self, colors: str | tuple[float]):
         """1, 2, or 4 colors, each color must be acceptable by pygfx.Color"""
         self._background_material.set_colors(*colors)
 
+    @property
+    def animation_funcs(self) -> dict[str, list[callable]]:
```
kushalkolar:
```suggestion
    def animations(self) -> dict[str, list[callable]]:
```

clewis7:
I think these will pass, but I might need to update the API docs also

clewis7:
@kushalkolar should be gtg :D

PR #904, opened 2025-09-29
thinner lines for Linear and LinearRegion selectors, fully transparent thick outer lines as hitboxes

kushalkolar:
After the alpha refactor by @almarklein this is finally possible! I think this looks much better, the width of the outer transparent line is a kwarg `hitbox_width`

https://github.com/user-attachments/assets/985da684-2585-4b1b-929d-e82be0956d87

https://github.com/user-attachments/assets/99bb7692-b270-40fe-a5d6-2a117fa51ea2

### DIFF
--- a/examples/selection_tools/linear_selector.py
+++ b/examples/selection_tools/linear_selector.py
@@ -2,7 +2,8 @@
 Linear Selectors
 ================
 
-Example showing how to use a `LinearSelector` with lines and line collections.
+Example showing how to use a `LinearSelector` with lines and line collections. The linear selector is the yellow
+vertical line.
 """
 
 # test_example = true
--- a/examples/selection_tools/linear_selector_image.py
+++ b/examples/selection_tools/linear_selector_image.py
@@ -2,8 +2,9 @@
 Linear Selectors Image
 ======================
 
-Example showing how to use a `LinearSelector` to selector rows or columns of an image. The subplot on the right
-displays the data for the selector row and column.
+Example showing how to use a `LinearSelector` to select rows or columns of an image. The subplot on the right
+displays the data for the selector row and column. Move the selectors independently or click the middle mouse
+button to move both selectors to the clicked location.
 """
 
 # test_example = false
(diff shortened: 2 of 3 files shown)


fastplotlib/graphics/selectors/_linear.py
```diff
@@ -111,6 +112,9 @@ def __init__(
         edge_color: str | tuple | np.ndarray, default "w"
             color of the selector
 
+        hitbox_width: float, default 14.0
+            the width around the selector which is responsive to mouse events
```
almarklein:
```suggestion
            the width around the selector which is responsive to mouse events, in logical pixels.
```

fastplotlib/graphics/selectors/_base_selector.py
```diff
@@ -282,6 +284,13 @@ def _move_start(self, event_source: WorldObject, ev):
         """
         position = self._plot_area.map_screen_to_world(ev)
 
+        # if the event source was an outer transparent line, get the
+        # corresponding inner line since it's just a proxy
+        if event_source in self._outer_edges:
+            print("outer edge pointer down")
```
almarklein:
```suggestion
```

fastplotlib/graphics/selectors/_linear.py
almarklein: I think I would call it *extra* width`.

Another option is to think of `hitbox_width` as a margin and doing `thickness=thickness + 2 * hitbox_width`. But I think the current implementation is easier to explain as an extra width on top of the normal width.

fastplotlib/graphics/selectors/_linear_region.py
```diff
@@ -215,6 +219,25 @@ def __init__(
                 pick_write=True,
             ),
         )
+
+        line0_outer = pygfx.Line(
+            pygfx.Geometry(
+                # share buffer with inner line so they can both be managed together
+                positions=line0.geometry.positions
+            ),
```
almarklein, line 227: Can it not share the whole geometry?

kushalkolar:
@almarklein do you think this is overall the right approach here after your alpha refactor?

fastplotlib/graphics/selectors/_linear_region.py
kushalkolar: ah I never thought of that, for me I see geometry and all I think of is the storage buffer on the GPU

clewis7:
docs are failing? and some of the tests should be passing no?

kushalkolar:
all non-release tests passing

kushalkolar:
Merge conflict in the lfs files O_O. I'm going to copy this over to a new PR before we discover novel ways to get git lfs smudge errors, @FlynnOConnell

kushalkolar:
superseeded by #914, I don't want to risk a git lfs error again with a merge conflict in an lfs pointer (how tf does that even happen?!? :fearful: )

Issue #839, opened 2025-05-28
Dynamically place axes on camera near clipping plane 


kushalkolar:
@almarklein wouldn't this always keep the axes visible?

almarklein:
Good to have this discussion. I used the trick to move the axes towards the camera, but that's not bulletproof, and can result into problems with depth clipping (as we've seen).

One idea I had was to specify (in your docs) that the axes are rendered at a specific depth, e.g. `100`. So that users can make sure that their data is below 100. Or above, if they want something on top of the axes.

A simpler approach could be to set `material.depth_test=False` and `ob.render_order=1000`.

PR #906, opened 2025-10-02
Refactor z-order to ensure axes are on top

### LINKED ISSUE #903: Ensure axes are on top with orthographic projections
Not sure what the best way is, maybe setting the z so that it's right on the camera's near clipping plane?

This is `examples/gridplot/gridplot.py`

image

almarklein:
Closes #903

* [x] Graphics are no longer offset in depth, because the depth may actually mean something.
* [x] Except ImageGraphics, they are now pushed back (older further away).
* [x] Lines, points, and axes use `depth_campare="<="` so that for objects at the same depth, the render order determines which one is on top.  
* [x] Set `render_queue`  of axes so they are rendered later than all the other objects.

This should make it that axes and graphics interact as expected in both 3D and 2D scenes. Note that in a 2D view (panzoom controller), if a scatter graphic has data that is nonzero for z, it can indeed cover the axes, but this is expected.

### DIFF
--- a/fastplotlib/graphics/_axes.py
+++ b/fastplotlib/graphics/_axes.py
@@ -183,14 +183,22 @@ class Axes:
         }
 
         # create ruler for each dim
-        self._x = pygfx.Ruler(alpha_mode="solid", **x_kwargs)
-        self._y = pygfx.Ruler(alpha_mode="solid", **y_kwargs)
-        self._z = pygfx.Ruler(alpha_mode="solid", **z_kwargs)
+        self._x = pygfx.Ruler(
+            alpha_mode="solid", render_queue=RenderQueue.axes, **x_kwargs
+        )
+        self._y = pygfx.Ruler(
+            alpha_mode="solid", render_queue=RenderQueue.axes, **y_kwargs
+        )
+        self._z = pygfx.Ruler(
+            alpha_mode="solid", render_queue=RenderQueue.axes, **z_kwargs
+        )
 
         # We render the lines and ticks as solid, but enable aa for text for prettier glyphs
         for ruler in self._x, self._y, self._z:
+            ruler.line.material.depth_compare = "<="
+            ruler.points.material.depth_compare = "<="
+            ruler.text.material.depth_compare = "<="
             ruler.text.material.alpha_mode = "auto"
-            ruler.text.material.render_queue = RenderQueue.auto + 50
             ruler.text.material.aa = True
 
         self._offset = offset
--- a/fastplotlib/graphics/image.py
+++ b/fastplotlib/graphics/image.py
@@ -324,9 +324,6 @@ class ImageGraphic(Graphic):
 
         self._plot_area.add_graphic(selector, center=False)
 
-        # place selector above this graphic
-        selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
-
         return selector
 
     def add_linear_region_selector(
@@ -402,9 +399,6 @@ class ImageGraphic(Graphic):
 
         self._plot_area.add_graphic(selector, center=False)
 
-        # place above this graphic
-        selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
-
         return selector
 
     def add_rectangle_selector(
@@ -447,9 +441,6 @@ class ImageGraphic(Graphic):
 
         self._plot_area.add_graphic(selector, center=False)
 
-        # place above this graphic
-        selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
-
         return selector
 
     def add_polygon_selector(
@@ -484,8 +475,5 @@ class ImageGraphic(Graphic):
         )
 
         self._plot_area.add_graphic(selector, center=False)
-
-        # place above this graphic
-        selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
 
         return selector
--- a/fastplotlib/layouts/_plot_area.py
+++ b/fastplotlib/layouts/_plot_area.py
@@ -10,6 +10,7 @@ from rendercanvas import BaseRenderCanvas
 
 from ._utils import create_controller
 from ..graphics._base import Graphic
+from ..graphics import ImageGraphic
 from ..graphics.selectors._base_selector import BaseSelector
 from ._graphic_methods_mixin import GraphicMethodsMixin
 from ..legends import Legend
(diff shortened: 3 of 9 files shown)


almarklein:
One thing that makes this tricky is that although a lot of examples / use-cases are 2D, it's hard to determine reliably when we can assume the scene to be 2D. Therefore I went with an approach that more or less assumes 3D but also works well in 2D.

kushalkolar:
Older images pushed back, so this means if I add a new image it will appear on top of the existing ones? This is desired and most intuitive, and is similar with other plotting libs.

If I add a like and then scatter points, these points will appear on top of the line?

Usually the scene is 3D if the camera.fov > 50

Btw there was this related ideas before: https://github.com/fastplotlib/fastplotlib/issues/839

almarklein:
> if I add a new image it will appear on top of the existing ones? 

Exactly. And yet the top image will still be below the points, lines, and axes.

> If I add a like and then scatter points, these points will appear on top of the line?

Yes. But here it's a bit more tricky, since (assuming the data is 2D) they are both at `z=0`, and depth-fighting may be a thing. 

There are a few reasons why depth-fighting should be unlikely:

* The depth range for ortho projection is usually favorable, so we have plenty of precision in the depth buffer.
* The use of `depth_compare = '<='` avoids depth-fighting for two objects at exactly the same depth.

For reference, other measures we can take for depth fighting:

* Use the `invariant` keyword for the `position` in the vertex shader (in PyGfx), which reduces differences due to float roundoff errors as a cause for depth-fighting.
* Use depth offset, e.g. assign an increasing depth offset to the graphics, in the order they were added.

> Usually the scene is 3D if the camera.fov > 50

Usually ... but it's perfectly sensible for sciency 3D data to be rendered using orthographic projection. Whether pan-zoom is used (and looking down z) would be a more appropriate signal. In any case, I really like the idea to create the scene in a way that works for both 2D and 3D cases, so users can switch and all keeps making sense. I think we can make that work.

> Btw there was this related ideas before: https://github.com/fastplotlib/fastplotlib/issues/839

In the comment in #839 I still assumed the scene was always 2d :)

almarklein:
So the `release` tests are know to fail - we need a new release of PyGfx. 

But the other tests keep failing with:

image

I had a look at the diffs, and it looks like cases where previously the data was over the axes, and now it's under 👌 

I tried to update the reference images by downloading them from the `screenshots` action, and replacing them. I am a little confused about the "screenshots-" vs "screenshots-imgui" artifacts, which both seem to contain the same screenshots?

I think I tried both, but to no avail...

examples/screenshots/nb-lines-data.png
kushalkolar, line 1: Were you testing these from outside the notebook in a new example which is how they got into this dir? The nb test screenshots are in the `examples/notebooks/screenshots` dir. You could make a new example to test that graphics added after get overlaid.

examples/screenshots/line_collection_slicing.png
kushalkolar, line 1: If individual axes are added to a graphic I think we'd want them to be above the graphic? This is done here: https://github.com/fastplotlib/fastplotlib/blob/5a9b8299b7f85d8551fa85b146d49a523e39f999/fastplotlib/graphics/_base.py#L509-L518

examples/screenshots/scatter_colorslice_iris.png
kushalkolar, line 1: I'm wondering why the colors of a few points have changed, those pixels are probably occupied by multiple datapoints. I think additive blending might fix the issue for this use case?

fastplotlib/graphics/_axes.py
```diff
@@ -183,14 +183,22 @@ def __init__(
...
+        )
+        self._z = pygfx.Ruler(
+            alpha_mode="solid", render_queue=RenderQueue.axes, **z_kwargs
+        )
 
         # We render the lines and ticks as solid, but enable aa for text for prettier glyphs
         for ruler in self._x, self._y, self._z:
+            ruler.line.material.depth_compare = "<="
+            ruler.points.material.depth_compare = "<="
+            ruler.text.material.depth_compare = "<="
```
kushalkolar, line 200: What exactly does this mean for the axes?

fastplotlib/graphics/image.py
```diff
@@ -324,9 +324,6 @@ def add_linear_selector(
 
         self._plot_area.add_graphic(selector, center=False)
 
-        # place selector above this graphic
-        selector.offset = selector.offset + (0.0, 0.0, self.offset[-1] + 1)
-
```
kushalkolar, line 329: If I understand correctly, this is no longer necessary because rendering the selectors on top is handled by the render queue and render order properties?

kushalkolar:
>  Lines, points, and axes use depth_campare="<=" so that for objects at the same depth, the render order determines which one is on top.

Can you explain this a bit? I also made a issue on pygfx to understand `depth_compare` better so maybe I'll wait for that.

fastplotlib/layouts/_plot_area.py
```diff
@@ -394,6 +395,29 @@ def remove_animation(self, func):
...
+        building a 3D scene or not. We could check whether the 3d dimension of line/point data
+        is all zeros, but maybe this is intended, and *other* graphics in the same scene
+        may be actually 3D. We could check camera.fov being zero, but maybe the user
+        switches to a 3D camera later, or uses a 3D orthographic camera.
+
+        The one exception, kindof, is images, which are inherently 2D, and for which
+        layering helps a lot to get things rendered correctly. So we basically layer the
+        images, in the order that they were added, pushing older images backwards (away
+        from the camera).
+        """
```
kushalkolar, line 411: I like this 😄!

fastplotlib/legends/legend.py
kushalkolar, line 1: the legend needs a lot of work which we'll get to later, it will be in an overlay render pass so let's ignore it for now.

examples/screenshots/nb-lines-data.png
almarklein: I downloaded the screenschots from the `screenshots` job, and copied them to the wrong folder. Will fix.

examples/screenshots/line_collection_slicing.png
almarklein: This already happens in the `Axes` object itself, because it has a high `render_queue`.

examples/screenshots/scatter_colorslice_iris.png
almarklein: Will update its alpha_mode

fastplotlib/graphics/_axes.py
almarklein: It means that the axes (which lives at z=0) is drawn over objects that are also drawn at z=0. With the default `depth_compare` "<", objects at z=0 would obfuscate the axes.

fastplotlib/graphics/image.py
almarklein: Exactly! 🚀

almarklein:
>>   Lines, points, and axes use depth_campare="<=" so that for objects at the same depth, the render order determines which one is on top.
>
> Can you explain this a bit? I also made a issue on pygfx to understand depth_compare better so maybe I'll wait for that.

See https://github.com/pygfx/pygfx/issues/1213#issuecomment-3384420886

PR #910, opened 2025-10-09
Adjust for PyGfx's new ticks

almarklein:
A lot of screenshots in fpl will need to be updated because of [this](https://github.com/pygfx/pygfx/pull/1209) upstream change.

By changing tick_marker and tick_size, we can also get this, let me know what you prefer:

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ keywords = [
 requires-python = ">= 3.10"
 dependencies = [
     "numpy>=1.23.0",
-    "pygfx==0.13",
+    "pygfx==0.14",
     "wgpu",          # Let pygfx constrain the wgpu version
     "cmap>=0.1.3",
     # (this comment keeps this list multiline in VSCode)


fastplotlib/graphics/_axes.py
```diff
@@ -161,26 +161,36 @@ def __init__(
...
-            x_kwargs = dict()
-
-        if y_kwargs is None:
-            y_kwargs = dict()
-
-        if z_kwargs is None:
-            z_kwargs = dict()
+        x_kwargs = x_kwargs or {}
+        y_kwargs = y_kwargs or {}
+        z_kwargs = z_kwargs or {}
```
kushalkolar, line 166: Nice and compact, however I thought explicit is better then implicit? Is there an advantage to using this expression?

fastplotlib/graphics/_axes.py
```diff
@@ -161,26 +161,36 @@ def __init__(
...
+            tick_size=8.0,
+            line_width=2.0,
+            color="#fff",
+        )
 
-        x_kwargs = {
-            "tick_side": "right",
+        x_kwargs = dict(
+            tick_side="right",
+            tick_marker="tick",  # 'tick' or 'tick_left' make sense
```
kushalkolar: I think tick marker can go into generic for each?

kushalkolar:
Nice! I think keep the both sided ticks

fastplotlib/graphics/_axes.py
almarklein: Yes, if we keep the double-side tick, that's cleaner. Will update.

fastplotlib/graphics/_axes.py
almarklein: No, just compactess 🤷

fastplotlib/graphics/_axes.py
kushalkolar: from the screenshot seems like you have the double sided ticks already?

kushalkolar: nvm I can't read

kushalkolar:
Let's merge this and then rebase #906 on main, and then update screenshots later? Because the selector line thickness has changed in #904 and has to be merged too which changes screenshots :sweat_smile: . Or create a temporary "dev" branch.

almarklein:
> Let's merge this and then rebase https://github.com/fastplotlib/fastplotlib/pull/906 on main, and then update screenshots later?

I mostly created this to update the screenshots due to the new PyGfx 😅, so that the new screenshots in #906 were only for the changes in that pr. But yeah, we should get things merged one way or another. 

Regarding the screenshots, one thing I don't understand is that that in a screenshots-build like [this one](https://github.com/fastplotlib/fastplotlib/actions/runs/18389693563), there's two artifacts, which both contain all screenshots that are also in the repo (via LFS). Which one should I pick?

BTW: Is the imgui bump in Pygfx / wgpu a problem for FPL?

kushalkolar:
> Regarding the screenshots, one thing I don't understand is that that in a screenshots-build like [this one](https://github.com/fastplotlib/fastplotlib/actions/runs/18389693563), there's two artifacts, which both contain all screenshots that are also in the repo (via LFS). Which one should I pick?

So one is supposed to have screenshots that have imgui elements (such as the toolbar), and the other is supposed to have screenshots without imgui elements (to make sure that everything still functions correctly without the optional imgui install). But now I realize that the imgui screenshots still get uploaded even in the non-imgui artifact because they exist in that dir. So in the screenshot regenerate action I should add a `rm examples/screenshots` before generating the screenshots. 

> BTW: Is the imgui bump in Pygfx / wgpu a problem for FPL?

Nope! I updated the fonts changes here: https://github.com/fastplotlib/fastplotlib/pull/873/commits/f1746be3046dd2b83a0346bf92c2e293a10c125d 

And now in optional dependencies I let wgpu handle imgui version: https://github.com/fastplotlib/fastplotlib/blob/6a0d3b37eac4658754a8d4766a4175ebf960d754/pyproject.toml#L60

I wonder if it would make sense to do this for pygfx too?

almarklein:
> I wonder if it would make sense to do this for pygfx too?

PyGfx does not depend on imgui; only some examples do.

kushalkolar:
> > I wonder if it would make sense to do this for pygfx too?
> 
> PyGfx does not depend on imgui; only some examples do.

It's an optional dependency: 

https://github.com/pygfx/pygfx/blob/37cdd2664020c52380e44fc37c43129882b76f4d/pyproject.toml#L43

Might be better to let wgpu handle it?

almarklein:
Ah, I see what you mean now. Yeah, would be nice to use `wgpu[imgui]` in `pygfx/pyproject.toml` too.

PR #913, opened 2025-10-11
Scatter markers and other scatter features

kushalkolar:
After https://github.com/pygfx/pygfx/pull/1208 we can have per-vertex scatter markers :tada: . After this PR I think we cover the full space of the pygfx points API. 

- [x] `VertexMarkers` buffer manager for fancy indexing of markers
- [x] `UniformMarker` graphic feature
- [x] `PointsRotations` buffer manager
- [x] `UniformEdgeColor`
- [x] `EdgeWidth`
- [x] `UniformRotations`
- [x] sprites implementing by passing an array to `ScatterGraphic(image=...)`
- [x] gaussian and basic material
- [x] tests
  - [x] scatter point features
  - [x] `VertexMarkers` buffer manager
  - [x] `PointsRotations` buffer manager
- [x] examples
- ~I think search sorted can be used to map user marker -> "standard" marker string, not just the ints which is already done,~ let's skip this until we hit performance issues for a use case. 

This PR also makes a few other changes to graphic features that aren't specific to scatters:

- `property_name` is now an instance attribute instead of a class attribute. And the value of the `type` field of the emitted event dict comes from the `property_name` instance attribute. This let's us use the `VertexColors` buffer manager for _both_ per-point vertex face colors AND edge colors and keep their property_name different when used within the same graphic. Similar thing for `TextureArray`

@almarklein the only usecase of per-vertex rotations is for points right, I don't see how that would make sense for any other type of object (ex: lines). 

The `PointsMaterial` that is chosen depends on the marker mode (basic, marker shapes, image (sprites), or gaussian). By default it will use `PointsMarkerMaterial` because distinct markers are the most common scientific use case. Once an object has been instantiated you can't switch between the materials. I think let's keep it that way unless someone has a usecase, I can't think of one @almarklein @clewis7 . 

I wonder what the purpose of the basic `PointsMaterial` is when we have markers. @almarklein have you benchmarked performance?

### DIFF
--- a/examples/scatter/scatter_image_as_points.py
+++ b/examples/scatter/scatter_image_as_points.py
@@ -33,7 +33,7 @@ figure = fpl.Figure(size=(700, 350))
 
 scatter = figure[0, 0].add_scatter(
     data=sine,
-    mode="image",  # mode must be "image", otherwise
+    mode="image",  # mode must be "image", otherwise the `image` arg is ignored and markers are used
     image=array,
     cmap="jet",  # the image is multiplied by the scatter point colors if provided
     sizes=25,


almarklein:
> @almarklein the only usecase of per-vertex rotations is for points right, I don't see how that would make sense for any other type of object (ex: lines).

Yes it's purpose if to rotate the marker; the little image that we draw at the position of the point. So very specific to the PointsMarkerMaterial even.

almarklein:
> Once an object has been instantiated you can't switch between the materials. I think let's keep it that way unless someone has a usecase, I can't think of one

Someday, a user will come by with a use-case 😉   But I agree that in this context its fine to select at instantiation. And ppl can always modify the `graphic.world_object`, right?

kushalkolar:
> > Once an object has been instantiated you can't switch between the materials. I think let's keep it that way unless someone has a usecase, I can't think of one
> 
> Someday, a user will come by with a use-case 😉   But I agree that in this context its fine to select at instantiation. And ppl can always modify the `graphic.world_object`, right?

Yea I could implement it but I don't want to do it now. For example if you go from markers to sprites, you have to clear the geometry markers buffer and anything related to markers and then create a new texture. Lots of things to keep track of.

kushalkolar:
OK I need to change `GraphicFeature.property_name` from a class attribute to an instance attribute so we can `VertexColors` buffer manager can be used for scatter point edge colors too.

kushalkolar:
basics work, but I just realized a `str` is also a `Sequence` so I need to undo a lot of what I thought was cleanup :facepalm: 

image

messing around with rotations :joy: 

https://github.com/user-attachments/assets/ee65dc95-9a76-46ed-bc74-face052f3328

kushalkolar:
Almost done, just need to write tests for the new buffer managers.

kushalkolar:
Added all scatter tests (I think) and all scatter backend tests passing :partying_face: . 

@clewis7 r4r, and I can make the ground truth screenshots for this PR after rebasing on main oncee #919 is reviewed and merged!

kushalkolar:
Need to update the garbage collection test nb because depending on the scatter mode some features can be `None`. :thinking:

kushalkolar:
It's going to fail with the release version until https://github.com/pygfx/pygfx/pull/1219 is in the release.

examples/scatter/scatter_image_as_points.py
```diff
@@ -0,0 +1,55 @@
...
+])
+
+# load an image of Almar's cat
+wikkie = np.flipud(iio.imread("imageio:wikkie.png"))
+
+figure = fpl.Figure(size=(700, 350))
+
+scatter = figure[0, 0].add_scatter(
+    data=sine,
+    mode="image",  # mode must be "image", otherwise
```
clewis7: otherwise...

kushalkolar: I forgot to finish my thought :laughing:

clewis7 (approved):

PR #914, opened 2025-10-11
thinner selectors

kushalkolar:
After the alpha refactor by @almarklein this is finally possible! I think this looks much better, the width of the outer transparent line is a kwarg `hitbox_width`

https://github.com/user-attachments/assets/985da684-2585-4b1b-929d-e82be0956d87

https://github.com/user-attachments/assets/99bb7692-b270-40fe-a5d6-2a117fa51ea2

superseeds #904

PR #915, opened 2025-10-11
remove existing screenshots in regenerate action

### LINKED ISSUE #912: screenshot artifacts
> Regarding the screenshots, one thing I don't understand is that that in a screenshots-build like [this one](https://github.com/fastplotlib/fastplotlib/actions/runs/18389693563), there's two artifacts, which both contain all screenshots that are also in the repo (via LFS). Which one should I pick?

So one is supposed to have screenshots that have imgui elements (such as the toolbar), and the other is supposed to have screenshots without imgui elements (to make sure that everything still functions correctly without the optional imgui install). But now I realize that the imgui screenshots still get uploaded even in the non-imgui artifact because they exist in that dir. So in the screenshot regenerate action I should add a rm examples/screenshots before generating the screenshots.

kushalkolar:
closes #912

@almarklein I think this will fix it right?

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -51,6 +51,9 @@ jobs:
         env:
           PYGFX_EXPECT_LAVAPIPE: true
         run: |
+          # delete existing screenshots
+          rm ./examples/screenshots/*.png
+          rm ./examples/notebooks/screenshots/*.png
           # regenerate screenshots
           RENDERCANVAS_FORCE_OFFSCREEN=1 REGENERATE_SCREENSHOTS=1 pytest -v examples
       - name: Generate screenshots notebook, exclude image widget


kushalkolar:
yup this works! Now there are no `no-imgui` screenshots in the non-imgui regeneration artifact :laughing: . And only the standard with imgui screenshots in the imgui regeneration artifact. 

@clewis7 should be gtg, tests will not pass because a few PRs have to be merged/have been merged that change most screenshots so I'll do it all at once.

kushalkolar:
@clewis7 let's merge this first and then I can rebase #914 on main so it has the nice cleaner screenshots artifact, and I will replace the screenshots on that PR.

almarklein:
Thanks for solving this 😃

PR #919, opened 2025-10-19
New screenshots :D 

kushalkolar:
After the flurry of changes that affect "viewed z-order", axes order, axes ticks, thinner selectors from #906 #910 #914


kushalkolar:
@clewis7 gtg

PR #920, opened 2025-10-19
`ImageWidget.set_data()` allows setting only a subset of the data arrays

kushalkolar:
Skips data arrays that already exist.

going to rebase on main after #919


kushalkolar:
accidentally pull from origin into this

PR #924, opened 2025-10-20
skip existing arrays in ImageWidget.set_data()

### LINKED ISSUE #901: `ImageWidget.set_data()` for a subset of the data arrays
```patch
Subject: [PATCH] ImageWidget.set_data() for some of the data arrays
---
Index: fastplotlib/widgets/image_widget/_widget.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/fastplotlib/widgets/image_widget/_widget.py b/fastplotlib/widgets/image_widget/_widget.py
--- a/fastplotlib/widgets/image_widget/_widget.py	(revision df922879c143b4079ee4dd1b5903d7784e7f95f2)
+++ b/fastplotlib/widgets/image_widget/_widget.py	(date 1758359542366)
@@ -929,6 +929,11 @@
         for i, (new_array, current_array, subplot) in enumerate(
             zip(new_data, self._data, self.figure)
         ):
+            # if the new array is the same as the existing array, skip
+            # this allows setting just a subset of the arrays in the ImageWidget
+            if new_data is self._data[i]:
+                continue
+
             # check last two dims (x and y) to see if data shape is changing
             old_data_shape = self._data[i].shape[-self.n_img_dims[i] :]
             self._data[i] = new_array

```

kushalkolar:
Skips data arrays that already exist, closes #901 

@clewis7 should be gtg

### DIFF
--- a/fastplotlib/widgets/image_widget/_widget.py
+++ b/fastplotlib/widgets/image_widget/_widget.py
@@ -929,6 +929,11 @@ class ImageWidget:
         for i, (new_array, current_array, subplot) in enumerate(
             zip(new_data, self._data, self.figure)
         ):
+            # if the new array is the same as the existing array, skip
+            # this allows setting just a subset of the arrays in the ImageWidget
+            if new_data is self._data[i]:
+                continue
+
             # check last two dims (x and y) to see if data shape is changing
             old_data_shape = self._data[i].shape[-self.n_img_dims[i] :]
             self._data[i] = new_array


kushalkolar:
> This is cool! I didn't know you could do this kind of comparison without `np.all()`

This is only checking that the object reference is the same. If "new_data" is an array with the same values as the existing data array then it won't pass this check.

Intended use is something like this:

```python
a, b, c, d # existing arrays

iw = fpl.ImageWidget([a, b, c, d])

# I want to replace only c

new = np.array(...)

iw.set_data([a, b, new, d])

# this will also end up replacing

a_copy = a.copy()

iw.set_data([a_copy, b, c, d])
```

PR #925, opened 2025-10-20
fewer github actions for CI

### LINKED ISSUE #922: test only 2 python versions?
We have a ton of github actions because we test python 3.11, 3.12, 3.13, all combinations of optional dependencies, and Linux & Mac OSes. We can cut down on this by testing on only python 3.11 and 3.13 (upper and lower bounds of official support), like what `pygfx` does. @clewis7 @almarklein thoughts? The main issue is that it takes a long time for all the actions to run.

kushalkolar:
closes #922 

This will reduce the number of actions we have to run, while still testing reasonably.

* test only on the boundaries of officially supported versions, right now python3.11 and python3.13
* Run screenshot tests only on the most recent version, right now python3.13

### DIFF
--- a/.github/workflows/ci-pygfx-release.yml
+++ b/.github/workflows/ci-pygfx-release.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI-pygfx-release
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


kushalkolar:
@clewis7 r4r

PR #929, opened 2025-10-21
VectorsGraphic

### LINKED ISSUE #560: Vector graphic
Would be useful to represent vectors as arrows instead of points. Might also be useful to define what basis this vector lies in.

kushalkolar:
closes #560 

Uses instanced meshes to create the 3D vector representations and place them in 3D space. 

- [x] `VectorsGraphic` graphic which is based on instanced meshes built by merging a cone and cylinder mesh. 
  - [x] `VectorDirections` which manages the positions of the arrows through modifying the translation elements of the transform matrix of the instance buffer
  - [x] `VectorDirections` which manages the directions of the arrows through modifying the entire transform matrix of the instance buffer
  - [x] supports static immutable single colors for now
  - [x] auto-determines size of the arrows by estimating the density of vectors, or user can provide a specific size or specific shape for the arrows.
 
Can implement per-vector colors after https://github.com/pygfx/pygfx/issues/1221 later in a future PR. 

image

Can do some fun electromagnetism animations with this.

Very simple API:

```python
figure = fpl.Figure(size=(700, 700))

start, stop, step = 0, 2 * np.pi, 0.2

# get uniform x, y positions
x, y = np.meshgrid(
    np.arange(start, stop, step),
    np.arange(start, stop, step)
)

# vectors, u and v indicate directions in x and y
u = np.cos(x)
v = np.sin(y)

# positions of each vector as [n_points, 2] array
positions = np.column_stack([x.ravel(), y.ravel()])

# directions of each vector as a [n_points, 2] array
directions = np.column_stack([u.ravel(), v.ravel()])

vectors = figure[0, 0].add_vectors(
    positions=positions,
    directions=directions,
)

figure.show()
```

### DIFF
--- a/fastplotlib/graphics/features/_vectors.py
+++ b/fastplotlib/graphics/features/_vectors.py
@@ -1,5 +1,3 @@
-from typing import Sequence
-
 import numpy as np
 import pylinalg as la
 


fastplotlib/graphics/vector_field.py
```diff
@@ -0,0 +1,366 @@
...
+            # get quaternion to rotate existing vector direction to new direction
+            rotation = la.quat_from_vecs(start_rot, self._directions[i])
+            # get the new transform
+            transform = la.mat_compose(
+                self._positions.value[i], rotation, magnitudes[i]
+            )
+            # set the buffer
+            world_object.instance_buffer.data["matrix"][i] = transform.T
+
+        world_object.instance_buffer.update_full()
```
kushalkolar: @almarklein this look right to you? The start position of the vector is pointing towards the viewer, so `[0, 0, 1]`.

fastplotlib/graphics/vector_field.py
```diff
@@ -0,0 +1,366 @@
...
+        geometry = create_vector_geometry(spacing=spacing, color=color, **shape_options)
+        material = pygfx.MeshBasicMaterial()
+        n_vectors = self._positions.value.shape[0]
+
+        world_object = pygfx.InstancedMesh(geometry, material, n_vectors)
+
+        range_3d = np.mean(np.ptp(self._positions[:], axis=0))
+        magnitudes = (
+            np.linalg.norm(self.directions[:], axis=1, ord=2) / range_3d
+        ) * size_scaling_factor
```
kushalkolar: @almarklein let me know if you have an idea on how to auto-determine the scale size better

fastplotlib/graphics/features/_vectors.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            self._positions[:] = value
+
+        for i in range(self._positions.shape[0]):
+            # only need to update the translation vector
+            graphic.world_object.instance_buffer.data["matrix"][i][3, 0:3] = (
+                self._positions[i]
+            )
+
+        graphic.world_object.instance_buffer.update_full()
```
kushalkolar, line 92: @almarklein just want to be sure this looks right 😄

fastplotlib/graphics/features/_vectors.py
```diff
@@ -0,0 +1,178 @@
...
+
+        for i in range(self._directions.shape[0]):
+            # get quaternion to rotate existing vector direction to new direction
+            rotation = la.quat_from_vecs(old_directions[i], self._directions[i])
+            # get the new transform
+            transform = la.mat_compose(graphic.positions[i], rotation, magnitudes[i])
+            # set the buffer
+            graphic.world_object.instance_buffer.data["matrix"][i] = transform.T
+
+        graphic.world_object.instance_buffer.update_full()
```
kushalkolar, line 184: @almarklein just want to be sure this looks right 😄

kushalkolar (commented):
few questions

fastplotlib/graphics/features/_vectors.py
```diff
@@ -0,0 +1,178 @@
...
+    def set_value(self, graphic, value: np.ndarray):
+        if value.shape[0] != self._positions.shape[0]:
+            raise ValueError(
+                f"number of vector positions in passed array != number of vectors in graphic: "
+                f"{value.shape[0]} != {self._positions.shape[0]}"
+            )
+
+        if value.shape[1] == 2:
+            # assume 2d
+            self._positions[:, :-1] = value
```
almarklein, line 81: This cannot happen, since above you check that the shapes are equal, right?

fastplotlib/graphics/features/_vectors.py
almarklein: Yes, the update_full is appropriate; the values that are updated are so close together that the chunking will do it anyway. But doing `update_full` bypasses the chunking algorithm (not waste cpu cycles on that).

fastplotlib/graphics/features/_vectors.py
```diff
@@ -0,0 +1,178 @@
...
+
+        else:
+            self._positions[:] = value
+
+        for i in range(self._positions.shape[0]):
+            # only need to update the translation vector
+            graphic.world_object.instance_buffer.data["matrix"][i][3, 0:3] = (
+                self._positions[i]
+            )
```
almarklein, line 90: Half wondering if you can avoid the for-loop by doing this with broadcasting ... ? I'm not sure if that works with structured arrays.

fastplotlib/graphics/features/_vectors.py
almarklein: 👌

fastplotlib/graphics/vector_field.py
```diff
@@ -0,0 +1,366 @@
...
+            cone_height_divisor=4.0,
+            stalk_radius_divisor=30.0,
+            stalk_height_divisor=4.0,
+        )
+
+        if vector_shape_options is None:
+            vector_shape_options = {}
+
+        for k in vector_shape_options:
+            if k not in shape_options:
```
almarklein:
```suggestion
        if any(k not in shape_options for k in vector_shape_options):
```

almarklein: Or something like:
```py
unknown_shape_keys = set(vector_shape_options.keys()) - set(shape_options.keys())
if unknown_shape_keys:
     ...
```

fastplotlib/graphics/vector_field.py
```diff
@@ -0,0 +1,366 @@
...
+            color of the vectors
+
+        vector_shape_options: dict
+            dict with the following fields that describe the shape of the vector arrows.
+            Larger values decrease the size of each component.
+
+                * cone_radius_divisor, default 10.0
+                * cone_height_divisor, default 4.0
+                * stalk_radius_divisor, default 30.0
+                * stalk_height_divisor, default 4.0
```
almarklein: What unit/space are these numbers? Relates to the question about sizing. Maybe a fraction relative to the full length of the vector?

fastplotlib/graphics/vector_field.py
almarklein: I can see how this is tricky ... 

Thinking about this as a naive user, I would want be able to provide the exact size of the vector, in world coordinates. Where the size is the (total) length of the vector. 

I find the `spacing` argument is a bit confusing. Also since there seem to be 2 params that affect the scale, but both work in an indirect way, it seems. I think it may work better if it was an explicit thing (this IS the size of the arrows) and then maybe a tip in the docstring to use 50% of the approximate distance between neighboring arrows (or something like that).

kushalkolar: Yea I might just do that. I was trying to see if I could come up with an automated way, give your field and it displays it in the best way.

almarklein: Could still allow the param to be None and then autodetermine it based on the "density" of the field ...

fastplotlib/graphics/features/_vectors.py
kushalkolar: positions are `[n, 3]`. The above checks for `shape[0]`, this here checks for `shape[1]` 😄

fastplotlib/graphics/features/_vectors.py
kushalkolar: The start vector is actually always `[0, 0, 1]` since that's the base transform on the mesh instance and the transform we apply is always w.r.t. the base transform:

```python
        for i in range(self._directions.shape[0]):
            # get quaternion to rotate existing vector direction to new direction
            rotation = la.quat_from_vecs(np.array([0, 0, 1]), self._directions[i])
            # get the new transform
            transform = la.mat_compose(graphic.positions[i], rotation, magnitudes[i])

            # set the buffer
            graphic.world_object.instance_buffer.data["matrix"][i] = transform.T

        graphic.world_object.instance_buffer.update_full()
```

Tested with a simple sine and cosine field:

https://github.com/user-attachments/assets/3b1c2637-6aa5-4682-90f5-92a940dd582d

fastplotlib/graphics/features/_vectors.py
kushalkolar: fast enough for now, I can revisit later

fastplotlib/graphics/vector_field.py
kushalkolar: Now it sets the size of a magnitude 1 vector by estimating the field density in all provided dims, or uses the user's specific size in world space.

kushalkolar:
Interactions work well:

https://github.com/user-attachments/assets/9ad0f329-d077-4c46-97b4-64b8ec57c398

```python
figure = fpl.Figure(size=(700, 700))

start, stop, step = 0, 2 * np.pi, 0.4

# get uniform x, y positions
x, y = np.meshgrid(np.arange(start, stop, step), np.arange(start, stop, step))

# vectors, u and v are x and y components indication directions
u = np.cos(x)
v = np.sin(y)

# positions of each vector as [n_points, 2] array
positions = np.column_stack([x.ravel(), y.ravel()])
# directions of each vector as a [n_points, 2] array
directions = np.column_stack([u.ravel(), v.ravel()])


vector_field = figure[0, 0].add_vector_field(
    positions=positions,
    directions=directions,
)

def update(change):
    val = change["new"]
    start, stop, step = 0, 2 * np.pi, 0.4
    start += val
    stop += val
    
    # get uniform x, y positions
    x, y = np.meshgrid(np.arange(start, stop, step), np.arange(start, stop, step))
    
    # vectors, u and v are x and y components indication directions
    u = np.cos(x)
    v = np.sin(y)
    
    directions = np.column_stack([u.ravel(), v.ravel()])
    
    vector_field.directions = directions

slider = FloatSlider(value=0, min=-5, max=5, step=0.01)
slider.observe(update, "value")

VBox([figure.show(), slider])
```

Need to wait for https://github.com/pygfx/pylinalg/pull/107
and https://github.com/pygfx/pygfx/pull/1220

before merging into fastplotlib main

fastplotlib/graphics/vector_field.py
kushalkolar: I did exactly this :smile:

kushalkolar:
The recursion error with the ome-zarr file makes no sense, going to get rid of it from tests and only show the code in the docs until they fix whatever is up.

examples/image_volume/image_volume_multi_channel.py
clewis7, line 1: This example will not run locally for me

clewis7: Also, seems unrelated to this PR

examples/vectors/vectors_swirl.py
clewis7, line 1: I think it would be cooler if the controller type were "fly"

pyproject.toml
clewis7, line 1: We should do a release after this one is merged. It will also be good to have a recent release before gov. meeting later this month

examples/vectors/vectors_interact_electric_charges.py
clewis7, line 1: This one is really cool!

examples/vectors/vector_field_interact_charges.py
```diff
@@ -0,0 +1,181 @@
+"""
+View Electric Field
+===================
+
+Interactively move the charges around by clicking and dragging the mouse to see
+the static field with the charges at their new positions. This is just computing
+static fields, not electrodynamics or magnetic field effects are taken into account.
```
clewis7:
```suggestion
static fields, no electrodynamics or magnetic field effects are taken into account.
```

examples/vectors/vector_field_interact_charges.py
```diff
@@ -0,0 +1,181 @@
...
+    colors=colors,
+    sizes=1,
+    edge_width=0.1,
+    uniform_edge_color=False,
+    alpha=0.7,
+    size_space="model",
+    metadata={"charges": charges},  # you can store anything as arbitrary metadata
+    alpha_mode="blend",
+)
+particles.world_object.render_order
```
clewis7: Code comment here for what this line is doing

examples/vectors/vector_field_interact_charges.py
```diff
@@ -0,0 +1,181 @@
...
+# interactivity code, very similar to the "Drag points" example
+is_moving = False
+particle_index = None
+# interact with particles by moving them with mouse
+@particles.add_event_handler("pointer_down")
+def start_drag(ev: pygfx.PointerEvent):
+    global is_moving
+    global particle_index
+
+    if ev.button != 1:
```
clewis7: Code comment for this check

fastplotlib/graphics/_vectors.py
```diff
@@ -0,0 +1,391 @@
...
+    @property
+    def directions(self) -> VectorDirections:
+        """Vector directions"""
+        return self._directions
+
+    @directions.setter
+    def directions(self, new_directions):
+        self._directions.set_value(self, new_directions)
+
+
```
clewis7, line 173: Can you add a docstring to the three functions below? Even if they are mostly just for internal use, it would be helpful for me to better understand what each of these is doing.

fastplotlib/graphics/_vectors.py
```diff
@@ -0,0 +1,391 @@
...
+        #     self._positions = positions
+        # else:
+        #     self._positions = VectorPositions(positions)
+        #
+        # if isinstance(directions, VectorDirections):
+        #     self._directions = directions
+        # else:
+        #     self._directions = VectorDirections(directions)
+
+        self._positions = VectorPositions(positions)
```
clewis7, line 87: Where are we checking that the number of positions is the same as the number of directions?

clewis7 (changes requested):

examples/image_volume/image_volume_multi_channel.py
kushalkolar: the reason it's modified in this PR is because it blocked CI and docs gen from running.

What do you get on your end? I also posted on `ome-zarr`: https://github.com/ome/ome-zarr-py/issues/505

pyproject.toml
kushalkolar: yes!

examples/vectors/vector_field_interact_charges.py
kushalkolar: this is doing nothing but  I was trying to get the particles to draw on top of the vectors but something is wrong :joy:

fastplotlib/graphics/_vectors.py
kushalkolar: I just copy pasted from pygfx but can add some

examples/vectors/vectors_swirl.py
kushalkolar: I think people usually look at large overviews of vector fields so being able to rotate and get the full-field view is more common than flying through it.

examples/vectors/vector_field_interact_charges.py
kushalkolar: ok I figured this out :D

examples/vectors/vector_field_interact_charges.py
kushalkolar: done

fastplotlib/graphics/_vectors.py
kushalkolar: added a few

fastplotlib/graphics/_vectors.py
kushalkolar: right I forgot this, the shapes should match actually.

examples/vectors/vectors_interact_electric_charges.py
kushalkolar: simulating changing fields and the induced magnetic field is more complicated than I thought, been much years since I did vector calc lol . Will do another time.

examples/image_volume/image_volume_multi_channel.py
clewis7: For me, it will just never load the figure even though I think url is being resolved

clewis7 (approved):
Aside from that one example that won't run, everything LGTM

PR #930, opened 2025-10-23
Fix docstring

kushalkolar:
I didn't notice that the docs didn't build in #913


kushalkolar:
@clewis7 gtg

kushalkolar:
also removed edge color from the spinning spiral example so it's more visible.

clewis7:
Sorry, I didn't notice either!

PR #938, opened 2025-11-12
better plot scaling logic

### LINKED ISSUE #783: Customizable camera linking does not work
Not sure why, pygfx demo works, fpl is doing something additional I think:

```python
import numpy as np
import fastplotlib as fpl
import pygfx

xs = np.linspace(0, 2 * np.pi, 100)
ys = np.sin(xs)

ys_big = np.random.rand(100) * 10

# create cameras, fov=0 means Orthographic projection
camera1 = pygfx.PerspectiveCamera(fov=0)
camera2 = pygfx.PerspectiveCamera(fov=0)

# create controllers, first add the "main" camera for the subplot
controller1 = pygfx.PanZoomController(camera1)
controller2 = pygfx.PanZoomController(camera2)

# add the other camera to each controller, but only include the 'x' state, i.e. 'y' for height is not included
# this must be done only after adding the "main" cameras to the controller as done above
controller1.add_camera(camera2, include_state={"x", "width"})
controller2.add_camera(camera1, include_state={"x", "width"})

# create figure using these cameras and controllers
figure = fpl.Figure(
    shape=(2, 1),
    cameras=[camera1, camera2],
    controllers=[controller1, controller2],
    size=(700, 560)
)


figure[0, 0].add_line(np.column_stack([xs, ys_big]))
figure[1, 0].add_line(np.column_stack([xs, ys]))

print(controller1._cameras)
print(controller2._cameras)

for subplot in figure:
    subplot.camera.zoom = 1.0

figure.show(maintain_aspect=False, autoscale=False)
```

I did make the following change to `PlotArea`:

```patch
Index: fastplotlib/layouts/_plot_area.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/fastplotlib/layouts/_plot_area.py b/fastplotlib/layouts/_plot_area.py
--- a/fastplotlib/layouts/_plot_area.py	
+++ b/fastplotlib/layouts/_plot_area.py	(date 1743405901011)
@@ -79,7 +79,9 @@
         self._camera = camera
         self._controller = controller
 
-        self.controller.add_camera(self._camera)
+        if self.camera not in self.controller.cameras:
+            self.controller.add_camera(self._camera)
+
         self.controller.register_events(
             self.viewport,
         )
```

### LINKED ISSUE #821: Share axes between subplots
I'm using fastplotlib to visualize a set of timeseries that has a few hundred sensors measuring two distinct variables with different units. This works fine with two scatter subplots stacked vertically, one for each variable.

Since the x-axis has identical range between my subplots (but the y-axis is very different), it would be really nice to have a linked pan/zoom/etc. between them. This corresponds exactly to the "sharex" (or "sharey") arguments that can be passed to plt.subplots() in matplotlib.

**Is there a way to achieve this in the current version of fastplotlib?**

I'm also interested to know if there is a way to get something resembling timestamps on the x-axis (but I would guess this is future work).

kushalkolar:
When one controller manages multiple cameras in a Figure, i.e. when multiple subplot all share a controller, the current implementation of `auto_center()` and `auto_scale()` sets the cameras of those _other subplots_ based on the bounding box of the scene in the _current subplot_. This is a problem when the scales are very different between subplots! 

This PR now sets the cameras of the _other subplots_ based on the other subplot's own scene bbox.

closes #783 , closes #821

Example with the same code from #783 

Notice how the y-scales are very different.

https://github.com/user-attachments/assets/cb70b395-fc2c-48ae-b82f-95c2fa463dd4

@clewis7 can we get this merged ASAP please :smile: 

This shouldn't break any existing functionality, the screenshot tests will tell.

### DIFF
--- a/examples/tests/test_examples.py
+++ b/examples/tests/test_examples.py
@@ -100,29 +100,10 @@ def test_example_screenshots(module, prep_environment):
     # import the example module
     example = import_from_path(module.stem, module)
 
-    if fpl.IMGUI:
-        # there doesn't seem to be a resize event for the manual offscreen canvas
-        example.figure.imgui_renderer._backend.io.display_size = (
-            example.figure.canvas.get_logical_size()
-        )
-        # run this once so any edge widgets set their sizes and therefore the subplots get the correct rect
-        # hacky but it works for now
-        example.figure.imgui_renderer.render()
-
-    example.figure._fpl_reset_layout()
-    # render each subplot
-    for subplot in example.figure:
-        subplot.viewport.render(subplot.scene, subplot.camera)
-
-    # flush pygfx renderer
-    example.figure.renderer.flush()
-
-    if fpl.IMGUI:
-        # render imgui
-        example.figure.imgui_renderer.render()
-
-    # render a frame
-    img = np.asarray(example.figure.renderer.target.draw())
+    # Render a frame. Twice, because layouts don't settle in one go?
+    m1 = example.figure.renderer.target.draw()
+    m2 = example.figure.renderer.target.draw()
+    img = np.asarray(m2)
 
     # check if _something_ was rendered
     assert img is not None and img.size > 0
@@ -164,7 +145,6 @@ def test_example_screenshots(module, prep_environment):
     ref_img = normalize_image(ref_img)
 
     similar, rmse = image_similarity(rgb, ref_img)
-
     update_diffs(module.stem, similar, rgb, ref_img)
     assert similar, (
         f"diff {rmse} above threshold for {module.stem}, see "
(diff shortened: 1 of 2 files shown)


kushalkolar:
@almarklein getting large diffs with colors across the entire canvas, did something change recently?

Regenerate is producing nans for some images so I'm guessing the image saving lib change something recently: https://github.com/fastplotlib/fastplotlib/actions/runs/19291516914/job/55163104901?pr=938#step:8:1528 

diff-image_simple-rgb

diff-image_rgb-rgb

diff-image_volume_non_orthogonal_slicing-rgb

diff-line_cmap-rgb

diff-line_collection_cmap_values_qualitative-rgb

diff-scatter_validate-rgb

almarklein:
Could you try pinning `pygfx==0.15.2`  in pyproject.toml?

almarklein:
But still, it also happens for `main`. And only for Python 3.13??

kushalkolar:
@clewis7 can merge #939 and then merge `main` into this and hopefully tests pass.

I want to make some examples for this and we can brainstorm how to make this more easily accessible in the API but let's do that later, I made #941 for that.

kushalkolar:
@clewis7 gtg!

PR #939, opened 2025-11-12
update pygfx pin

kushalkolar:
@almarklein curious what exactly the new lower bounds on wgpu and rendercanvas do

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ keywords = [
 requires-python = ">= 3.10"
 dependencies = [
     "numpy>=1.23.0",
-    "pygfx==0.15.2",
+    "pygfx==0.15.3",
     "wgpu",          # Let pygfx constrain the wgpu version
     "cmap>=0.1.3",
     # (this comment keeps this list multiline in VSCode)


almarklein:
Basically make sure that imgui keeps working.

Longer answer: we made changes in the latest rendercanvas and wgpu. The changes to rendercanvas were meant to be compatible with earlier versions of wgpu, but they were not for the imgui adapter. So wgpu-py was updated to fix that, but pygfx 0.15.0 has an upper bound on wgpu, so 😕

kushalkolar:
Doesn't seem to be related to imgui, the non imgui regen screenshots is also failing with nans 🤔. Any chance you could take a look?

almarklein:
Odd, it does not seem to be a problem for PyGfx ...

I have zeroes in on the problem - it seems related to some of the size-refactoring applied recently in rendercanvas. It looks like `context.get_current_texture()`  is called before the canvas updated the size, so the returned texture is just one pixel.

I need to go now, but will work on a solution this evening / tomorrow.

almarklein:
Ok, so rendercanvas kindof assumes that there will be no rendering outside of a draw event. But the logic to take screenshots *does* do that. The recent refactoring in rendercanvas apparently made is less resilient to that.

Let's fix that in both places; make FPL not render outside of a draw event, and make rendercanvas resilient for when it does happen.

Remaining errors:
```
=========================== short test summary info ============================
FAILED examples/notebooks/test_gc.ipynb::test_gc.ipynb - ValueError: extent x-range must be non-negative: [0. 0. 0. 0.]
FAILED examples/notebooks/quickstart.ipynb::quickstart.ipynb - ValueError: extent x-range must be non-negative: [0. 0. 0. 0.]
```

Plus the doc-gallery errors; these actually also happen at the Pygfx repo.

almarklein:
Looks like the upstream fixes did the trick!

kushalkolar:
@clewis7 gtg!

Thanks @almarklein !! :partying_face:

PR #947, opened 2025-11-19
cursor tool and better tooltips

kushalkolar:
This ended up being more than just the cursor tool since I needed to do a few other things to make this work well :laughing: 

# Transforms

This PR adds useful functions to `Graphic`. `map_model_to_world()` and the inverse `map_world_to_model()`: https://github.com/fastplotlib/fastplotlib/pull/947/changes#diff-a57994ecff1cde4e05c8175265219f0b5200da144ce8733b9eb560c09bcea7ddR501-R558

This lets you map model/data space <-> world space. It accounts for any transforms (translation, i.e. `Graphic.offset`, scale and rotation) and maps the actual data position to world space. 

`Graphic.scale` is a new graphic feature.

I added examples that show how all these transforms relate to each other, and how to map between them. They are in [`examples/space_transforms`](https://www.fastplotlib.org/ver/cursor-simpler/_gallery/index.html#spaces-and-transforms)

I now forgot why I added map model <-> world, but I'm sure they will be useful :sweat_smile: .

I also added `PlotArea.map_world_to_screen()`, which is the inverse of `map_screen_to_world()`. The `Cursor` uses this in the following way to display tooltips in other subplots:

1. Pointer is at a certain location in Subplot 1. We know the world space position of the pointer at this location, draw the `Cursor` at this location in Subplot 1. Since the pointer is over this location, we can also draw a tooltip here. Tooltips are drawn in the overlay render pass and are therefore always in _screen space_. 
2. Now we need to draw the cursor at the same world space location in all other subplots, this is simple. But we need the screen space positions for the corresponding world space positions in these other subplots. So we use `map_world_to_screen()` to know where to draw the tooltip in these other subplots. This is also used to get the pick info at this screen position.

# Tooltip

Created `TextBox`, which is a simple text box object. `Tooltip` inherits from `TextBox` and just adds a few uesful properties. 

The tooltips are managed by `PlotArea`, see code review comments for an explanation. 

# Map `WorldObject` -> `Graphic`

I created a dict that maps world object id -> Graphic. 

https://github.com/fastplotlib/fastplotlib/pull/947/changes#diff-a57994ecff1cde4e05c8175265219f0b5200da144ce8733b9eb560c09bcea7ddR40-R42

Useful when manually picking the picking Texture to get picking info about a certain screen position (ex: Cursor tool to get pick info other subplots). Also used to get the pick info when updating the tooltip on every render.

# Cursor

A simple cursor tool, has two modes "crosshair" which uses 2 infinite lines, and "marker" that uses a pygfx point with marker material.

This just maps the same position in world space across all subplots. I think we should keep it simple this way, unlike the ideas proposed in #662 .

Size, color, etc. are settable properties.

The cursor also sets the tooltip in other subplots that have been added to the cursor! Since we know the world space position in all other subplots, we can `map_world_to_screen()` and get the pick info for that position in the other subplot, and then display the tooltip in the other subplots with that pick info https://github.com/fastplotlib/fastplotlib/pull/947/changes#diff-dfc9ee2c327058ca7075a70aed1eb973aeb47e1e08e65a67c512a318ed6036aeR254-R257

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -289,8 +289,7 @@ class Graphic:
                 if isinstance(
                     child, (pygfx.Image, pygfx.Volume, pygfx.Points, pygfx.Line)
                 ):
-                    # need to call int() on it since it's a numpy array with 1 element
-                    # and numpy arrays aren't hashable
+                    # unique 32 bit integer id for each world object
                     global_id = child.id
                     WORLD_OBJECT_TO_GRAPHIC[global_id] = self
                     # store id to pop from dict when graphic is deleted


examples/misc/cursors.py
clewis7, line 1: Should add a similar example for `"marker"` mode

fastplotlib/tools/_cursor.py
```diff
@@ -0,0 +1,357 @@
...
+        self.mode = mode
+        self.size = size
+        self.color = color
+        self.marker = marker
+        self.edge_color = edge_color
+        self.edge_width = edge_width
+        self.alpha = alpha
+        self.size_space = size_space
+
+        self._pause = False
```
clewis7: code comment

fastplotlib/tools/_cursor.py
clewis7, line 1: doc strings for properties and funcs

fastplotlib/tools/_cursor.py
```diff
@@ -0,0 +1,357 @@
...
+                line_v.geometry.positions.data[:, 0] = pos[0]
+
+                line_v.geometry.positions.update_full()
+
+        self._position[:] = pos
+
+    def add_subplot(self, subplot: Subplot):
+        """add this cursor to a subplot"""
+        if subplot in self._cursors.keys():
+            raise KeyError
```
clewis7: more descriptive error here

fastplotlib/tools/_cursor.py
```diff
@@ -0,0 +1,357 @@
...
+            c.material.color = new_color
+
+        self._color = new_color
+
+    @property
+    def marker(self) -> str:
+        return self._marker
+
+    @marker.setter
+    def marker(self, new_marker: str):
```
clewis7, line 160: should validate the `new_marker` with valid markers (similar to what scatter does)

fastplotlib/graphics/_base.py
kushalkolar, line 1: @almarklein 

We want to do manual picking by calling `renderer.get_pick_info` and then get the graphic that owns the world object which was the picking target. Is this a reliable, recommended way to do it:

I've created a dict that stores the world object global_id, from `world_obj.uniform_buffer.data["global_id"]` and use that to map from world object -> graphic which owns it. 

Does the world object global id ever change? 

The usecase is shown in this comment: https://github.com/fastplotlib/fastplotlib/pull/947#issuecomment-3556905807

kushalkolar:
This is gonna need a bit more work.

Display tooltip at corresponding location in other subplots by manually picking at that location to get the info about the graphic at that location.

Basically:

If system cursor is on subplot A, and we want to display tooltip at same world space location in subplot B:

```python
# system cursor over position in subplot A, get world pos
pos = subplot_A.map_screen_to_world(ev)

# get corresponding screen space position in subplot B
x, y = subplot_B.map_world_to_screen(pos)
# get pick info of this screen pixel
pick = subplot.get_pick_info((x, y))

if pick is None:  # no graphic at this screen pixel in subplot B
    tooltips2.visible = False
    return

# display graphic metadata or anyrelevant info in the tooltip in subplot B
info = pick["graphic"].metadata
tooltips2.display((x, y), str(info))
```

https://github.com/user-attachments/assets/721bf321-793c-4a29-ba93-869ef29561fc

kushalkolar:
@clewis7 I'm thinking about changing who "owns" tooltips. Right now `Figure` owns tooltips and can auto-display them. I'm thinking of a tooltip being owned by each `Subplot` instead, so we can have a method:

```python
Subplot.display_tooltip(pos: tuple[float, float], space: Literal["world", "screen"], info: str):
    ...
```

Tooltips are drawn in the overlay render pass which is in screen space, so the tooltip position in the end always has to be in screen space. However it is useful to think of them in world space since that's where the graphics are. It'll just be much easier to know the position of a graphic in world space where you want to show the tooltip, and the `Subplot.display_tooltip()` will project that pos to screen space under automatically.

fastplotlib/graphics/_base.py
almarklein: The id of a world object can be accessed via `world_obj.id`. And it is guaranteed to stay the same.

clewis7:
> @clewis7 I'm thinking about changing who "owns" tooltips. Right now `Figure` owns tooltips and can auto-display them. I'm thinking of a tooltip being owned by each `Subplot` instead, so we can have a method:
> 
> ```python
> Subplot.display_tooltip(pos: tuple[float, float], space: Literal["world", "screen"], info: str):
>     ...
> ```
> 
> Tooltips are drawn in the overlay render pass which is in screen space, so the tooltip position in the end always has to be in screen space. However it is useful to think of them in world space since that's where the graphics are. It'll just be much easier to know the position of a graphic in world space where you want to show the tooltip, and the `Subplot.display_tooltip()` will project that pos to screen space under automatically.

And why would tooltips not be owned by the graphic they correspond to? I could also imagine a property that is `image_graphic.tooltip.display = False`

kushalkolar:
> And why would tooltips not be owned by the graphic they correspond to? I could also imagine a property that is `image_graphic.tooltip.display = False`

I didn't think of that! This might be a better way to do it, and customization of tooltips per-graphic makes most sense I think.

clewis7:
>I didn't think of that! This might be a better way to do it, and customization of tooltips per-graphic makes most sense I think.

I think then, if you wanted to not display all of the tooltips for graphics in the subplot, you would just need to iterate:
```python
for g in subplot.graphics:
     g.tooltip.display = False 
```

kushalkolar:
ok all this stuff works now:

tooltips update on every render, useful when data under the cursor changes:

https://github.com/user-attachments/assets/b08cc030-3c47-4d62-beb3-941dcbdfba4e

Tooltips are now enabled on graphics by default, and a cursor can trigger a tooltip to show up in other subplots:

https://github.com/user-attachments/assets/b8c07081-c421-428b-b987-5f403c79b19b

Added more mapping functions, these are explained in the guide:

    - ``Subplot.map_screen_to_world((x, y))``
    - ``Subplot.map_world_to_screen((x, y, z))``
    - ``Graphic.map_model_to_world((x, y, z))``
    - ``Graphic.map_world_to_model((x, y, z))``

Need to test a bit, add some more examples maybe, and possibly cleanup some of the code if it can be simplified. 

Need to add tests to make sure the right default tooltip display comes up for all graphics. 

I should add examples that explain all the spaces (model, world, screen) using a variety of graphics. Show a few points on the graphic and put text that indicates model, world and screen space positions. 
  - Graphic that isn't rotated or scaled, model and world are same
  - Graphic scaled
  - Graphic rotated
  - Graphic scaled, rotated, translated

Do this with Image, Line and an ellipsoid, should be enough.

@clewis7 you can take a look, but I'm guessing you're deep in CPU design optimizations for the next few days :laughing: . I'll summarize everything in this PR later, required quite a bit more than I thought to get this to work.

kushalkolar:
OK things getting slightly messy, I think this will be better:

* `TextBox` that can be used to display a textbox anywhere, just text with a rectangle mesh behind it. Lives in the overlay scene.
* `Tooltip` is a subclass of `TextBox`. One tooltip instance always exists per subplot. `Graphic.tooltip` merely sets the tooltip in its subplot.

The `PlotArea` could also update its tooltip with whatever pick info is under a pixel on every render.

Perhaps the renderer `"pointer_move"` event could just be used to set the tooltips in each subplot by getting the graphic from the pick_info at that position.

kushalkolar:
I think things are much simpler now. 

- Created `TextBox` class
- `Tooltip` which is a light subclass of `TextBox`
- `PlotArea` has a `Tooltip` instance as an attribute. `"pointer_move"` events will set the tooltip. 
- if `Tooltip.continuous_mode`, then `PlotArea` will update the tooltip on every render
- Cursor uses `Subplot.get_pick_info()` to get the graphic, pick info, and set the tooltip in that subplot.
 
~One thing left out is I need to think of how to set the tooltip w.r.t. the transform when it's on the current subplot with a cursor.~ Figured it out, let the cursor manage tooltips entirely when a subplot is added to it.

https://github.com/user-attachments/assets/063e8848-b125-4096-b007-76661ca7fb0a

kushalkolar:
~One last thing to figure out, programmatically setting the cursor isn't triggering the tooltip to appear~

nevermind it does work, but not on the first render since the canvas is blank

kushalkolar:
For the examples to demonstrate the different spaces I think I'll make one example for each of these (instead of separate subplots in one Figure) and two sets, one with an image and one with a line:

- no transforms
- translation
- scaling
- rotation
- translation + scaling
- translation + scaling + rotation

fastplotlib/tools/_cursor.py
kushalkolar: renamed this to `enabled` so it's uniform with the rest of the API

fastplotlib/tools/_cursor.py
kushalkolar: I think pygfx does this validation when setting

fastplotlib/layouts/_plot_area.py
```diff
@@ -123,6 +126,10 @@ def __init__(
         self.scene.add(self._ambient_light)
         self.scene.add(self._camera.add(self._directional_light))
 
+        self._tooltip = Tooltip()
+        self.get_figure()._fpl_overlay_scene.add(self._tooltip._fpl_world_object)
+        self.renderer.add_event_handler(self._fpl_set_tooltip, "pointer_move")
```
kushalkolar, line 131: `PlotArea` manages a `Tooltip` instance. Tooltips are draw in the overlay render pass, and are therefore in screen space. The `"pointer_move"` event determines where it appears (in screen space). Therefore there is only ever 1 tooltip instance (and thus only one mesh, text, and line object required to draw it) per `PlotArea`. If they're managed by a graphic then things get complicated.

fastplotlib/layouts/_plot_area.py
```diff
@@ -428,27 +569,37 @@ def clear_animations(self, removal: str = None):
...
                 f"Animation type: {removal} must be one of 'pre' or 'post'. To remove all animation "
                 f"functions, pass `type=None`"
             )
 
+        for func in to_remove:
+            if func in self._animate_funcs_persist:
+                # skip
+                continue
+            self.remove_animation(func)
+
```
kushalkolar: I need to get rid of this `persist` thing. It was an idea for updating the tooltip on every render cycle when I implemented tooltips being managed by a `Graphic` and is not required anymore.

fastplotlib/graphics/_base.py
```diff
@@ -251,6 +283,24 @@ def world_object(self) -> pygfx.WorldObject:
...
         WORLD_OBJECTS[self._fpl_address] = wo
 
+        # add to world object -> graphic mapping
+        if isinstance(wo, pygfx.Group):
+            for child in wo.children:
+                if isinstance(
+                    child, (pygfx.Image, pygfx.Volume, pygfx.Points, pygfx.Line)
+                ):
+                    # need to call int() on it since it's a numpy array with 1 element
+                    # and numpy arrays aren't hashable
```
kushalkolar:
```suggestion
                    # unique 32 bit integer id for each world object
```

fastplotlib/layouts/_plot_area.py
```diff
@@ -333,6 +350,117 @@ def map_screen_to_world(
...
+            pick info if a graphic is at this position, else None
+
+        """
+
+        info = self.renderer.get_pick_info(pos)
+
+        if info["world_object"] is not None:
+            # if this world object is owned by a graphic
+            if info["world_object"].id in WORLD_OBJECT_TO_GRAPHIC.keys():
+                info["graphic"] = WORLD_OBJECT_TO_GRAPHIC[info["world_object"].id]
```
kushalkolar, line 407: graphics get added to the `WORLD_OBJECT_TO_GRAPHIC` dict in `Graphic._set_world_object()`: https://github.com/fastplotlib/fastplotlib/pull/947/changes#diff-a57994ecff1cde4e05c8175265219f0b5200da144ce8733b9eb560c09bcea7ddR286-R303

kushalkolar:
@clewis7 finally r4r! And just before midnight here :joy:

kushalkolar:
There is a bug in Volume, will figure out.

Edit: this is a pygfx thing, posted: https://github.com/pygfx/pygfx/issues/1251

kushalkolar:
Until https://github.com/pygfx/pygfx/pull/1253 is in the next release of pygfx I've kinda disabled tooltips for `ImageVolume`.

clewis7 (approved):

PR #948, opened 2025-11-19
Run CI with rendercanvas lazy size proposal

almarklein:
Trying out https://github.com/pygfx/rendercanvas/pull/148


almarklein:
For some reason the CI runs are not shown in this pr. But in the *Action* tab they can be inspected, and they all pass.

PR #950, opened 2025-11-23
add imgui file dialog example

FlynnOConnell:
WIP, Relies on #868, might be simplified by #849.

image

This example could save another synthetic dataset to png or mp4 to showcase loading different sizes / # of dims.

Could also instead copy the [imgui_basic example](https://github.com/fastplotlib/fastplotlib/blob/main/examples/guis/imgui_basic.py) for the UI and put the open file / open folder in a menu:

image

### DIFF
--- a/examples/guis/imgui_file_dialog.py
+++ b/examples/guis/imgui_file_dialog.py
@@ -74,7 +74,25 @@ class FileDialogWidget(EdgeWindow):
         """Load an image file and update the graphic."""
         try:
             img_data = iio.imread(filepath)
-            self._image_graphic.data = img_data
+
+            # validate image format: must be 2D (grayscale) or 3D with 1-4 channels
+            if not (img_data.ndim == 2 or (img_data.ndim == 3 and img_data.shape[2] <= 4)):
+                self._status_msg = (
+                    f"Unsupported format: {img_data.shape}. "
+                    f"Expected (H, W) or (H, W, C) with C <= 4"
+                )
+                return
+
+            # check if the new image shape matches the current buffer
+            if img_data.shape == self._image_graphic.data.value.shape:
+                self._image_graphic.data = img_data
+            else:
+                # shape mismatch: remove old graphic and create a new one
+                subplot = self._image_graphic._plot_area
+                subplot.remove_graphic(self._image_graphic)
+                self._image_graphic = subplot.add_image(img_data)
+                subplot.auto_scale()
+
             self._status_msg = f"Loaded: {filepath}"
         except Exception as e:
             self._status_msg = f"Error: {e}"


Vipitis:
That is a really useful example. I looked at it briefly and might expand it with sequence thumbnails and an asnyc/threaded data loader as I have a lot of different sequences in a dataset to evaluate but they are on a slowish network drive.

kushalkolar:
If you can make a simple example that just uses imageio to load existing image files and just displays an `ImageGraphic` I think that would be simpler and I can merge it. I'm holding off on any PRs for ImageWidget until #951 

> That is a really useful example. I looked at it briefly and might expand it with sequence thumbnails and an asnyc/threaded data loader as I have a lot of different sequences in a dataset to evaluate but they are on a slowish network drive.

Sequences, as in image sequences? Sounds like making a numpy-like interface for lazy loading is the way to go, we do it _a lot_ in another project: https://github.com/apasarkar/masknmf-toolbox/blob/main/masknmf/arrays/array_interfaces.py

examples/guis/imgui_file_dialog.py
```diff
@@ -0,0 +1,284 @@
...
+
+        self._current_data_shape = new_data.shape
+        self._status_msg = f"Loaded!\nShape: {new_data.shape}"
+        self._status_color = imgui.ImVec4(0.3, 1.0, 0.3, 1.0)  # green for success
+        print(f"Loaded: {self._current_path}, shape: {new_data.shape}")
+
+    except Exception as e:
+        self._status_msg = f"Error: {str(e)}"
+        self._status_color = imgui.ImVec4(1.0, 0.3, 0.3, 1.0)  # red for errors
+        print(f"Error loading data: {e}")
```
kushalkolar: I think you're missing an indent

examples/guis/imgui_file_dialog.py
```diff
@@ -0,0 +1,92 @@
...
+        imgui.spacing()
+        imgui.separator()
+        imgui.spacing()
+        imgui.text_wrapped(self._status_msg)
+
+    def _load_image(self, filepath):
+        """Load an image file and update the graphic."""
+        try:
+            img_data = iio.imread(filepath)
+            self._image_graphic.data = img_data
```
kushalkolar: Need to create a new image graphic if the buffer shape doesn't match.

Vipitis:
> Sequences, as in image sequences? Sounds like making a numpy-like interface for lazy loading is the way to go, we do it _a lot_ in another project: https://github.com/apasarkar/masknmf-toolbox/blob/main/masknmf/arrays/array_interfaces.py

I got an easy solution by making the numpy array of the correct size as zeros (maybe even `np.empty` could work) and then I just write the pixel data into that using `threading.Thread` while the GUI already shows. Here the loading order is predetermined, so random jumps won't "lazy load". Will still look into the thumbnails when I have more time.

examples/guis/imgui_file_dialog.py
```diff
@@ -74,7 +74,25 @@ def _load_image(self, filepath):
...
+                    f"Expected (H, W) or (H, W, C) with C <= 4"
+                )
+                return
+
+            # check if the new image shape matches the current buffer
+            if img_data.shape == self._image_graphic.data.value.shape:
+                self._image_graphic.data = img_data
+            else:
+                # shape mismatch: remove old graphic and create a new one
+                subplot = self._image_graphic._plot_area
```
kushalkolar, line 91: `EdgeWindow` has a `figure` instance so you can use that to get the subplot. `Graphic._plot_area` isn't part of the public API

FlynnOConnell:
@Vipitis Instead of fixing the the issues above I used claude to extract a data loader from one of my other repos into a python package [imgui_data_loader](https://github.com/FlynnOConnell/imgui_data_loader/tree/master) you might be interested in! @kushalkolar I will do the above at some point or could just import that data loader

PR #953, opened 2025-11-28
Add support for mesh and surface

### LINKED ISSUE #487: 2D polygon graphics
2D Polygons meshes

### LINKED ISSUE #917: Mesh based graphics
Just laying out ideas:

- surfaces, provide 3 dimensions of data that define x, y, z positions
  - apply colormap on surface using 2 dimensions
  - apply a `TextureArray` on a surface (might be related to the previous point)
- arbitrary polyhedron shapes
- importing meshes and being able to programmatically change x, y, z, positions of the mesh? 

Ideas from Pan: https://github.com/pygfx/pygfx/issues/836#issuecomment-2272605613
Code from Pan below the above comment: https://github.com/pygfx/pygfx/issues/836#issuecomment-2272664027

almarklein:
Closes #487, closes #917, ref #959

* [x] Add `MeshGraphgic`
* [x] Add  `plot.add_mesh()`  and `plot.add_surface()`.
* [x] Add light to a scene by default.
* [x] Add examples.
* [x] Support for 2D colormaps.  
* [x] This relies more on PyGfx colormap mechanics. I am not sure why fastplotlib 'renders' its own colormap into vertex colors for lines and scatter.

Possible upstream (pygfx) improvements:
* Support for clim everywhere where a map is supported.
* Possibility to set map as a numpy array?


almarklein:
I cannot find the new screenshots. Do I need to whitelist the new `examples/mesh` dir somewhere?

kushalkolar:
Good start!

`_graphic_methods_mixin.py` is meant to be auto-generated, so we should make a `SurfaceGraphic` that probably just subclasses `MeshGraphic`.

Working on figuring out how to make the surface data dynamically settable so we can have a `data` `@property`. I think all the properties should be mutable like the other graphics where it make sense.

For general mesh:
- positions
- indices
- mapcoords
- cmap
- colors

For surface:
- data (which under the hood computes positions, indices, mapcoords)
- cmap
- colors

~Seems like the normals have to be re-computed if using phong material when positions change. No difference with basic.~ Edit: Can just set the `material.map` again. Is there a flag for this we can set?

~Anyways, I think we can make the basic material mode as the default and also allow phong and slice.~ Nah phong default seems best for most functions.

Are [morph targets](https://docs.pygfx.org/stable/_gallery/feature_demo/morph_targets.html#sphx-glr-gallery-feature-demo-morph-targets-py) useful for this? Or just setting the position directly is the way to go?

I'm not sure if any of the existing "2D" selectors are useful here, in the future we could make a "cube selector", maybe based on the [pygfx gizmo example](https://docs.pygfx.org/stable/_gallery/feature_demo/helpers_gizmo.html#sphx-glr-gallery-feature-demo-helpers-gizmo-py) using an expandable cube that can be used to select the triangles inside the cube?

Other than that that #545 might be useful here when it's available, and we could also make a selector that just allows selecting faces on a mesh (I think pygfx has an example for this, but we could maybe make a easy way to Cntrl + Click to build a selection?). Anyways not a priority for now, just thoughts.

kushalkolar:
I'm going to add a few commits to this branch based on https://github.com/fastplotlib/fastplotlib/pull/953#issuecomment-3590135523

fastplotlib/graphics/mesh.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return pygfx_cmap
+
+
+class MeshGraphic(Graphic):
+    _features = {
+        "positions": VertexPositions,
+        "indices": BufferManager,
+        "mapcoords": (BufferManager, None),
+        "colors": (VertexColors, UniformColor),
```
kushalkolar, line 27: I'm making some mesh-specific versions of these, not much work, mostly boilerplate

fastplotlib/graphics/mesh.py
```diff
@@ -0,0 +1,172 @@
...
+            if colors is None:
+                uniform_color = "w"
+                self._colors = UniformColor(uniform_color)
+            elif isinstance(colors, str) or isinstance(colors, tuple):
+                uniform_color = colors
+                self._colors = UniformColor(uniform_color)
+            elif isinstance(colors, VertexColors):
+                per_vertex_colors = True
+                self._colors = colors
+                self._colors._shared += 1
```
kushalkolar: we're not using this anymore (I thought I removed it completely), we figured if people are sharing buffers they need to keep track of things by themselves, overengineered to keep track of it here 😅

fastplotlib/graphics/mesh.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+
+        if mapcoords is None:
+            self._mapcoords = None
+        elif isinstance(mapcoords, BufferManager):
+            self._mapcoords = mapcoords
+        else:
+            self._mapcoords = mapcoords = BufferManager(
+                mapcoords, isolated_buffer=isolated_buffer, property_name="mapcoords"
+            )
```
kushalkolar: I think changing texcoords would be done in-conjunction with changes to positions, indices, or the cmap/image so we can just make this a simple property instead of a `BufferManager`

fastplotlib/graphics/mesh.py
```diff
@@ -0,0 +1,172 @@
...
+            pick_write=True,
+        )
+
+        # Set all the data
+        if per_vertex_colors:
+            geometry.colors = self._colors.buffer
+        if mapcoords is not None:
+            geometry.texcoords = self._mapcoords.buffer
+        if pygfx_cmap is not None:
+            material.map = pygfx_cmap
```
kushalkolar: I'm making a mesh specific `MeshCmap` feature, and I think it'll be easier for users to have a different property specific for images.

fastplotlib/layouts/_graphic_methods_mixin.py
```diff
@@ -432,6 +433,208 @@ def add_line_stack(
...
+            positions,
+            indices,
+            colors,
+            mapcoords,
+            cmap,
+            isolated_buffer,
+            **kwargs,
+        )
+
+    def add_surface(
```
kushalkolar: going to make a subclass for surfaces based on this!

fastplotlib/layouts/_plot_area.py
```diff
@@ -117,6 +117,9 @@ def __init__(
         self._background = pygfx.Background(None, self._background_material)
         self.scene.add(self._background)
 
+        self.scene.add(pygfx.AmbientLight())
+        self.scene.add(self._camera.add(pygfx.DirectionalLight()))
```
kushalkolar: I think these can be settable properties of the `PlotArea`

fastplotlib/graphics/mesh.py
kushalkolar: nvm, too complicated XD

fastplotlib/layouts/_graphic_methods_mixin.py
```diff
@@ -432,6 +433,208 @@ def add_line_stack(
...
+
+        """
+
+        def check_z(z):
+            if z.ndim != 2:
+                raise ValueError("Z must be a 2D array.")
+
+        # In VisVis (https://github.com/almarklein/visvis/blob/main/visvis/functions/surf.py)
+        # the tuple can be 2-element or 4-element, to pass a color per z-value.
+        # I disabled that by commenting related logic. Maybe it can be enabled someday.
```
kushalkolar: Been playing around with this, I think it would be clearer if there's an explicit kwarg for this, can look into it later when a usecase arises.

kushalkolar:
> I cannot find the new screenshots. Do I need to whitelist the new `examples/mesh` dir somewhere?

example dirs for tests are listed here: https://github.com/fastplotlib/fastplotlib/blob/7b4c68746ddb7028fb32ef3adb239506be5210dc/examples/tests/testutils.py#L18-L34

Did you automate finding them in pygfx?

Similarly for docs that dir needs a readme, making these changes in #954

kushalkolar:
> This relies more on PyGfx colormap mechanics. I am not sure why fastplotlib 'renders' its own colormap into vertex colors for lines and scatter.

`pygfx.cm.create_colormap()` didn't exist when we implemented lines and scatter, is this what you mean? 

Edit: I see `LineMaterial.map` which can be a `TextureMap`. Is it possible to apply what we call a "cmap transform" as well, for instance in this example it samples from the colormap using the y-values of a sine function: https://www.fastplotlib.org/ver/dev/_gallery/line/line_cmap.html#sphx-glr-gallery-line-line-cmap-py 

But this can be arbitrary, we often use it to indicate things like velocity along a 2d trajectory.

> I cannot find the new screenshots. Do I need to whitelist the new `examples/mesh` dir somewhere?

a comma was also missing, [added it](https://github.com/fastplotlib/fastplotlib/pull/953/commits/2ddb97f9a43fd413f20c260ded7debf6c3abe60a) , slipped by in the vectors PR

fastplotlib/graphics/mesh.py
```diff
@@ -0,0 +1,473 @@
...
+    triangulate_polygon,
+)
+
+
+class MeshGraphic(Graphic):
+    _features = {
+        "positions": MeshVertexPositions,
+        "indices": MeshIndices,
+        "colors": (VertexColors, UniformColor),
+        "cmap": MeshCmap,
```
kushalkolar, line 28: @almarklein your original question:

> Was the `mapcoords` feature removed intentionally? I can see it being useful. Maybe I don't properly understand features vs normal properties.

"Graphic Features" are evented properties. You can add event handlers to them to capture when they change.

The `BufferManager` is also a graphic feature, but allows numpy-like slicing and setting and automatically calls `update_range()` on the `pygfx.Buffer` object. 

Since `mapcoords` are usually changed in conjunction with a change in positions or cmap, are there use cases to change the `mapcoords` by itself without changing other things? If so then it should be a feature too.

fastplotlib/graphics/features/_mesh.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+
+        data = self._fix_data(data)
+        super().__init__(
+            data, isolated_buffer=isolated_buffer, property_name=property_name
+        )
+
+    def _fix_data(self, data):
+        if data.shape == (3,):
+            pass
```
kushalkolar: @almarklein your question:

> Is this a single triangle?

I was also confused by this, I interpreted it from here in the polygon selector:

https://github.com/fastplotlib/fastplotlib/blob/2a430f8e1805ba4dc51de06201f1a73cde5d385b/fastplotlib/graphics/features/_selection_features.py#L412-L415

fastplotlib/graphics/mesh.py
almarklein: Yeah I think so. For context, in classic rendering, the texcoords are usually fixed, so that the map (usually a 2d image) is correctly displayed. But for scientific applications, I see the value of mapcoords as a domain-specific value of interest, which is visualized/colorized with a colormap. Imagine force/stress in a geometric model:

image

Theses values can also be 2D. I.e. 2 values per vertex, and then look them up in a 2D colormap. Pygfx supports up to 3D.

BTW: historically these are called texture coordinates, or `texcoords`, because the colormap is a texture. But from a higher level perspective they are just 'values'. I kinda like `mapcoords` as a middle ground. I'm tempted to adopt that name in Pygfx too...

fastplotlib/graphics/features/_mesh.py
almarklein: This will still make that indices is always (n,3), it's just that n can be zero.

almarklein:
> pygfx.cm.create_colormap() didn't exist when we implemented lines and scatter, is this what you mean?
>
> Edit: I see LineMaterial.map which can be a TextureMap. Is it possible to apply what we call a "cmap transform" as well, for instance in this example it samples from the colormap using the y-values of a sine function: https://www.fastplotlib.org/ver/dev/_gallery/line/line_cmap.html#sphx-glr-gallery-line-line-cmap-py

Ah, `create_colormap()` not existing at the time explains it. Though it would be nice if fpl could simply leverage Pygfx's colormap support, by setting `geometry.texcoords` and `material.map`.

If I understand correctly, the current line's `geometrytexcoords` are simply the vertex indices, and `map_transform` is used to map that to real values. So in fact, the current `map_transform` **is** the `mapcoords`. Which is good news, I guess, because it means moving to pygfx doing the colormapping should be relatively easy (in a different pr).

*edit: easy as in not that complex. Still quite a bit of work, I imagine.*

fastplotlib/layouts/_plot_area.py
almarklein: mmm ... Having an ambient light and one directional light on the camera is a sweet default that makes sure phong-shaded meshes look good. A user that wants custom lights will probably want to add more lights to the scene, and control on which objects they are attached.  That would be tricky to expose as a prop, I think?

FWIW these lights add negligible overhead for objects that don't use lights.

fastplotlib/graphics/features/_mesh.py
```diff
@@ -0,0 +1,297 @@
...
+            pygfx_cmap = pygfx.cm.create_colormap(cmap)
+        else:
+            tex = pygfx.Texture(map, dim=map.ndim - 1)
+            pygfx_cmap = pygfx.TextureMap(tex)
+
+    return pygfx_cmap
+
+
+class MeshVertexPositions(VertexPositions):
+    """Manages mesh vertex positions, same as VertexPosition but data must be of shape [n, 3]"""
```
almarklein: This extra class feels like unnecessary bloat to me. I propose to remove it. The `VertexPositions` class already converts 2D data to 3D, so users can create in-plane meshes if they want.

fastplotlib/graphics/features/_mesh.py
almarklein: fixed.

fastplotlib/graphics/mesh.py
almarklein: But maybe we can wait with making this a feature, until we decide on what to do with the colormaps of lines etc.

kushalkolar: Ah I understand it now and how it's the same as cmap_transform but multi dimensional. One easy usecase is heat diffusion in 3d objects. I added it to #959 so we can do it later. Thanks for the context and ideas, I rarely do things with meshes! :smile:

kushalkolar:
> If I understand correctly, the current line's `geometrytexcoords` are simply the vertex indices, and `map_transform` is used to map that to real values. So in fact, the current `map_transform` **is** the `mapcoords`. Which is good news, I guess, because it means moving to pygfx doing the colormapping should be relatively easy (in a different pr).

Yes! Thanks I understand it better now :)

fastplotlib/graphics/features/_mesh.py
kushalkolar: True I realize it now looking at them again :laughing:

fastplotlib/layouts/_plot_area.py
kushalkolar: I think it's useful to have easy access to them though, and if a new camera is set on the `PlotArea` it needs to get the directional light. This is what I mean:

https://github.com/fastplotlib/fastplotlib/pull/953/files#diff-2a0f782c81512ab17081f35fd6cac201fafd429eded696cb76981946bc6baaecR120-R294

kushalkolar:
@almarklein I think this is done :partying_face: . Let me know what you think

almarklein:
I was getting ready to update the buffer resizing logic for the poly data, but saw you already fixed that too. Awesome! Ready from my end then! 👍

fastplotlib/layouts/_plot_area.py
almarklein: Ah I see. Nice fix!

PR #970, opened 2025-12-22
add parrot icon, favicon, use for docs and window icon

### LINKED ISSUE #966: display party parrot as window icon
When opening a window, it would be nice if the fastplotlib logo mr party parrot was an Icon for the window 

Heres what it currently looks like: 

Image

----

For example, here's our [image viewer](https://github.com/MillerBrainObservatory/mbo_utilities/tree/master):

Image

FlynnOConnell:
Closes #966


kushalkolar:
I think you can just make a square icon with the face like here: https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib/assets/fastplotlib_face_logo.png 

Also make sure these get into git LFS and not directly committed

fastplotlib/layouts/_figure.py
```diff
@@ -701,6 +703,53 @@ def close(self):
         if self._sidecar:
             self._sidecar.close()
 
+    def _set_window_icon(self):
+        """Set the window icon based on canvas type."""
```
kushalkolar: [`layouts/_utils.py`](https://github.com/fastplotlib/fastplotlib/blob/main/fastplotlib/layouts/_utils.py) is a better place for this

And it can probably be handled in `make_canvas_and_renderer()`

FlynnOConnell:
image

image

image

The taskbar icon code was getting very complicated to just get the fastplotlib icon working consistently on Windows. GLFW was slightly easier but I think it's probably better implemented in rendercanvas

kushalkolar:
a nice start!

PR #971, opened 2025-12-25
`NDWidget`

kushalkolar:
# Internals

## RangeContinuous

Just a dataclass that stores the start, stop, and step of a _reference range_. A reference range is usually in scientific units, such as seconds/ms for time. Depth is another example of a reference range. A reference range just specifies the min and max of a dimension that will be used as a slider dimension. The imgui sliders use the reference range to determine the min, max, and step size of the UI elements (sliders, step-one button, etc.).

`RangeDiscrete` is mostly a placeholder for now, not fully implemented yet, waiting for a proper usecase. Genes might be an example? Session index for multi-session data could be another potential usecase.

Users must define a reference range for every dimension they want to use as a slider dimension.

TODO: Would be useful to have an automatic reference range that's generated for a slider dimension when it's not specified. For example, sometimes you just want to dump a bunch of calcium arrays and look at them, and they're already in the same time-space. Could have a simple auto-reference range mode that just sets the min and max for eacher slider dim based on the current arrays in the `NDWidget`.

## ReferenceIndex

This manages the reference index for an `NDWidget`. It's the only place where indices should be set.

TODO: pushing/popping reference ranges.

## NDProcessor

This manages n-dimensional data of any type. Subclasses must implement a `get()` method which takes a dict mapping dim names -> index for that dim. If specified, window functions are applied on slider dimensions. `_get_slider_dims_indexer` takes the indices dict and outputs a dict of dim_names -> slice. `_apply_window_functions()` uses these slices and applies the window functions. The final output after window functions has the same number of dims as the data, but with 1 element in any dimensions that were sliced (i.e. all slider dims, spatial dims remain untouched). In the end, the `get()` method must return an array that can be directly mapped to graphic data.

Key properties in an `NDProcessor`:

`data` - usually an xarray, but subclasses can manage data that is of any type. The NDProcessor `NDPP_Pandas` for positional data stored in a pandas dataframe is an example of this.

`shape`, `ndim`, `dims`, `spatial_dims`, `slider_dims` - self explanatory, dims are always named

`window_funcs` - window functions, dict {dim_name: (func, window_size)}. A window function must take `axis` and `keep_dims` as kwargs.
`slider_dim_transforms` - dict of functions for each slider dim that maps a reference range value to the local array index value. If a user doesn't provide a transform for a dim (for example, they just want to look at many arrays which are already in the same time-space, such as calcium movies) the it just uses an identify mapping.

`NDPositionsProcessor` and `NDImageProcessor` have logic specific to those types of data.

`NDPositionsProcessor` has an additional `display_window` properties which acts on the `datapoints` dimension. The `datapoints` dim is also both a slider dim and spatial dim, but it is excluded for `window_funcs`. `datapoints_window_func` is a separate property on `NDPositionsProcessor` that applies window funcs for that dimension since it's a bit different from other slider dims.

## NDGraphic

A n-dimensional graphical representation. It has an `NDProcessor` and `fpl.Graphic` that it manages. `ReferenceIndex` sets the indices on an `NDGraphic`, which then call `get()` on its `NDProcessor` to retrieve the new graphic data, and then sets it on the graphic. It has property aliases for the various processor properties, so that users can set window funcs, slider dim transforms, etc.

`NDPositions` and `NDImage` are two `NDGraphic` subclasses that have specific code for positional and image specific representations.

`NDPositions` also manages a linear selector for timeseries representations, and allows swapping between scatter, line stack, line collection, and heatmap representations. It also manages an "auto x range" mode for timeseries representations

`NDImage` also manages the `HistogramLUTTool` and swapping between `ImageGraphic` and `ImageVolumeGraphic`. It also sets the camera and controller when these are swapped.

## NDWSubplot

The main entry point for users to add `NDGraphics` to a subplot in an `NDWidget`. It mainly has `add_nd<...>` methods for users to provide data and kwargs to add `NDGraphics`.

Need to cleanup this subclass so the args and kwargs for each add method are properly specified and documented.

## NDWidget

Just a very simple class that has a `fpl.Figure`, an instance of `ReferenceIndex`, and the `NDWSubsplots`.

Very simplified diagram:

`s1, ... sn` are slider dims, `d1, d2, ...` are spatial dims.

image

# User API

The user mainly interacts via the NDW subplots and `NDGraphics`. For example with images:

```python
data = np.random.rand(1000, 30, 64, 64)

# must define a reference range for each dim
ref = {
    "time": (0, 1000, 1),
    "depth": (0, 30, 1),
}


ndw = fpl.NDWidget(ref_ranges=ref, size=(700, 560))
ndw.show()

ndi = ndw[0, 0].add_nd_image(
    data,
    ("time", "depth", "m", "n"),  # specify all dim names
    ("m", "n"),  # specify spatial dims IN ORDER, rest are auto slider dims
    name="4d-image",
)

# change spatial dims on the fly
# ndi.spatial_dims = ("depth", "m", "n")
```

Positional data:

```python
# generate some toy timeseries data
n_datapoints = 100_000  # number of datapoints per line
n_freqs = 20  # number of frequencies
n_ampls = 15  # number of amplitudes
n_lines = 8

xs = np.linspace(0, 1000 * np.pi, n_datapoints)

data = np.zeros(shape=(n_freqs, n_ampls, n_lines, n_datapoints, 2), dtype=np.float32)

for freq in range(data.shape[0]):
    for ampl in range(data.shape[1]):
        ys = np.sin(xs * (freq + 1)) * (ampl + 1) + np.random.normal(
            0, 0.1, size=n_datapoints
        )
        line = np.column_stack([xs, ys])
        data[freq, ampl] = np.stack([line] * n_lines)


# must define a reference range, this would often be your time dimension and corresponds to your x-dimension
ref = {
    "freq": (1, n_freqs + 1, 1),
    "ampl": (1, n_ampls + 1, 1),
    "angle": (0, xs[-1], 0.1),
}

ndw = fpl.NDWidget(ref_ranges=ref, size=(700, 560))

nd_lines = ndw[0, 0].add_nd_timeseries(
    data,
    ("freq", "ampl", "n_lines", "angle", "d"),
    ("n_lines", "angle", "d"),
    slider_dim_transforms={
        "angle": xs,
        "ampl": lambda x: int(x + 1),
        "freq": lambda x: int(x + 1),
    },
    x_range_mode="view-range",
    name="nd-sine"
)

nd_lines.graphic.cmap = "tab10"

subplot = ndw.figure[0, 0]
subplot.controller.add_camera(subplot.camera, include_state={"x", "width"})

ndw.show(maintain_aspect=False)
```

# Main things left todo

- [x] docs
- [x] docstrings
- [x] ~tests, and test edge cases such as when `NDProcessor.get()` returns empty arrays~
- [x] some of examples
~user input validation, see what Edo did recently with nemos~
~generalize reference range to a reference dim class with other property, such as UI slider throttling. Range would be one such property. Units could be another and we could use pint?~
- [x] auto reference ranges
- [x] push reference range on `ReferenceIndex`
  - [x] pop reference range
- [x] other graphical features such as colors, cmap, sizes, etc. Also maintain things like cmap across lines/scatters when switching graphical representations. 
- [x] if data is set to None, delete the graphic, indices setter should just ignore. Might need to update `ReferenceIndex` to also ignore it. 
- [x] `NDSubplot.delete_ndgraphic()`
- ~`NDSubplot` as a subclass of `Subplot`~ I've ruled this out as a bad idea. Composition is the right pattern here, it becomes a mess otherwise. 
- [x] prevent graphic stacks from mangling when using `LineStack` or `ScatterStacks` due to max y-value in `p` dim
- [x] pandas lazy import for that processor
~optimize pandas processor `get()`~
- [x] property to get the current display range in ref units on NDPositionsGraphic.
~nice repr, this seems like a good tutorial https://dev.to/dcodeyt/creating-beautiful-html-tables-with-css-428l~
- [x] renaming things for consistency, clarity
  - [x] `ReferenceIndex` -> `ReferenceIndices`, it manages multiple indices
  - [x] `ref_indices` -> `indices`, keep constructor and properties symmetric like the rest of fpl
  - [x] `spatial_dims` -> ???, see https://github.com/fastplotlib/fastplotlib/pull/1067#issuecomment-5183589879
- [x] `ImageWidget` with almost same API built on top of `NDWidget`

implements #951


kushalkolar:
Got basic timeseries with linestack working. I've also got some code snippets for interpolating to display heatmap with non-uniformly sampled timeseries data. I should be able to have this fully working with time series very soon :D 

https://github.com/user-attachments/assets/a14bbc29-8b27-4de0-a0cb-13a08c356dad

kushalkolar:
Got heatmap to display timeseries working. It should also work with non-uniformly sampled data by interpolating, need to test.

Also need to implementing switching between heatmap and line representations, need to delete the graphic when switching.

https://github.com/user-attachments/assets/fef4b773-dbcf-4de3-8d64-142895e34730

kushalkolar:
So timeseries can be represented with arrays of one of the following shapes (let's ignore x-axis values for now).

If we have:

l: number of timeseries
p: number of datapoints in a timeseries

We can have the following shapes:

```
p: only y-values
p, 2: yz vals
l, p: l-timeseries with y values
l, p 2: l-timeseries with yz values
```

Extended to n-dimensional arrays (for example, trajectories projected onto principal components?). If each non-timeseries dim is $d_1, d_2, ... d_n$, then the above becomes:

```math
\begin{align*}
d_1, ... d_n, p\\
d_1, ... d_n, p, 2\\
d_1, ... d_n, l, p\\
d_1, ... d_n, l, p, 2\\
\end{align*}
```

I don't think we can auto-detect if `l` is present or not and the user should specify, something like:

`multi_timeseries: bool = True`

Scatters can be similar to some cases of nd-lines :thinking: , but x values would be directly specified and the current index is parametric (example with time indicating position in a low dim space). This would actually be true for lines as well if representing trajectories.

So for nd-line maybe we have two versions, parametric (y and z are not functions of x, but x, y, z are a function of some other dim) and non-parametric (simple timeseries lines where y and z are functions of x).

kushalkolar:
I made a more generalist `NDPositions` which can map data that is:

`[s1, s2, ... sn, l, p, 2 | 3]`

where: 
`s1, s2, .... sn` are slider dims
`l` is number of lines or number of scatters, this dimension is optional and the user must specify whether or not it exists
`2 | 3` is the last dim, indicating xy or xyz positions.

It can map arrays of these dims to a line, line collection, line stack, scatter, or list of scatters (similar to multi-line).

I think this is a much more elegant way to deal with things, and `NDTimeSeries` is not necessary. The user can provide a slider mapping (to map from reference units to array index) for the `p` dimension which is the same as the "x-axis" for time series data! 

Example if we have data that is `[n_timepoints, 2]`, and the x-positions here (in the last dim) are in seconds. The `NDWidget` reference units for the slider can also be in seconds, and we can provide a mapping function that goes from the slider reference units to the `n_timepoints` index.

I think we can also use this for heatmaps and interpolation. Use the reference units to determine a uniform x-range for the current display window, and we can interpolate using `[n_timepoints, 2]` data.

EDIT: I think that the `NDPositions` will also work for `PolygonGraphic` ! Can think about meshes in general later.

kushalkolar:
For positions graphics, I should actually do `[n_datapoints, n_lines, 2 | 3]` so everything before the last 1 or 2 dims is always sliders.

kushalkolar:
Some more ideas:

Allow any 2-3 dims to be used as the graphic dimensions and specify the slider dims. This would also allow using named dims (such as those used in xarray).
API like:

```
add_nd_<scatter|lines|heatmap>(
  data=<array>, # array like, defaults to last 2-3 dims are graphical, first (n - 2 | 3) dims are sliders and in order
  x=<int | str | None>  # optionally specify x graphical dim using the integer dim index or dim name
  y = ... # same for other graphical dims
  slider_dims=<None | tuple[int | str]> # specify slider dims, None is auto first (n - 2 | 3), or specify a tuple of slider dims as int or str, examples: (0, 1, 2, 3), ("time", "plane")
  ... # other args
)
```

We interpret the given order of the `slider_dims` passed as $$s_1, s_2, ... s_n$$, regardless of their order within the actual array. This will make it clear to users which dims they are syncing when they use the sliders.

EDIT: A limitation of the above is that a user can't collapse multiple "graphic/display dimensions" into "final graphic/display dimensions" if they're hard-coded this like. So something like:

```
add_nd_<...>(
  data=<array>,
  display_dims=<tuple[int | str]>, # specify display dims in [x, y, z] order, OR display dims that collapse to xyz after the finalizer function
...
)
```

An example for images would be collapsing `[z, m, n]` to display a projection over z as slider dims are moved. But we could also have > 3 dims that are used, and then collapsed to 3 or fewer dims for the "final graphic/display dims".

kushalkolar:
We can use `LineCollections` to display multiple lines, like behavior tracks of keypoints, with shape `[n_keypoints, n_timepoints, 2]`. This works well with `NDPositions` and display windows.

I was thinking of what's the best way to show a scatter for each keypoint, and I think I should make a `ScatterCollection` that behaves like a `LineCollection` so the same array with the same shape can be given, the only difference is the graphical representation would be a scatter instead of lines. For typical behavior keypoint viz, the `display_window` would usually be just 1, but it can be greater than 1 for any viz that needs to show a window of scatter points.

kushalkolar:
ok I think stuff is working

https://github.com/user-attachments/assets/e16f0803-ca67-4389-a3aa-e67b65030657

kushalkolar:
I think I need to make a `PolygonCollection` too :thinking: . Would be very similar to the `ScatterCollection`.

kushalkolar:
so the basics are all working :partying_face: 

LineStack and heatmap representations are swappable:

https://github.com/user-attachments/assets/5a6075c5-cc4b-43e6-82fc-59d935c96e82

Scatter and line collection to show behavior trajectories:

https://github.com/user-attachments/assets/01925425-fd1e-493a-938c-2f20aa956b30

image

kushalkolar:
A set of imgui UIs that allow controlling some aspects of the "nd graphics" could be useful, such as:
- graphical representation, dropdown menu to choose scatter, line, heatmap
- display_window
- multi bool

kushalkolar:
## Stuff I should finish before implementing the orchestrator:

- [x] merge #974 , so I can implement changing the `display_window`
- ~implement logic to allow the `display_window` to centered, on the left or on the right~
- [x] adding a linear selector, it's "default" positions and its position when the main slider for the `n_datapoints` dimension changes should correspond whether the `display_window` is in "center", "start" or "end" mode. Moving the linear selector should probably change the index in all other NDobjects but not its "parent" object. So that the visual representation of the parent object doesn't change, ex: if the linear selector is moved we don't want a linestack to also "move"
- [x] spatial function
- [x] implement mapping from a slider reference index with units (such as time) to array index. 
- [x] figure out how to implement stuff like colors, cmaps given that we're essentially performing out of core rendering with these graphics
- [x] auto `display_window`? Zoom in/out and set graphic data at different subsample levels

## Things that make me "uncomfortable" that I need to settle:

### dim shapes

Dim shape for nd-positions is `[s1, s2, ... sn, n, p, 2 | 3]` where:

`n`: number of lines, scatters, or heatmap rows (optional, can be 1)
`p`: number of datapoints, often corresponds to a time index.

`p` would be a slider dim, but there's this `n` dim that's in between `p` and all the other slider dims. We could instead use a shape of `[s1, ... sn, p, n, 2 | 3]` but that feels weird and we'd have to do a transpose to get the array for the graphical representation, i.e. `[p, n, 2 | 3]` -> `[n, p, 2 | 3]` is required for the graphical representation. 

Do we just document this well, that the `n` dimension is not a slider dim but `p` is? 

When using nd-positions data in conjunction with nd-image data, we'd have something like this:

```
# nd-positions
[s1, ... sn, n, p, 2 | 3]

# nd-image
[s1, ... sn, p, r, c, 1 | 3 | 4]
```

Where `r`: rows, `cols`: columns. `1 | 3 | 4` denotes grayscale or RGB(A).

The `p` dimension in the nd-image array above would correspond to the `p` in the nd-positions. nd-images don't have any `n` dimension, it doesn't make sense there.

kushalkolar:
Working on "implement mapping from a slider reference index with units (such as time) to array index.", which requires proper implementation of `slider_dims` and `n_slider_dims` properties. Need to figure out how to properly separate the `p` n_datapoints dimension from other slider dims, and also apply the window funcs on the `p` dim. Will do tomorrow.

kushalkolar:
Window funcs working for `p` (n_datapoints) dim and all graphical representations 

https://github.com/user-attachments/assets/43fb79e2-914a-4bf6-849c-cce6b0c50b00

clewis7:
The videos all look really cool! 

I like that you can arbitrarily change the window size

I like the switching between LineStack and heatmap too  

Some of the names are a bit long/weird; I would maybe consider making the property setters that users would interact with shorter/simpler

Other things that stand out to me:
* Setting the `indices` with a tuple `(change["new"])` 
* Same thing for a setting the window func (I think in the past we have done a dictionary approach to represent this kind of mapping)

kushalkolar:
Some thoughts:

Need to have a max limit beyond which window funcs are ignored. Now with OOC we can view timeseries that is tens of GBs+ in size, but we wouldn't want to put the `display_window` and the entire dataset AND apply a window function using all the data, it would be very slow. At the same time, it doesn't make sense to apply the `window_func` on a subsample of the full data, because then the `window_size` is no longer valid. So we should only use the `window_func` below a certain settable limit for `n_datapoints`. Maybe by default it can be `max_p * some_number`.

kushalkolar:
>  figure out how to implement stuff like colors, cmaps given that we're essentially performing out of core rendering with these graphics

I think that if a `cmap_transform` is provided for the "global" dataset, it should look correct even within the display windows.

kushalkolar:
ok next I'm going to an "auto dynamic display_window" which sets the `display_window` based on the current x-range. Panning and zooming will set the `display_window`. This also makes the linear selector management much easier! This will be specifically useful for timeseries data.

I think the way to do this is:

- Animation function to poll current extent, this is is world space. Ignore if unchanged. 
- Since we already map ND data from its own "model index" to array index, the world space is actually identical to the ND data's "model index". So we can use the same mapping to set the set the current visible window of graphic data. 

Keep manually settable `display_window` for data that are spatial in xy, such as behavior tracks.

kushalkolar:
Here's an example of how to the OOC timeseries with some toy data:

```python
n = 100
p = 10_000_000
xs = np.linspace(0, 100_000 * np.pi, p, dtype=np.float32)# + np.random.normal(0, 0.2, size=p)

def xs_mapping(x) -> int:
    """maps world space x position -> array index"""
    return int((x * 100) // np.pi)

xs_mapping(xs[132154])

data = np.zeros((n, p, 2), dtype=np.float32)
for i in range(n):
    data[i] = np.column_stack([xs, np.sin(xs * 5) + 0 + np.random.normal(scale=0.5, size=p)])


ndp = NDPositions(data, graphic=fpl.LineStack, display_window=1000, multi=True)

fig = fpl.Figure()

fig[0, 0].add_graphic(ndp.graphic)

xrange = (0, 10)

def update_view():
    xmin, ymin, xmax, ymax = fig[0, 0].frame.rect
    yavg = (ymax - ymin) / 2
    
    xrange_world = [fig[0, 0].map_screen_to_world((x, y))[0] for x, y in [(xmin + 1, yavg), (xmax - 1, yavg)]]
    global xrange
    if xrange_world == xrange:
        return
    xrange = xrange_world

    x0, x1 = xs_mapping(xrange[0]), xs_mapping(xrange[1])
    
    disp_window = x1 - x0
    ndp.display_window = disp_window
    ndp.indices = (int(np.mean([x0, x1])),)


fig.add_animations(update_view)
```

I think we can have a `display_window` that can be set as "auto-x" which is specific only for timeseries, since we can map the visible x-range in world space directly to array indices. This isn't possible with arbitrary ND-position representations, such as behavior tracks in xy space.

The position of the linear selector and index should be independent of the currently rendered portion of the timeseries data. So we can have an additional property, such as `visible_indices`, which is separate from `indices`. The `LinearSelector` sets the `indices`. Can think properly about this later. 

In the orchestrator we should hide these details from the user, perhaps by separating arbitrary nd positions from timeseries representation which set ``display_window="auto-x"` internally, as well as add the update func as an animation function.

kushalkolar:
I can make a subclass of ndpositions specific to timeseries data which also allows specifying xs and ys separately. It should also allow specifying arrays of shape [s1,... sn, n, p, 1 | 2] where p is the size of the ys, n is the number of lines to be stacked. The processor will then slice xs and ys (or the array of shape shown above) separately and then column_stack before passing it off to the super class to do the rest (window func etc). Dask seems like it's too slow or uses too much ram for doing this kind of lazy slicing and stacking.

kushalkolar:
Might be useful to have a non-isolated buffer here so `NDPositions` can directly use a buffer created by `NDPositionsProcessor`. `NDPositionsProcessor` could managed the buffer :thinking: . This would avoid a copy step.

kushalkolar:
Next stuff to do:

Pull in ND Image branch! 

Think about and implement mapping of input data to other graphical buffers, like colors or sizes, in NDProcessor. Maybe return a dict where the keys define the features to set and values define new values for the feature? These could map to colors, markers, sizes, etc. 

For the base `NDPositionsProcessors` and any array-based ones, we could allow kwargs like `colors`, `sizes`, `alpha`, `cmap_transform`, `markers` that can be set to dimension in the data array.

- `colors`: this could be a dimension of size 3 | 4 indicating RGB(A) color
- `cmap_transform`, `alpha`, `sizes`: dimension of size 1 to get 1D values that map to datapoints
- `markers`: also dimension of size 1, use a int mapping to define shape values

When we have a slice, we can get the corresponding values for these features by embedding them in the very last dimensions. 

For dataframes this will be much easier, just provide column names. Need to look into xarray, it'll probably be easier there too.

Similarly, allow using an NDProcessor method to set tooltips. NDProcessor can have an optional method that spits out tooltip info for a displayed data point index.
The current display_window indices in array-index-space are a property of NDProcessor
NDProcessor.tooltip_info() takes in a graphic datapoint index (whatever is given by pick_info["vertex_index"] or pick_info["index"]), and spits out tooltip info. It has access to an NDProcessor.displayed_indices property so that it can get any relevant info based on however the input is formatted.

kushalkolar:
Idea for dataframes, pass in a dataframe and a list of 2-tuples or 3-tuples indicating columns for (x, y, z) vals. Allow passing the names of other columns that are used to map to other graphic features (colors, sizes, markers, etc.) or tooltips. Kinda like seaborn. This way we don't need a specific NDProcessor for deeplabcut-style results.

kushalkolar:
I need to restore the current ImageWidget in main so we can deprecate and remove it in a future release.

kushalkolar:
List of thoughts:

- [ ] allow multiple histograms in a subplot, name them, correspond to individual images or heatmaps
- [x] reference ranges object instead of just a list of tuples
- [ ] Relatedly, indices object that has a nice repr, keeps tracks of the reference range objects etc. I think we should be able to push and pop dimensions on this object. If a reference range isn't specified we should just use the max for each slider dim for all ND graphics in the NDWidget. Each ndgraphic could also have this index class but only with info for the dims that graphic has. Make a nice repr for ndgraphics which also has info of the processor. 
- [x] auto-setting the x-range on timeseries data. Can probably make subplot properties like `x_range` etc.
- [x] the position of the heatmap seems wrong, should check if LineStack is the same.
- [x] tidy up missing stuff in NDProcessors like window functions and spatial function things, clean things up so window computing is done in a private method in the base class etc. 
- [x] linear selector stuff for time series
- [ ] edge cases: how to deal with when `NDProcessor` returns empty array

I think I can make the linear selector not respond to index changes when it the selector itself changes the index by using pause_graphics on the event handler that sets the linear selector index within the NDGraphic

kushalkolar:
basic orchestration of ndgraphics works :sob: 

@Vipitis you do some point tracking work too right? 

https://github.com/user-attachments/assets/e7c890c7-24a9-4062-ba19-781632de3e5b

```python
# start, stop, step range for time
 reference_range = [(vid_indexing[0], vid_indexing[-1], 0.025, "time (s)")]
 
 # Create an ND Widget with the reference dimensions
 ndw = fpl.NDWidget(ref_ranges=reference_range, shape=(1, 2))
 
 # add video as an nd image
 ndw[0, 0].add_nd_image(
     vid, 
     index_mappings=(vid_second_to_index,), 
     processor_kwargs={"compute_histogram": False, "rgb": True}
 )
 
 # add behavior data as an nd scatter
 nd_scatter = ndw[0, 0].add_nd_scatter(
     df_tracks,  # tracks dataframe
     keypoints_cols[:, :-1],  # provide the columns to take the tracks data from in the form [(keypoint_n_x), (keypoint_n_y), ...], 
     likelihood_cols,  # show likelihood as tooltips
     processor=ndp_extras.NDPP_Pandas, # use the processor for pandas dataframes
     display_window=5,
     index_mappings=(tracks_second_to_index,)  # time (s) -> array index mapping
 )
```

kushalkolar:
For NDPositions for other graphic features like colors, sizes, markers, cmap_transform, etc, maybe the way to do it is to just pass an additional array of the same shape as the data (i.e. s1,... sq, n, p, _) but with the corresponding feature values instead. And then we can use the same index and display window to get the values for those features at the given indices. 

Could also take a mapping function that takes the current indices and current data_slice array to create the feature values on the fly.

kushalkolar:
I just noticed these weird glitches on my desktop on both my AMD and Nvidia GPU. Never noticed it before. I wonder if it's related to replacing the buffer when the number of datapoints changes but I'm not sure. The line turns red whenever the buffer has been replaced. One or more lines is rendered according to some previous positions data and not the current one.

I checked the values in the actual `pygfx.Buffer` and they are identical between the lines so IDK why they are glitching in the wrong locations. Hard to reproduce, seems sporadic. The only difference between the lines is the y-offset position (i.e. `WorldObject.world.y`)

@almarklein any guesses?

EDIT: I posted on pygfx

https://github.com/user-attachments/assets/f1ce34e3-2a40-4d6a-bcf7-7bca6037d604

kushalkolar:
linear selector stuff working :smile: 

https://github.com/user-attachments/assets/5a418721-8f7c-48b8-bc21-9d847a71d6f9

@clewis7 very clever for suggesting I should just keep the same linear selector logic and set the `display_window` to the entire data to get the way we usually use linear selectors on an entire timeseries :partying_face: 

https://github.com/user-attachments/assets/a5378515-8f3c-484e-a484-0c1cca337c74

kushalkolar:
I think github is broken, there's a merge conflict which I solved locally and then pushed. Those changes are in the branch (can see on github if I go to that branch), but it's not showing up on the commits list here. And when I try to resolve the conflicts here on the PR it keeps saying "this page is out of date" and I make the changes :man_facepalming:

kushalkolar:
seems like github is working again, the commits have shown up in the PR :shrug:

kushalkolar:
I just realized I think we can do subplot.frame.rect to get the number of pixels in x across the subplot and use that to figure out a good auto number of data points to display for OOC rendering and LOD of timeseries data. 

EDIT: this needs more thought, easy to compute if it's a single line across x but more compicated for real data lol
EDIT2: rendering is just sampling so I think this is a typical sampling problem that can be solved

kushalkolar:
We can have something like an index for even the l dim (number of lines, rows, scatters, etc) and p (n datapoints) dim, (context: last 3 dims are l x p x 2 | 3).
This would basically be a selection vector! Like neuron or trace or keypoint or specific timepoint. And we could allow multiple such "global selection vectors" in a NDWidget to represent selection of neurons, keypoints, or anything that has a graphical representation! This is gonna abstract away even more event code :smile: 

When you plot contours on a movie and traces, that's l x p x 2. l is the same for both, it's number of components. We can use a selection vector to represent interaction with them. And if you have behavior, you can use another selection vector for keypoints. When you add an NDgraphic, you just specify which selection vector it's on. If looking at calcium and behavior you give the calcium arrays all the same component/neuron selection vector, and you give the behavior related arrays keypoint or whatever relevant selection vector that corresponds to the l dim

 This way we don't event have to write event code for even this interactivity, it gets even more declarative.

Can also think of how to use the selector tools (rect and polygon) to auto-update the selection vectors.

FlynnOConnell:
- checkbox for show/hide histograms
- ndwidget control tab UI with list of ndgraphics, hover graphic and highlight corresponding graphic
- options and sliders to change display window, no display window (see full data), graphic features
- show active ndgraphic with new graphic that has slightly different opacity or something like this
- toggle between display window as a slider vs. with panning & zooming

kushalkolar:
ok I need to do the following:

~`NDGraphics` need eventing for the `indices`. Need to allow adding event handlers to get when the indices change. Right now this is necessary so that `NDWidget` knows when a graphic internally changes its indices for any reason so that the global index can be changed too.~

Better idea: `GlobalIndexVector` object. It also has the references ranges instances. 

Each `NDGraphic` gets the instance of `GlobalIndexVector`. When the graphic has to change the index, it changes it via the `global_index_vector` instance only. The `global_index_vector` then updates all graphics.

Need to draw this out properly and take linear selectors into consideration as well.

fastplotlib/ui/_base.py
kushalkolar, line 1: @FlynnOConnell you do more imgui than any of us, can you please review this.

FlynnOConnell: @kushalkolar should be able to review this today

examples/ndwidget/ndimage.py
kushalkolar, line 1: @clewis7 this example and the timeseries one below is the kind of end user API I'm going for

kushalkolar:
For a given `NDGraphic`, if more indices are provided than exist for that graphic's `n_slider_dims`, just use the right most `n_slider_dims` that exist for that graphic's processor.

If an array is added with more slider dims than in the global index vector, push dim by adding `ReferenceRangeContinuous` using that array's size in the new dim. Set the reference range `unit` as `None` for that dim to indicate these are just direct array indices and are "auto managed", increase the stop dim if a new array is added with a different size in that dim.

Auto pop dim from `GlobalIndexVector` if its `unit` is `None`.

FlynnOConnell:
Running the timeseries.py example

1. Turning off display window
```
  File "C:\Users\flynn\repos\fastplotlib\fastplotlib\widgets\nd_widget\_nd_positions\core.py", line 442, in indices
    data_slice = self.processor.get(indices)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\flynn\repos\fastplotlib\fastplotlib\widgets\nd_widget\_nd_positions\core.py", line 283, in get
    graphic_data = window_output[tuple(slices)]
                                       ^^^^^^
UnboundLocalError: cannot access local variable 'slices' where it is not associated with a value
``` 
2. I got this somewhere but can't reproduce it
```
  File "C:\Users\flynn\repos\fastplotlib\fastplotlib\widgets\nd_widget\_nd_positions\core.py", line 463, in indices
    xr = data_slice[0, 0, 0], data_slice[0, -1, 0]
         ~~~~~~~~~~^^^^^^^^^
IndexError: index 0 is out of bounds for axis 0 with size 0
```

kushalkolar:
@FlynnOConnell please don't push to this branch

ok force pushed to remove the commit, only myself and caitlin have push access to this branch now

FlynnOConnell:
> @FlynnOConnell please don't push to this branch
> 
> ok force pushed to remove the commit, only myself and caitlin have push access to this branch now

bruh i asked if you want me to push to this branch and you said yes

kushalkolar:
> > @FlynnOConnell please don't push to this branch
> > ok force pushed to remove the commit, only myself and caitlin have push access to this branch now
> 
> bruh i asked if you want me to push to this branch and you said yes

misheard/misspoke, no worries lol, been a long few days

fastplotlib/ui/_base.py
FlynnOConnell: @kushalkolar I'm not sure exactly what to review here, but I reviewed the timeseries.py and as we discussed last night, we've got different visions of how the UI here should look

I think it might be more productive to complete the API before worrying about the UI; flesh out more of the features that will be in this ndwidget properties menu

FlynnOConnell:
Not sure if this is a problem with my copy: 

```
flynn at pop-os in ~/repos/work
$ git clone git@github.com:fastplotlib/fastplotlib.git
...
Encountered 1 file(s) that should have been pointers, but weren't:
        examples/screenshots/imgui_top.png
```

fastplotlib/ui/_base.py
FlynnOConnell: > @kushalkolar I'm not sure exactly what to review here, but I reviewed the timeseries.py and as we discussed last night, we've got different visions of how the UI here should look
> 
> I think it might be more productive to complete the API before worrying about the UI; flesh out more of the features that will be in this ndwidget properties menu

I'm envisioning this ndwidget properties menu: 

image

As something more like this: 

image

Some kind of nested menu, where each subplot will have an entry

kushalkolar: ah so right click on a subplot and see stuff about the `NDGraphics` in it, yea I think that would be easier to navigate.

kushalkolar:
`xarray` has some nice things for transforms (our mappings), and indexer objects. However making custom transforms and indexer is still in their experimental API so I suggest we do the mappings and index management ourselves. If `xarray` stabilizes these parts of their API we can look into it in the future. 

https://xarray-indexes.readthedocs.io/blocks/transform.html
https://docs.xarray.dev/en/latest/internals/how-to-create-custom-index.html

xarray's `array.sel(indices, method="nearest")` is very slow compared to plain `np.searchsorted()` so we will manage the mappings ourselves.

By defining the spatial dims xarray could also help with OOC rendering if it can wrap arrays that are lazy loadable.

kushalkolar:
ok xarray migration is halfways there. Works with `NDPositions`

Main things left to do before more progress with users
- [x] migrate NDImage to xarray
- [x] `spatial_func`
- [x] display histogram for image and heatmap timeseries
- [x] migrate any other useful features from old nd-ImageWidget
- [ ] cleanup API for users
  - [x] subplot API, `add_nd_<graphic>` methods mostly
  - [x] properties on `NDGraphic` and `NDImage` that mostly alias the `NDProcessor`
  - [x] `NDWidget`, `ReferenceRanges`, and how users define reference ranges
  - [ ] input validation
- [ ] other graphical features (colors, etc.)
- [x] test NDPositions with real nd-data, stack of sines which vary in ampl, freq etc., write some examples so it can be used to verify/iterate longterm
- [x] test NDImage

kushalkolar:
ok got the behavior viz working again, now with named dims.

@apasarkar, @FlynnOConnell  you will have to adjust your lazy array base classes slightly to work with xarray:

You need to add these methods, they don't actually have to be implemented
```python
    def __array__(self, dtype=None, copy=None):
        if copy:
            return copy(self)
            
        return self

    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
        raise NotImplementedError

    def __array_function__(self, func, types, *args, **kwargs):
        raise NotImplementedError
```

You will need to account for an `Ellipsis` object as the last object in the indexing tuple. You may need to adjust any validation logic accordingly, but I _think_ that if you just ignore/get rid of the last object in the tuple (the `Ellipsis` obj) you're fine.

```python
    def __getitem__(self, indices: tuple[slice, ...]) -> np.ndarray:
        # indices can be a tuple of slice | Ellipsis
        # need to accoutn for Ellipsis as the last object in the tuple
        pass
```

example implementation with the decord lazy video reader wrapper:

```python
class LazyVideo(LazyArray):
    def __init__(
        self,
        path: Union[Path, str],
        min_max: Tuple[int, int] = None,
        **kwargs,
    ):
        self._video_reader = VideoReader(str(path), **kwargs)

        try:
            frame0 = self._video_reader[10].asnumpy()
            self._video_reader.seek(0)
        except IndexError:
            frame0 = self._video_reader[0].asnumpy()
            self._video_reader.seek(0)

        self._shape = (self._video_reader._num_frame, *frame0.shape)

        self._dtype = frame0.dtype

        if min_max is not None:
            self._min, self._max = min_max
        else:
            self._min = frame0.min()
            self._max = frame0.max()

    @property
    def dtype(self) -> str:
        return self._dtype

    @property
    def shape(self) -> Tuple[int, int, int]:
        """[n_frames, m, n, 3 | 4]"""
        return self._shape

    @property
    def min(self) -> float:
        warn("min not implemented for LazyTiff, returning min of 0th index")
        return self._min

    @property
    def max(self) -> float:
        warn("max not implemented for LazyTiff, returning min of 0th index")
        return self._max

    @lru_cache(maxsize=32)
    def __getitem__(self, indices: tuple[slice, ...]) -> np.ndarray:
        # indices can be a tuple of slice | Ellipsis

        # apply to frame index, then remaining dims
        a = self._video_reader[indices[0]].asnumpy()[indices[1:]]
        self._video_reader.seek(0)
        return a
```

kushalkolar:
ideas:

If a ref range is not specified for a slider dim, it auto makes one just using the identity transform using the max of that array for that slider dim.
default_image_dims and default_positional_dims kwargs to NDWidget could be used, better than a config and hard-coded guessing

kushalkolar:
ok update imgui sliders UI, need to do right click menus for NDGraphic properties, as well as some image properties like brightness, contrast, gamma for any ImageGraphic in general.

fastplotlib/ui/_base.py
```diff
@@ -180,37 +191,205 @@ def get_rect(self) -> tuple[int, int, int, int]:
...
+        # Centered text
+        text_pos = imgui.ImVec2(
+            box_screen_pos.x + (win_width - text_size.x) * 0.5,
+            box_screen_pos.y + padding.y,
+        )
+        draw_list.add_text(
+            text_pos, imgui.get_color_u32(imgui.ImVec4(1, 1, 1, 1)), title
+        )
+
+        imgui.dummy(imgui.ImVec2(win_width, box_size.y))
```
kushalkolar, line 372: As per our AI use disclosure rules, I used the claude LLM to get a starting point for the imgui resize splitter code, but as usual with LLMs the output was mostly hot garbage and I rewrote most of it to make it work. Since it's not any core backend logic and just imgui UI I figured use of an LLM as a starting point for code gen was fine.

kushalkolar:
nd graphic options are now accessible through a right click menu popup:

image

Which then opens a window to change stuff until it's closed explicitly:
image

I think these kinds of floating windows can be pretty useful for many things :thinking: @FlynnOConnell 

Checkboxes in the right click window to select the spatial dims.

kushalkolar:
Another thing I should look into, when the `display_window` and indices auto-adjusts for timeseries by polling the `Subplot.x_range`, it also should adjust for other subplots that share the same controller. Useful for @GilyGinosar spectrograms

kushalkolar:
Comments from Caitlin:

- if ref range is smaller than the real data, we should raise a warning, need to think about how to do this when there are index mappings

kushalkolar:
Idea: I wonder if the `l` dimension size can be dynamic for `NDPositions`. Would be useful for line stacks to show only a few traces.

kushalkolar:
ok so TIL broadcasting via fancy indexing makes a copy:

```python
# this makes a copy
a[np.zeros(10000000, dtype=int), ...]
```

but explicitly `broadcast_to` just creates a view, so I can do this which would be over 1 exabytes if it all existed in RAM:

```python
np.broadcast_to(c1, (1000000000000000, 105, 4))
```

kushalkolar:
~weird graphics artifacts with markers, will figure out later, but basics work~

Edit: Fixed

~There is a weird thing going on with the markers buffer O_O~

Need to separate `cmap` and `cmap_transform` as two distinct features to implement that on `NDGraphics`, it also needs access to the parent graphic for stuff like checking the data shape.

kushalkolar:
Idea! I think we can allow global indices to be shared between NDWidgets so then we can have multiple windows. 

Edit: this works!

fastplotlib/widgets/nd_widget/_ndw_subplot.py
kushalkolar, line 1: I now realize we can probably auto-gen `NDWSubplot.add_nd_graphic` methods similar to `Subplot.add_graphic` methods

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -0,0 +1,1147 @@
...
+                    0, limits=(-np.inf, np.inf), edge_color="cyan"
+                )
+                self._linear_selector.add_event_handler(
+                    self._linear_selector_handler, "selection"
+                )
+                self._subplot.add_graphic(self._linear_selector)
+            else:
+                self._linear_selector = None
+        else:
+            self._linear_selector = None
```
kushalkolar, line 701: I think we should have an `NDTimeseries` subclass of `NDPositions` which has the timeseries specific stuff, i.e. `x_range_mode` `linear_selector`, etc., instead of messy if-else logic. 🤔

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -0,0 +1,1145 @@
...
+        self._last_x_range[:] = self.graphic._plot_area.x_range
+
+        if self._linear_selector is not None:
+            with pause_events(self._linear_selector):  # we don't want the linear selector change to update the indices
+                self._linear_selector.limits = xr
+                # linear selector acts on `p` dim
+                self._linear_selector.selection = indices[
+                    self.processor.spatial_dims[1]
+                ]
+
```
kushalkolar, line 835: similar to other comment, I think we should have an NDTimeseries subclass of NDPositions  🤔

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -0,0 +1,1145 @@
...
+        new_index = (xr[0] + xr[1]) / 2
+
+        if (new_index == self._ref_index[self.processor.spatial_dims[1]]) and (
+            last_width == new_width
+        ):
+            return
+
+        self.processor.display_window = new_width
+        # set the `p` dim on the global index vector
+        self._ref_index[self.processor.spatial_dims[1]] = new_index
```
kushalkolar, line 1017: similar to other comment, I think we should have an NDTimeseries subclass of NDPositions  🤔

fastplotlib/widgets/nd_widget/_nd_image.py
```diff
@@ -0,0 +1,544 @@
...
+        ),  # must be in order! [rows, cols] | [z, rows, cols]
+        rgb_dim: str | None = None,
+        window_funcs: tuple[WindowFuncCallable | None, ...] | WindowFuncCallable = None,
+        window_order: tuple[int, ...] = None,
+        spatial_func: Callable[[ArrayLike], ArrayLike] = None,
+        compute_histogram: bool = True,
+        slider_dim_transforms=None,
+        name: str = None,
+    ):
+        """
```
kushalkolar, line 300: What do you think about having the detailed docstring of the arguments only in the corresponding NDGraphic, ex: in `NDImage`, and not in the `NDImageProcessor` since people will usually look at the `NDImage` arguments (via the auto subplot.add_nd_<...> described in the PR comment)

fastplotlib/widgets/nd_widget/_base.py
```diff
@@ -0,0 +1,727 @@
...
+                 ``("keypoints", "time", "xyz")``
+
+            A custom subclass's ``data`` object doesn't necessarily need to have these dims, but the ``get()`` method
+            must operate as if these dimensions exist and return an array that matches the spatial dimensions.
+
+        spatial_dims: Sequence[str]
+            Subset of ``dims`` that are spatial (rendered) dimensions **in order**. All remaining dims are treated as
+            slider dims. See subclass for specific info.
+
+        slider_dim_transforms: dict mapping dim_name -> Callable, an ArrayLike, or None
```
kushalkolar, line 79: What do you think about the name ``index_mappings`` vs ``slider_dim_transforms`` or `slider_dim_mappings``? I was thinking `slider_dim_` prefix makes it explicit what it's about, and xarray uses the name transform for a similar concept (but they use it for ref-space -> array inedx mappings in spatial dimensions unlike us). I'm evently split lol :joy:

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -0,0 +1,1145 @@
...
+        # TODO: cleanup how this cmap stuff works, require a cmap to be set per-graphic
+        #  before allowing cmaps_transform, validate that stuff makes sense etc.
+        cmap: str = None,  # across the line/scatter collection
+        cmap_each: Sequence[str] = None,  # for each individual line/scatter
+        cmap_transform_each: np.ndarray = None,  # for each individual line/scatter
+        markers: np.ndarray = None,  # across the scatter collection, shape [l,]
+        markers_each: Sequence[str] = None,  # for each individual scatter, shape [l, p]
+        sizes: np.ndarray = None,  # across the scatter collection, shape [l,]
+        sizes_each: Sequence[float] = None,  # for each individual scatter, shape [l, p]
+        thickness: np.ndarray = None,  # for each line, shape [l,]
```
kushalkolar, line 594: What do you think of these args for the other graphic features. ``l`` is the "stack" dimension (i.e. number of lines or scatters in a collection or stack), ``p`` is the datapoints dim.

Also I think I will change colors to this:

```python
colors: np.ndarray, # of shape [l,], specify color for each line/scatter
colors_each: np.ndarray, # of shape [l, p], specify color for every datapoint of every line/scatter
```

fastplotlib/graphics/features/_base.py
```diff
@@ -320,7 +318,7 @@ def __repr__(self):
 def block_reentrance(set_value):
     # decorator to block re-entrant set_value methods
     # useful when creating complex, circular, bidirectional event graphs
-    def set_value_wrapper(self: GraphicFeature, graphic_or_key, value):
+    def set_value_wrapper(self: GraphicFeature, graphic_or_key, value, **kwargs):
```
kushalkolar, line 321: the kwarg is used by the `LinearRegionSelectionFeature`

fastplotlib/graphics/features/_selection_features.py
```diff
@@ -118,7 +118,7 @@ def axis(self) -> str:
         return self._axis
 
     @block_reentrance
-    def set_value(self, selector, value: Sequence[float]):
+    def set_value(self, selector, value: Sequence[float], *, change: str = "full"):
```
kushalkolar, line 121: For the new `HistogramLUTTool`, even logic was much easier if we know whether both `vmin`, `vmax` changed or only one of them.

fastplotlib/graphics/features/_volume.py
```diff
@@ -34,7 +34,7 @@ class TextureArrayVolume(GraphicFeature):
         },
     ]
 
-    def __init__(self, data, isolated_buffer: bool = True):
+    def __init__(self, data):
```
kushalkolar, line 37: an isolated buffer is always created now. There really isn't a usecase for non-isolated buffer IMO, user should use OOC rendering if the data is huge. In `NDWidget` the final array copy is also very fast anyways because we're always only ever looking at a slice of large data.

fastplotlib/tools/_histogram_lut.py
kushalkolar, line 1: This is a complete rewrite, the git diff viewer is useless lol

kushalkolar:
@clewis7

This also contains the replaceable buffers PR #974  which also needs a review but those bit of code are simpler and they're in the other parts of the codebase outside the ndwidget subpackage

the following modules/classes are ready for review:

base classes `NDProcessor` and `NDGraphic`: https://github.com/fastplotlib/fastplotlib/pull/971/changes#diff-38237d8750a6678e1e5ffe318e6ccd6144f1aeab91bfad9c1b803963da62b666

`NDImageProcessor` and `NDImage`: https://github.com/fastplotlib/fastplotlib/pull/971/changes#diff-af2caa1af4b5c04b6b406797c16c77699c807894e99c402ced57c3166f35a0a2

`ReferenceRange` stuff and `ReferenceIndex` (was previously called `GlobalIndex`, I renamed it because I thought it'd be intuitive for "reference index" to be in "reference space", and because people might think global means global variable especially when you could have multiple ndwidgets with different reference spaces they each mange if someone is going crazy): https://github.com/fastplotlib/fastplotlib/pull/971/changes#diff-ff78d59b530fa54e6290bf9472a461c955e2fdc8d8c78ac265f0563b3fc0532c

General thoughts on `NDWSubplot`, I think we should autogen this: https://github.com/fastplotlib/fastplotlib/pull/971/changes#diff-41e35095bf2376587f8c124262ab9c63abe820980b3832446d4c918e4de3fa8e 

NDPositionsProcessor and NDPositions aren't ready for review, but I had a few comments on those and would like your thoughts (should I do ND Timeseries subclass and the "_each" arts) 

random questions:

`ndw["bah"].add_ndimage()` or `ndw["bah"].add_nd_image()`, and likewise for the rest :laughing: 

~What do you think about the name `index_mappings` vs `slider_dim_transforms` or `slider_dim_mappings`? I was thinking `slider_dim_` prefix makes it explicit what it's about, and xarray uses the name transform for a similar concept (but they use it for ref-space -> array inedx mappings in spatial dimensions unlike us). I'm evenly split~

I think maybe `slider_maps`, explicit and short :thinking: 

any other naming things

kushalkolar:
After speaking with Edoardo I now see a clear use case for a slider dim inverse transforms and the `RangeDiscrete` :laughing:

fastplotlib/widgets/nd_widget/_base.py
FlynnOConnell: FWIW "transforms" makes me think of a transformation matrix or vector, mapping makes more sense to me

kushalkolar:
I think we can get rid of the xarray dependency and manage named arrays ourselves, solves some issues with xarray wrapping and might give a small performance boost which adds up when we have many arrays. 

First see if we get better performance by removing xarray, if not then maybe just keep it.

In the async branch I just did this which lets us index with named dims:

```python
    def _get_raw_data_slice(self, indices: dict[str, Any]) -> Generator[FutureArrayProtocol | ArrayProtocol, ArrayProtocol, ArrayProtocol]:
        """
        Base implementation to get the raw data slice from the xarray-wrapped array.
        Always yields to support async getters.
        """
        if len(self.slider_dims) > 0:
            indexer = self._get_slider_dims_indexer(indices)
            # get the data slice w.r.t. the desired windows
            # yield so this is async if the underlying array returns a FutureArray-like
            # we convert to a numpy array outside, not here, since that resolves the Future
            index_tuple = tuple(indexer.get(dim, slice(None)) for dim in self.dims)
            raw_slice = yield self._unwrapped_data[index_tuple]

        else:
            # return everything directly
            raw_slice = yield self._unwrapped_data[:]

        return raw_slice
```

And then for the final transpose we just need:

```python
int_dims = tuple(self.dims.index(d) for d in self.spatial_dims)
np.transpose(windowed_slice.squeeze(), int_dims)
```

kushalkolar:
ok no longer need xarray! :partying_face:

kushalkolar:
I'm also going to look into making the entire get data -> window funcs -> spatial funcs -> set on graphic async, probably just by threading. So we shouldn't block the main render thread while wait for this full pipeline. 

And need to have a way to mark that all new graphic data has been set and we are ready to render the next frame. 

Few more things:
- Don't convert to a numpy until until the very end, i.e. right before setting `graphic.data`. This way window_funcs, spatial_func etc. can also operate in torch/cupy/jax/etc. and on the GPU and things are moved off only at the very end and also remains fully async.
- `NDProcessor.pipeline` property that takes `tuple(dict[dim, window_size], func)` which does a fully customizable pipeline that takes precedence over `window_funcs` and `spatial_func` if provided. The pipeline `func` takes a sliced array with the given windows, the `dims`, `spatial_dims`, and must return the final array that maps to the graphic.

kushalkolar:
~Idea: When the NDProcessor is wrapping async data that uses shared memory we could use that shared memory directly as the graphic buffer. This has some limitations, ex: cannot tile images. Needs some thought to implement properly.~

EDIT: see #1042

examples/misc/buffer_replace_gc.py
clewis7, line 1: Add to the docstring whether this works or not

examples/misc/scatter_animation.py
clewis7, line 1: uniform_size -> size_mode

examples/ndwidget/ndimage.py
```diff
@@ -0,0 +1,54 @@
...
+    "time": (0, 1000, 1),
+    "depth": (0, 30, 1),
+}
+
+
+ndw = fpl.NDWidget(
+    ref_ranges=ref,
+    size=(700, 560)
+)
+ndw2 = fpl.NDWidget(
```
clewis7, line 29: add note here: sharing ref idxs between ndwidgets syncs them

examples/ndwidget/timeseries.py
```diff
@@ -0,0 +1,62 @@
...
+    "freq": (1, n_freqs + 1, 1),
+    "ampl": (1, n_ampls + 1, 1),
+    "angle": (0, xs[-1], 0.1),
+}
+
+ndw = fpl.NDWidget(ref_ranges=ref, size=(700, 560))
+
+nd_lines = ndw[0, 0].add_nd_timeseries(
+    data,
+    ("freq", "ampl", "n_lines", "angle", "d"),
```
clewis7, line 44: list with kwarg

FlynnOConnell:
why is the throttle on RangeContinuous defaulted to 200ms? That makes scrubbing through T noticably worse - I can set the throttle myself, just curious what the benefit is

kushalkolar:
> why is the throttle on RangeContinuous defaulted to 200ms? That makes scrubbing through T noticably worse - I can set the throttle myself, just curious what the benefit is

These aren't final values, high throttle is necessary for video decoding on cpu. I'll probably set a 100ms throttle when a video is added, maybe 50ms or lower throttle as default. Something to play with. 

Also throttling is only for the imgui slider. All other methods of updating the indices are not time-throttled. If you click the "play" button it will actually display every single frame, similarly if you use the linear selector or pan/zoom on a timeseries those updates are not throttled and they are immediate; it relies on the IO being fast and reliable in these cases, else it will block rendering if the IO is slow/inefficient/bad but that's on the IO not on NDWidget.

kushalkolar:
Ok torn out `NDTimeseries` from `NDPositions`. I was actually able to use latest claude with max effort to do the refactor but it took a few hours, it was 80% and 20% garbage and needed hand-holding to get it right. Basically seems like these are good enough to do simple refactors while I can be slightly lazier and mostly review & test.

clewis7:
This [commit](https://github.com/fastplotlib/fastplotlib/pull/971/commits/e7b149eb464eff9450bcaff234bf8840d32675e1) is for making  pygfx `Texture` objects that underlie fpl `ImageGraphic` to always be instantiated with wgpu usage args that allow an existing `Texture` to be written to and then sampled from 

For now, this is the minimum that allows my zero host round trip to work 

Other solution that might be more elegant: 
* Allow a user to pass these as kwargs that can then go to `TextureArray` 
     * I think the only problem with this right now is that all kwargs get passed to the `Graphic` parent class
     
     
Follow up on above: I think the best thing to do is just have a `texture_usage` kwarg to `ImageGraphic` that will go to the `TextureArray` 

Done in this [commit](https://github.com/fastplotlib/fastplotlib/pull/971/commits/a75b6bf0ddf7310fc3ccf528eba9890558856d8a)  

Let's the user define exactly what texture kwargs they want

kushalkolar:
> This [commit](https://github.com/fastplotlib/fastplotlib/pull/971/commits/e7b149eb464eff9450bcaff234bf8840d32675e1) is for making  pygfx `Texture` objects that underlie fpl `ImageGraphic` to always be instantiated with wgpu usage args that allow an existing `Texture` to be written to and then sampled from 
> 
> For now, this is the minimum that allows my zero host round trip to work 
> 
> Other solution that might be more elegant: 
> * Allow a user to pass these as kwargs that can then go to `TextureArray` 
>      * I think the only problem with this right now is that all kwargs get passed to the `Graphic` parent class  

Does it work if you pass in the fully or'd flags? 

If you do:

```python
TextureArray(..., usage=wgpu.TextureUsage.COPY_DST | wgpu.TextureUsage.TEXTURE_BINDING)
```

kushalkolar:
Ah I understand it now, it wasn't exposed in `Image`!

kushalkolar:
A FEW MORE BASIC EXAMPLE, TESTS, AND THEN WE GONNA MERGE THIS!

kushalkolar:
@clewis7 THIS IS READY TO MERGE!!!! :partying_face:

clewis7 (approved):
🥳 🥳 🥳  

@kushalkolar will let you do the honors of merging!!

PR #973, opened 2026-01-12
Share scenes between subplots

### LINKED ISSUE #952: Tracking example with detailed view
Hey,

I am having the following usecase: frame sequence (t, y, x) as well as bounding boxes annotations (for example centerXY, width, height), and maybe additional label/flags etc. This is for vizualizing tracking data in a video, perhaps I can find a public sample to share later.

So I figured something out using the ImageWidget to show 1. the full screen and 2. a crop into the bounding box. The important part is that bounding boxes change in **size and aspect ratio** over time. So I couldn't precrop them all into a 2nd array and also want to see some margins outside the bounding box which gets tricky on the edges. Instead I just changed the pygfx camera which felt a bit awkward. And also needed me to double the data and bounding box lines. I feel like there just just be one scene with two cameras/views (as the two different subplots).
Perhaps as an advanced user I would also love to write my own fragment shader where I get the first resulting texture (cmap, lines, zoom) and my array data as uniforms to output a texture that is displayed in a second sbplot/figure. So I can implement my own zoom and other effects.

here is the snippet without mock data just yet.
```py
# movie (t, y, x), bboxes (t, 4) xywh, lines (t, 5, 2)
iw = fpl.ImageWidget(data=[movie, movie], rgb=False, cmap="gray", figure_shape=(2,1), figure_kwargs={"size" : (650, 900)})
figure = iw.figure
bbox = figure[0,0].add_line(lines[0], colors="red")
bbox2 = figure[1,0].add_line(lines[0], colors="red", alpha=0.5)
figure[1,0].axes.visible = False
figure[1,0].toolbar = False
# TODO: disable camera control for crop, gang cmaps

def set_bbox(index):
    bbox.data[:,0:2] = lines[index["t"]]
    bbox2.data[:,0:2] = lines[index["t"]]

def set_crop(index):
    t = index["t"]
    long_edge = max(bboxes[t,2], bboxes[t,3])
    if long_edge: # to not crash on DIV by 0
      # zoom in so it's always showing the whole box + some margin
      scale = (1.0/long_edge) * 0.55
      # move the camera
      figure[1,0].camera.set_state({"position": (bboxes[t,0]*movie.shape[2], bboxes[t,1]*movie.shape[1], -1), "zoom": scale})

# TODO: initialize crop properly
# TODO labels as title

iw.add_event_handler(set_bbox)
iw.add_event_handler(set_crop)
iw.show()

```

I am writing the issue instead of a PR that adds an example to more ask a question if there ins't a better way. Also saw there is a rework for the image widget in progress but from quickly looking through that PR - I couldn't find something like this.

Vipitis:
Resolves #952 

I followed the suggestion and mostly copied the code for `controller_ids`. And doing the scene creation further up (but hopefully not breaking existing functionality).

Here is the current state by modifying one of the controller examples, I will try to find time to add like a multi object tracking example instead.

https://github.com/user-attachments/assets/324638d8-0547-4283-bdd8-cbe3f63ae7fe

With the Axes from other subplots being part of the scene it's a bit awkward. Not sure how viable it is, but having a few parts of the scene be "per subplot" (axes, screen space text labels, LUT) while the rest is shared (meshes, images, materials, etc)?... Not sure what pygfx enables - but I will think about it and have a look.

draft as there is still a lot of stuff broken, I will hopefully find some more free time next weekend/monday.

### DIFF
--- a/examples/scenes/multiobject_tracking.py
+++ b/examples/scenes/multiobject_tracking.py
@@ -1,0 +1,106 @@
+"""
+Vizualizing Multi-Object Tracking
+=================================
+
+By sharing the scene between subplots, we can dynamically add detail views for each detected object.
+"""
+# %% imports
+import fastplotlib as fpl
+import numpy as np
+import imageio.v3 as iio
+
+
+
+# %% Mock data
+# TODO: replace with YOLO-tracking/detection example?
+
+
+movie = iio.imread("imageio:cockatoo.mp4")
+# make shape (t, h, w, 1) (for monochrome examples)
+movie = movie[:,:,:, 0] # take one channel
+print(movie.shape)
+
+# boxes are are x,y,w,h... but there can be multiple per frame - so it's a list of boxes that can also be empty.
+center_path = np.array([[movie.shape[2]/2 + 200*np.sin(t/30), movie.shape[1]/2 + 50*np.cos(t/5)] for t in range(len(movie)-60)])
+width = 80 + 20*np.sin(np.linspace(0, 10*np.pi, len(movie)-60))
+height = 100 + 30*np.cos(np.linspace(0, 8*np.pi, len(movie)-60))
+bboxes = {t+30: [[center_path[t][0]-width[t]/2, center_path[t][1]-height[t]/2, width[t], height[t]]] for t in range(len(movie)-60)}
+# add a bit of jitter noise and some dropouts
+fake_preds = {}
+for t in bboxes:
+    fake_preds[t] = [box.copy() for box in bboxes[t]]
+    for box in fake_preds[t]:
+        box[0] += np.random.normal(scale=5.0)
+        box[1] += np.random.normal(scale=5.0)
+        box[2] += np.random.normal(scale=3.0)
+        box[3] += np.random.normal(scale=8.0)
+    # fake_preds[t] = [box.copy() for box in bboxes[t]]
+    if np.random.rand() < 0.1:
+        fake_preds[t] = []
+
+def get_lines(box) -> fpl.LineGraphic:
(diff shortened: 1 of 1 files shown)


kushalkolar:
A good start! The axes is funny 😂. Maybe an overlay render pass should be used for axes in orthographic projections, or have them fixed and intersect at the origin.

I think text is rendered fine, do you have an example where it's an issue? Also what do you mean by LUT here?

Vipitis:
instead of calling it `scene_ids` it would also work as `camera_ids` (Or views). I will try mixing orthographic and perspective cameras in a 3D scene and show how that works and looks, especially with text.

A lookup table would essentially be like a color map, but for the camera instead of the data/material. So you can have different views onto the same data.

kushalkolar:
actually I think we can do an independent scene per-subplot which has 2 scenes, one has the rulers and another is identical between the "subplots that share a scene but have different cameras". I think that would be the easiest solution.

Vipitis:
hey, I agree something like this would make most sense. Also had the idea of keeping one scene per plot but sharing graphics (including their world objects)... Not deep enough into pygfx to understand what's possible. 

Sadly I am a bit low on free time recently and this dropped really far down on my priorities. I don't want to be a blocker or anything so feel free to do close/commit etc if you want to. It will probably remain stale from my side for a few more weeks.

kushalkolar:
No rush! I just came up with the idea because a user just asked me about a use case similar to this.

kushalkolar:
No rush! I just came up with the idea because a user just asked me about a use case similar to this.

kushalkolar:
I think after the next release (probably tomorrow) could be a good time to get back to this if you want

PR #974, opened 2026-01-13
auto-replace buffers

### LINKED ISSUE #869: Create a new buffer if necessary when setting an existing Graphic
We can implement this pretty easily, would make things easier for users, tradeoff is that users are less aware that setting data this way is slower than setting data that fits in an existing buffer. We can just document it well.

```python
image = fig[0, 0].add_image(np.random.rand(10, 10))

# allow this, internally delete the existing buffer 
# and world object and create a new one
image.data = np.random.rand(20, 20)
```

Might not have to create a new world object, might be possible to just replace the geometry I think.

kushalkolar:
closes #869

Replace buffers when the user sets all the data with a new array that has a different number of positions (vertices) pr a different set of dimensions (images) than the existing buffer. Example, user sets all line or scatter data, `line.data = <bigger or smaller array>`.

Implemented in all `BufferManager` subclasses, and the image `TextureArrays`

- [x] `VertexPositions`
- [x] `VertexColors`
- [x] `VertexRotations`
- [x] `VertexPointSizes`
- [x] `VertexMarkers`
- [x] `MeshIndices`, gets it automatically from parent class `VertexPositions`
- [x] images and volumes - create an entirely new `TextureArray` if setting all the data with a new image that has different dims
- [x] removes `isolated_buffer`, OOC rendering is the better way to handle very large data
- [x] add `color_mode` property for positions graphics, will do `marker_mode`, `size_mode`, `edge_color_mode` in a future PR for scatters.
- [x] test deferencing
- [x] test garbage collection

Slicing and setting data, i.e. `graphic.data[<slice>] = <array>` is unchanged. New buffers are created only when the entire data is set with new data that needs a bigger or smaller buffer.

### DIFF
--- a/fastplotlib/graphics/__init__.py
+++ b/fastplotlib/graphics/__init__.py
@@ -7,7 +7,7 @@ from ._vectors import VectorsGraphic
 from .mesh import MeshGraphic, SurfaceGraphic, PolygonGraphic
 from .text import TextGraphic
 from .line_collection import LineCollection, LineStack
-
+from .scatter_collection import ScatterCollection
 
 __all__ = [
     "Graphic",
@@ -22,4 +22,5 @@ __all__ = [
     "TextGraphic",
     "LineCollection",
     "LineStack",
+    "ScatterCollection",
 ]
--- a/fastplotlib/graphics/scatter.py
+++ b/fastplotlib/graphics/scatter.py
@@ -53,7 +53,7 @@ class ScatterGraphic(PositionsGraphic):
         image: np.ndarray = None,
         point_rotations: float | np.ndarray = 0,
         point_rotation_mode: Literal["uniform", "vertex", "curve"] = "uniform",
-        sizes: float | np.ndarray | Sequence[float] = 1,
+        sizes: float | np.ndarray | Sequence[float] = 5,
         uniform_size: bool = True,
         size_space: str = "screen",
         **kwargs,
--- a/fastplotlib/graphics/scatter_collection.py
+++ b/fastplotlib/graphics/scatter_collection.py
@@ -1,0 +1,517 @@
+from typing import *
+
+import numpy as np
+
+import pygfx
+
+from ..utils import parse_cmap_values
+from ._collection_base import CollectionIndexer, GraphicCollection, CollectionFeature
+from .scatter import ScatterGraphic
+from .selectors import (
+    LinearRegionSelector,
+    LinearSelector,
+    RectangleSelector,
+    PolygonSelector,
+)
+
+
+class _ScatterCollectionProperties:
+    """Mix-in class for ScatterCollection properties"""
+
+    @property
+    def colors(self) -> CollectionFeature:
+        """get or set colors of scatters in the collection"""
+        return CollectionFeature(self.graphics, "colors")
+
+    @colors.setter
+    def colors(self, values: str | np.ndarray | tuple[float] | list[float] | list[str]):
+        if isinstance(values, str):
+            # set colors of all scatter to one str color
+            for g in self:
+                g.colors = values
+            return
+
+        elif all(isinstance(v, str) for v in values):
+            # individual str colors for each scatter
+            if not len(values) == len(self):
+                raise IndexError
+
+            for g, v in zip(self.graphics, values):
+                g.colors = v
+
+            return
+
+        if isinstance(values, np.ndarray):
+            if values.ndim == 2:
+                # assume individual colors for each
+                for g, v in zip(self, values):
+                    g.colors = v
+                return
+
+        elif len(values) == 4:
+            # assume RGBA
+            self.colors[:] = values
+
+        else:
+            # assume individual colors for each
+            for g, v in zip(self, values):
+                g.colors = v
+
+    @property
+    def data(self) -> CollectionFeature:
+        """get or set data of lines in the collection"""
+        return CollectionFeature(self.graphics, "data")
+
+    @data.setter
+    def data(self, values):
+        for g, v in zip(self, values):
+            g.data = v
+
+    @property
+    def cmap(self) -> CollectionFeature:
+        """
+        Get or set a cmap along the scatter collection.
+
+        Optionally set using a tuple ("cmap", <transform>) to set the transform.
+        Example:
+
+        scatter_collection.cmap = ("jet", sine_transform_vals, 0.7)
+
+        """
+        return CollectionFeature(self.graphics, "cmap")
+
+    @cmap.setter
+    def cmap(self, args):
+        if isinstance(args, str):
+            name = args
+            transform = None
+        elif len(args) == 1:
+            name = args[0]
+            transform = None
+        elif len(args) == 2:
+            name, transform = args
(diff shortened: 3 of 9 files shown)


kushalkolar:
OK I got the basics down for lines. One caveat is that now there's no check to make sure that the buffer sizes match for data and colors for example.

Consider this:

```python
lg = subplot.add_line(np.random.rand(100))
lg.data = np.random.rand(200)
```

Now the last 100 line points are just black since they're not defined. A user has to explicitly set the colors of the new vertices that extend beyond the previously defined buffer.

```python
lg.colors = ["w"] * 200
```

Or they can use a uniform color for things like this.

kushalkolar:
ok for images I think easiest way is to create new world objects :thinking: 

EDIT: Yup this will be fine, I was initially worried that we'd have to re-add event handlers if we create a new world object, but I don't think that's necessary since `ImageGraphic` and `ImageVolumeGraphic` use a `pygfx.Group` as the world object, so we can just clear the group and add new image tiles.

EDIT again: Just tested with image click events, if an event handler is added to the graphic then when new tiles are created the graphic is still responsive to those events because it's the `pygfx.Group` that is handling the events and not the individual image tiles.

kushalkolar:
@clewis7 would like to get your review before I do the gc tests

kushalkolar:
I think we should make uniforms the default for all graphics 🤔.

Also good time to start a "performance tips" section of the docs. Replacing the buffer has a huge performance hit so you don't want do it all the time.

kushalkolar:
I found a memory leak, if `old_buffer._wgpu_object.destroy()` is not called then it lingers on in GPU VRAM even if the Binding object no longer exists. 

Posted on pygfx: https://github.com/pygfx/pygfx/issues/1264 

For now I implemented manually destroying the `GPUBuffer` when it's replaced.

kushalkolar:
@clewis7 what do you think about:

For lines, default is:
`uniform_colors = True`

For scatters, default is:
`uniform_colors = False`

I think independent scatter colors is a more common usecase than lines. We can auto-determine it in the future based on the colors and cmap arg but I'll do that in the future.

EDIT: But then the API is inconsistent between lines and scatters :/

clewis7:
> OK I got the basics down for lines. One caveat is that now there's no check to make sure that the buffer sizes match for data and colors for example.
> 
> Consider this:
> 
> ```python
> lg = subplot.add_line(np.random.rand(100))
> lg.data = np.random.rand(200)
> ```
> 
> Now the last 100 line points are just black since they're not defined. A user has to explicitly set the colors of the new vertices that extend beyond the previously defined buffer.
> 
> ```python
> lg.colors = ["w"] * 200
> ```
> 
> Or they can use a uniform color for things like this.

I think this behavior is confusing. Perhaps it is better to just by default take the current color scheme and extend it to the new or fewer points.

clewis7:
> @clewis7 what do you think about:
> 
> For lines, default is: `uniform_colors = True`
> 
> For scatters, default is: `uniform_colors = False`
> 
> I think independent scatter colors is a more common usecase than lines. We can auto-determine it in the future based on the colors and cmap arg but I'll do that in the future.
> 
> EDIT: But then the API is inconsistent between lines and scatters :/

Hmmm, I do agree that scatters are more likley to have individual colors. How much a performance reduction is it if you make the lines also have `uniform_buffer=True` as the default??

kushalkolar:
Maybe we can do an auto mode :thinking: 
And have it change dynamically too when the `colors` feature is modified.

kushalkolar:
Got "auto" color mode working for lines and scatter, and can switch between vertex and uniform modes and the corresponding buffers are auto created and the previous one removed. Need to test that garbage collection works when switching from vertex -> uniform. 

I think I'm going to just implement this for colors for now. Will do scatter-specific markers, edge colors, and sizes later since I think expanding the data buffers isn't going to be too common. And when the user does expand the data buffer, they should also be aware that they should set the scatter point markers and other properties for the new points anyways. I'll work on on tests, testing this properly, make sure garbage collection works and then we merge because I need this for #971 . Dynamic `display_windows` require auto-replaceable buffers.

kushalkolar:
updated backend tests to account for `color_mode` and uniforms being default for everything, tests passing on my end

No clue why black show so many changes on github actions, I just blacked locally :shrug: . Maybe black changed things as usual, we can do that in another PR.

Need to update screenshot examples.

kushalkolar:
Added de-referencing tests and I think stuff works! 

Automated tests for check that GPU VRAM is freed seems hard. Manually checking as I showed here seems like the best way :shrug: https://github.com/pygfx/pygfx/issues/1264#issuecomment-3837748139

I tried using this: https://pypi.org/project/nvidia-ml-py/#description

But during offscreen tests it seems like GPU VRAM isn't being used :shrug: . I say let's figure this out later

I just need to add the following tests and do a few other small things:

- [x] image, test w.r.t. the tile sizes for tests
- [x] image, with examples gallery test
- [x] lines, scatters, and collections with example gallery tests
- [x] animation examples to show you can change buffer sizes but also note that it has performance drawbacks and should only be done when really necessary, or if you can take the performance hit when the buffer is replaced.
- [x] note in the docs about performance drawbacks
- [x] docstring on `color_mode`

kushalkolar:
OK so there's an issue with scatters (even with the "simple" mode that doesn't use `markers` or `edge_colors`) where I get a wgpu bind group error :sob: . The resource has been destroyed, so this manual destroying can cause issues. Will post to pygfx and figure out.

EDIT: ok seems like the dereferencing is clearing GPU VRAM, I guess it does it eventually, not sure what's up with https://github.com/pygfx/pygfx/issues/1264#issuecomment-3837748139 

Line and images seem to work, kept this running for 20 mins and it's still going.

https://github.com/user-attachments/assets/302c16d2-c408-44c6-8440-b551ce65b6ec

https://github.com/user-attachments/assets/8af1fb55-3028-4677-9110-a6ef858a24d9

Also watching `nvidia-smi` to see that there's no VRAM leak, after 20 mins no memory leak, goes up and back down as expected :smile: 

https://github.com/user-attachments/assets/aa80c698-165b-4b93-8ba0-05120bf44214

kushalkolar:
I think I might be able to get rid of the bindgroup issues with scatter by destroying the old buffer after we set the new buffer, instead of right now where it destroys first and then sets the new buffer.

kushalkolar:
ok interesting, without `wgpu_obj.destroy()`, seems like it's working for this example, but not the testing UI I made to check buffer gc :man_shrugging: 

https://github.com/user-attachments/assets/a76d4f69-a5cf-4586-bb33-813957a119d7

kushalkolar:
I think tests should pass now! :smile: 

@clewis7 I'm guessing you won't be able to review today so I will merge this into #971 , let me know if you have any objections XD . You can review it later as part of #971

clewis7:
> I think tests should pass now! 😄
> 
> @clewis7 I'm guessing you won't be able to review today so I will merge this into #971 , let me know if you have any objections XD . You can review it later as part of #971

No objections, can review later :D

PR #976, opened 2026-01-13
docs: note tooltips require `imgui` bundle to work

mokurin000:


kushalkolar:
imgui isn't required for tooltips, only for optional UI elements and this is already documented in the guide: https://fastplotlib.org/user_guide/guide.html#imgui

mokurin000:
> imgui isn't required for tooltips, only for optional UI elements and this is already documented in the guide: https://fastplotlib.org/user_guide/guide.html#imgui

Okay, I saw it documented as

> Fastplotlib comes built-in with imgui UIs for subplot toolbars and a standard right-click menu with a number of options. You can also make custom GUIs and embed them within the canvas, see the examples gallery for detailed examples.

But I still think it worthy to mention custom UI elements require imgui-bundle in the related example documentation

Anyway, since UI elements is not a necessary part of tooltips, closing this

Issue #977, opened 2026-01-14
Feat: register custom font for Dear ImGui


MeowKatee:
## Purpose

I want to include CJK characters in my figure name

<img width="700" height="560" alt="Image" src="https://github.com/user-attachments/assets/c07904f6-b41c-45fc-a648-e4422769d69a" />

The text inside control UI menu was messed up

## Current workaround

I did set font for `pygfx` to enforce it render CJK with a local font file

```python
from pygfx.utils.text import font_manager, FontProps

font = font_manager.add_font_file(
    "fonts/fusion-pixel-12px-proportional-zh_hans.ttf"
)
font_manager._default_font_props = FontProps(
    font.family,
    style="normal",
    weight="regular",
)
```

But looking into the source code, I think I cannot set default font for ImGui similarly, it must be passed to `Figure.__init__`

kushalkolar:
You can see how we loads the imgui fonts, if you have the ttf file _I think_ you can push a different font even after a figure has been created since it's part of the global imgui config: https://github.com/fastplotlib/fastplotlib/blob/359770a4c79d680ea486a2433612f02b8b3b7b85/fastplotlib/layouts/_imgui_figure.py#L69-L98 

@FlynnOConnell is more familiar with imgui and fonts

MeowKatee:
Okay, I will try do so

MeowKatee:
<img width="700" height="560" alt="Image" src="https://github.com/user-attachments/assets/193ba2ea-7f82-43ce-815b-ac7b2bed7c46" />

I tried to:
```python
figure = fpl.Figure(
    size=(700, 560),
    names=[log_name],
)

# fix CJK font for imgui
imgui_io = imgui.get_io()
ifont = imgui_io.fonts.add_font_from_file_ttf(
    font_path,
    12,
    imgui.ImFontConfig(),
)
imgui.push_font(ifont, 0.0)
```

But no luck

MeowKatee:
I also tried to enable merge mode for my font, but it raises:

```text
Traceback (most recent call last):
  File "C:\Users\kavin\Documents\Repos\socks-tls-ping\rtt-draw-fastplotlib.py", line 98, in <module>
    ifont = imgui_io.fonts.add_font_from_file_ttf(
        font_path,
        12,
        ifconf,
    )
RuntimeError: IM_ASSERT( Fonts.Size > 0 && "Cannot use MergeMode for the first font" )   ---   imgui_draw.cpp:3064
```

[...]

PR #978, opened 2026-01-14
feat: support passing custom fonts to imgui

MeowKatee:
close gh-977

image


MeowKatee:
Since this was added after sans serif and fp6 icon, users can only provide fallback fonts for non-existing glyths

I will provide an option for the ordering

mokurin000:
@kianmeng Hi! Could you take a review for this PR?
IMO it would be better to have custom font settings without users manually workaround. For example, the `custom_fonts` argument could work for imgui, and possible future UI backends.

Also, my first try of workaround didn't work, though I didn't dive it deeper, I did the same steps as what was in fastplotlib to push_font.

kianmeng:
@mokurin000 Not project owner, you can always ping @kushalkolar and @clewis7 ?

mokurin000:
> @mokurin000 Not project owner, you can always ping @kushalkolar and @clewis7 ?

Ahh sorry for accident mention, I mistaken identity of him

kushalkolar:
Sorry for the late response, busy with other priorities. I recommend going ahead with the approach [here](https://github.com/fastplotlib/fastplotlib/issues/977#issuecomment-3772591220). In general we recommend doing specialized imgui tasks directly with imgui as much as possible without adding extra things to the fastplotlib API that are imgui-specific.

mokurin000:
> Sorry for the late response, busy with other priorities. I recommend going ahead with the approach [here](https://github.com/fastplotlib/fastplotlib/issues/977#issuecomment-3772591220). In general we recommend doing specialized imgui tasks directly with imgui as much as possible without adding extra things to the fastplotlib API that are imgui-specific.

Okay, I was having concern about the default font change, but imo it's rarely get changed, as there are'nt real difference for font packs, and the PUA area code points have the de facto standard, so it should rarely break.

Issue #979, opened 2026-01-16
Incorrect assets loading when being packaged by nuikit, pyinstaller, e.g.


mokurin000:
The current implementation calculates path of font assets by `imgui_bundle.__file__`, 

https://github.com/fastplotlib/fastplotlib/blob/7075e46f79d23da43b2ca4658c15b5c8380c63d6/fastplotlib/layouts/_imgui_figure.py#L69-L79

which may be missing or unreliable when the `imgui_bundle` is frozen. (e.g. by Pyinstaller)

<img width="1041" height="332" alt="Image" src="https://github.com/user-attachments/assets/9b73c2bc-a261-408c-81b3-7a9ccb5c5b94" />

```python
from importlib import resources

with resources.as_file(
    resources.files(imgui_bundle).joinpath(...)
) as font_path:
    pass
```

This is a modern and recommended way to access bundled resource.

mokurin000:
imgui-bundle was having issue on GLFW3.dll detection, it's now get fixed in upstream,
it does find data resource directory based on `__file__` too, this should be fine with modern packaging tool and hooks

mokurin000:
Closing, as the current packaging tools would hook things up and zipapp is insufficient to handle `.pyd` dependencies.

PR #980, opened 2026-01-16
load default fonts from resouces correctly

mokurin000:
this fixes missing or unreliable `Path(imgui_bundle.__file__).parent` for frozen packaging.
close gh-979

### DIFF
--- a/fastplotlib/layouts/_imgui_figure.py
+++ b/fastplotlib/layouts/_imgui_figure.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from typing import Literal, Iterable
+from importlib import resources
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


kushalkolar:
Seems like https://github.com/pthom/imgui_bundle/issues/437 fixes this upstream, so fastplotlib with imgui can be bundled with pyinstaller or nuitka distributables?

mokurin000:
> Seems like [pthom/imgui_bundle#437](https://github.com/pthom/imgui_bundle/issues/437) fixes this upstream, so fastplotlib with imgui can be bundled with pyinstaller or nuitka distributables?

Yeah, they both does hack on assets bundle, the only problem was the DLL path looking.

PR #986, opened 2026-02-03
Fixes bug where tooltip info is incorrectly retrieved when tooltip in…

apasarkar:
…fo is not none


kushalkolar:
I think black changed let's do that later

PR #988, opened 2026-02-03
rename test that was never being run before

kushalkolar:
the events tests were never being run because the filename doesn't begin with `test`, now they should run :laughing:


kushalkolar:
@clewis7 gtg!

PR #990, opened 2026-02-04
settable `LineStack.separation`

### LINKED ISSUE #898: change separation of an existing `LineStack`
```python
class LineStack(LineCollection):
    def __init__(
        self,
        data: List[np.ndarray],
        thickness: float | Iterable[float] = 2.0,
        colors: str | Iterable[str] | np.ndarray | Iterable[np.ndarray] = "w",
        alpha: float = 1.0,
        cmap: Iterable[str] | str = None,
        cmap_transform: np.ndarray | List = None,
        name: str = None,
        names: list[str] = None,
        metadata: Any = None,
        metadatas: Sequence[Any] | np.ndarray = None,
        isolated_buffer: bool = True,
        separation: float = 10.0,
        separation_axis: str = "y",
        kwargs_lines: list[dict] = None,
        **kwargs,
    ):
        """
        Create a stack of :class:`.LineGraphic` that are separated along the "x" or "y" axis.

        Parameters
        ----------
        data: list of array-like
            List or array-like of multiple line data to plot

            | if ``list`` each item in the list must be a 1D, 2D, or 3D numpy array
            | if  array-like, must be of shape [n_lines, n_points_line, y | xy | xyz]

        thickness: float or Iterable of float, default 2.0
            | if ``float``, single thickness will be used for all lines
            | if ``list`` of ``float``, each value will apply to the individual lines

        colors: str, RGBA array, Iterable of RGBA array, or Iterable of str, default "w"
            | if single ``str`` such as "w", "r", "b", etc, represents a single color for all lines
            | if single ``RGBA array`` (tuple or list of size 4), represents a single color for all lines
            | if ``list`` of ``str``, represents color for each individual line, example ["w", "b", "r",...]
            | if ``RGBA array`` of shape [data_size, 4], represents a single RGBA array for each line

        alpha: float, optional
            alpha value for colors, if colors is a ``str``

        cmap: Iterable of str or str, optional
            | if ``str``, single cmap will be used for all lines
            | if ``list`` of ``str``, each cmap will apply to the individual lines

            .. note::
                ``cmap`` overrides any arguments passed to ``colors``

        cmap_transform: 1D array-like of numerical values, optional
            if provided, these values are used to map the colors from the cmap

        name: str, optional
            name of the line collection as a whole

        names: list[str], optional
            names of the individual lines in the collection, ``len(names)`` must equal ``len(data)``

        metadata: Any
            metadata associated with the collection as a whole

        metadatas: Iterable or array
            metadata for each individual line associated with this collection, this is for the user to manage.
            ``len(metadata)`` must be same as ``len(data)``

        separation: float, default 10
            space in between each line graphic in the stack

        separation_axis: str, default "y"
            axis in which the line graphics in the stack should be separated


        kwargs_lines: list[dict], optional
            list of kwargs passed to the individual lines, ``len(kwargs_lines)`` must equal ``len(data)``

        kwargs_collection
            kwargs for the collection, passed to GraphicCollection

        """
        super().__init__(
            data=data,
            thickness=thickness,
            colors=colors,
            alpha=alpha,
            cmap=cmap,
            cmap_transform=cmap_transform,
            name=name,
            names=names,
            metadata=metadata,
            metadatas=metadatas,
            isolated_buffer=isolated_buffer,
            kwargs_lines=kwargs_lines,
            **kwargs,
        )

        self._sepration_axis = separation_axis
        self._separation = separation

        self.separation = separation

    @property
    def separation(self) -> float:
        """distance between each line in the stack, in world space"""
        return self._separation

    @separation.setter
    def separation(self, value: float):
        separation = float(value)

        axis_zero = 0
        for i, line in enumerate(self.graphics):
            if self._sepration_axis == "x":
                line.offset = (axis_zero, *line.offset[1:])

            elif self._sepration_axis == "y":
                line.offset = (line.offset[0], axis_zero, line.offset[2])

            axis_zero = (
                    axis_zero + line.data.value[:, axes[self._sepration_axis]].max() + separation
            )

        self._separation = value
```

kushalkolar:
Useful to be able to set the separation. Also the default is now 0.0 since 10 is often way too large when y-values are small. Will probably have to update gallery examples, will do later. 

closes #898

### DIFF
--- a/fastplotlib/graphics/line_collection.py
+++ b/fastplotlib/graphics/line_collection.py
@@ -639,16 +639,30 @@ class LineStack(LineCollection):
             **kwargs,
         )
 
+        self._sepration_axis = separation_axis
+        self._separation = separation
+
+        self.separation = separation
+
+    @property
+    def separation(self) -> float:
+        """distance between each line in the stack, in world space"""
+        return self._separation
+
+    @separation.setter
+    def separation(self, value: float):
+        separation = float(value)
+
         axis_zero = 0
         for i, line in enumerate(self.graphics):
-            if separation_axis == "x":
+            if self._sepration_axis == "x":
                 line.offset = (axis_zero, *line.offset[1:])
 
-            elif separation_axis == "y":
+            elif self._sepration_axis == "y":
                 line.offset = (line.offset[0], axis_zero, line.offset[2])
 
             axis_zero = (
-                axis_zero + line.data.value[:, axes[separation_axis]].max() + separation
+                    axis_zero + line.data.value[:, axes[self._sepration_axis]].max() + separation
             )
 
-        self.separation = separation
+        self._separation = value
--- a/fastplotlib/graphics/line_collection.py
+++ b/fastplotlib/graphics/line_collection.py
@@ -561,7 +561,7 @@ class LineStack(LineCollection):
         metadata: Any = None,
         metadatas: Sequence[Any] | np.ndarray = None,
         isolated_buffer: bool = True,
-        separation: float = 10.0,
+        separation: float = 0.0,
         separation_axis: str = "y",
         kwargs_lines: list[dict] = None,
         **kwargs,
@@ -610,7 +610,7 @@ class LineStack(LineCollection):
             metadata for each individual line associated with this collection, this is for the user to manage.
             ``len(metadata)`` must be same as ``len(data)``
 
-        separation: float, default 10
+        separation: float, default 0.0
             space in between each line graphic in the stack
 
         separation_axis: str, default "y"


kushalkolar:
@clewis7 gtg, I don't need to change the examples they explicitly set the separation already :D

fastplotlib/graphics/line_collection.py
```diff
@@ -561,7 +561,7 @@ def __init__(
         metadata: Any = None,
         metadatas: Sequence[Any] | np.ndarray = None,
         isolated_buffer: bool = True,
-        separation: float = 10.0,
+        separation: float = 0.0,
```
clewis7, line 564: Default separation needs to be bigger than 0, otherwise it will just plot all of the user's lines on top of each other, which might be confusing/unexpected

If 10 is too big, just do 1.0 or something small

clewis7 (changes requested):

fastplotlib/graphics/line_collection.py
kushalkolar: 0 stacks them with no gap between the ymax of the previous and ymin of the next. 1 can be too much when the yrange of the data is very small.

kushalkolar: So I was thinking 0 would be a good default, just plain stacking

clewis7: So all the lines are on top of each other? Why would someone not just use a `LineCollection` then? 

Maybe the default should be a value based on the data. Similar to vmin/vmax, do a quick calc of the y range or x range depending  on the separation axis and use that as the default.

kushalkolar: 0 means it is placed above the y-max of the previous line, so stacking with no extra spacing

kushalkolar: see this line https://github.com/fastplotlib/fastplotlib/pull/990/changes#diff-03fd64063e483d20627556522309e2b0db62af2bfac92d975125122470f14809R665

kushalkolar:
I'm now thinking if we should just have LineStack functionality into LineCollection. Sometimes you see a stack and then you want to see them on top of each other for a few traces (easy peasy since we now have the `VisibilitySelector`). Maybe `LineCollection.separation = None` means no stacking, `LineCollection.separation = 0.0` means stacking directly on top based on previous line's y-max. We keep `LineStack` as just a simple subclass for API stability since I think a lot of people are using that already.

kushalkolar:
superseeded in #1072 1072

PR #993, opened 2026-02-17
setting initial scale was missing from `Graphic._set_world_object`

kushalkolar:
I forgot this when I added the scale feature in #947


kushalkolar:
@clewis7 gtg, failures are just nb screenshots which is due to rendercanvas changes that I will get to later

PR #995, opened 2026-02-18
bugfix setting limits in `LinearSelector`

kushalkolar:
I'm finding so many bugs as I push fpl to the limits with `NDWidget` :laughing: . @clewis7 gtg!


kushalkolar:
eh there are more problems, will just fix in NDWidget branch and make issues to keep track of what bugfixes were fixed in the NDWidget PR too XD

PR #999, opened 2026-02-20
Adding top gui to fpl

clewis7:
image

I thought this would follow the patterns of the other guis (just change the y pos of where pygfx renders to), but the render space does not actually move down unless I manually set the rect 

Maybe that is not something that needs to be done for bottom and right but does for a left or top gui 

Have enough for the thing I need to do for clepio and can circle back to this 

@kushalkolar thoughts welcome since you did the frame/rect manager stuff

### DIFF
--- a/fastplotlib/layouts/_frame.py
+++ b/fastplotlib/layouts/_frame.py
@@ -304,7 +304,7 @@ class Frame:
         w -= 2  # subtract 2, so we get a 1 pixel edge on both sides
 
         # add 4 pixels above and below title for better spacing
-        y = + y + 4 + self._title_graphic.font_size + 4
+        y = y + 4 + self._title_graphic.font_size + 4
 
         # spacing on the bottom if imgui toolbar is visible
         if self.toolbar_visible:
--- a/fastplotlib/layouts/_imgui_figure.py
+++ b/fastplotlib/layouts/_imgui_figure.py
@@ -195,9 +195,6 @@ class ImguiFigure(Figure):
 
         self._fpl_reset_layout()
 
-    def _fpl_reset_layout(self, *ev):
-        """set the viewport rects for all subplots, *ev argument is not used, exists because of renderer resize event"""
-        self.layout.canvas_resized(self.get_pygfx_render_area())
 
     def get_pygfx_render_area(self, *args) -> tuple[int, int, int, int]:
         """
--- a/fastplotlib/layouts/_rect.py
+++ b/fastplotlib/layouts/_rect.py
@@ -68,21 +68,21 @@ class RectManager:
         # check that widths, heights are valid
 
         # account for potential x and y offset
-        local = rect.copy()
-        local[0] -= x_offset
-        local[1] -= y_offset
+        rect_offset = rect.copy()
+        rect_offset[0] -= x_offset
+        rect_offset[1] -= y_offset
 
-        if local[0] + local[2] > cw:
+        if rect_offset[0] + rect_offset[2] > cw:
             raise ValueError(
-                f"invalid rect: {rect}\n x + width > canvas width: {local[0]} + {local[2]} > {cw}"
+                f"invalid rect: {rect}\n x + width > canvas width: {rect_offset[0]} + {rect_offset[2]} > {cw}"
             )
-        if local[1] + local[3] > ch:
+        if rect_offset[1] + rect_offset[3] > ch:
             raise ValueError(
-                f"invalid rect: {rect}\n y + height > canvas height: {local[1]} + {local[3]} >{ch}"
+                f"invalid rect: {rect}\n y + height > canvas height: {rect_offset[1]} + {rect_offset[3]} >{ch}"
             )
 
-        self._rect_frac[:] = local / mult
-        self._rect_screen_space[:] = rect
+        self._rect_frac[:] = rect_offset / mult
+        self._rect_screen_space[:] = rect_offset
 
     @property
     def x(self) -> np.float64:


kushalkolar:
I'm brain fried I'll review tomorrow or this weekend, but if right and bottom imgui windows still work then it's probably fine, let's worry about left guis when there's a compelling use case?

examples/guis/imgui_top.py
```diff
@@ -0,0 +1,75 @@
...
+    size=30,  # width or height of the GUI window within the figure
+    location="top",  # the edge to place this window at
+    title=" ",  # window title
+)
+
+# add it to the figure
+figure.add_gui(gui)
+
+width, height = figure.canvas.get_logical_size()
+figure[0,0]._frame.rect = np.array([0, gui.size, 640, height - gui.size])
```
kushalkolar: This can probably be done automatically?

clewis7: yes! just need to think about the best place to do it

clewis7:
Proper logic for setting the pyfgx render space and allowing space for guis is in `_imgui_figure.py` and `ui/_base.py` 

The issue now is getting the rect manager and frame to obey the change in the y-coord 

image

The height is changing appropriately but the y-coord to start rendering from is not being obeyed

clewis7:
Okay, the reason that it is not shifting down is because updates to the layout use the fractional layout to resize, and the rect manager does not take into account x or y offset when doing this

clewis7:
@kushalkolar gtg if you can give a review!

fastplotlib/layouts/_imgui_figure.py
```diff
@@ -195,6 +195,10 @@ def add_gui(self, gui: EdgeWindow):
 
         self._fpl_reset_layout()
 
+    def _fpl_reset_layout(self, *ev):
+        """set the viewport rects for all subplots, *ev argument is not used, exists because of renderer resize event"""
+        self.layout.canvas_resized(self.get_pygfx_render_area())
+
```
kushalkolar: Confused, this exact method exists in the parent class 🤔

fastplotlib/layouts/_frame.py
```diff
@@ -304,7 +304,7 @@ def get_render_rect(self) -> tuple[float, float, float, float]:
         w -= 2  # subtract 2, so we get a 1 pixel edge on both sides
 
         # add 4 pixels above and below title for better spacing
-        y = y + 4 + self._title_graphic.font_size + 4
+        y = + y + 4 + self._title_graphic.font_size + 4
```
kushalkolar: what does the `+` change do?

fastplotlib/layouts/_rect.py
```diff
@@ -39,8 +39,8 @@ def _set(self, rect):
 
     def _set_from_fract(self, rect):
         """set rect from fractional representation"""
-        _, _, cw, ch = self._canvas_rect
-        mult = np.array([cw, ch, cw, ch])
+        rect = np.asarray(rect, dtype=float).copy()
+        x_offset, y_offset, cw, ch = self._canvas_rect
```
kushalkolar, line 43: The `_canvas_rect` here is actually the pygfx render area part of the canvas right? I should've made that easier to understand, we could change the name and add more comments?

fastplotlib/layouts/_rect.py
```diff
@@ -54,24 +54,34 @@ def _set_from_fract(self, rect):
...
         """set rect from screen space representation"""
-        _, _, cw, ch = self._canvas_rect
+        x_offset, y_offset, cw, ch = self._canvas_rect
         mult = np.array([cw, ch, cw, ch])
         # for screen coords allow (x, y) = 1 or 0, but w, h must be > 1
         # check that widths, heights are valid
-        if rect[0] + rect[2] > cw:
+
+        # account for potential x and y offset
+        local = rect.copy()
```
kushalkolar: what is the name `local` for

fastplotlib/layouts/_rect.py
```diff
@@ -54,24 +54,34 @@ def _set_from_fract(self, rect):
...
-        if rect[1] + rect[3] > ch:
+        if local[1] + local[3] > ch:
             raise ValueError(
-                f"invalid rect: {rect}\n y + height > canvas height: {rect[1]} + {rect[3]} >{ch}"
+                f"invalid rect: {rect}\n y + height > canvas height: {local[1]} + {local[3]} >{ch}"
             )
 
-        self._rect_frac[:] = rect / mult
+        self._rect_frac[:] = local / mult
         self._rect_screen_space[:] = rect
```
kushalkolar: the fract is set with local, but the screen space is set directly?

kushalkolar:
Tested that the window layouts subplots move and resize properly with these changes? We can make automated tests for that later.

fastplotlib/layouts/_imgui_figure.py
clewis7: oh yes, I need to remove that

fastplotlib/layouts/_frame.py
clewis7: Nothing, when things weren't working, I was trying moving the title around to make sure I wasn't going crazy

fastplotlib/layouts/_rect.py
clewis7: local is just  a var I am using to include the offset, can choose something more informative

fastplotlib/layouts/_rect.py
clewis7: good catch, because we set directly for the opposite mode (fract directly when setting with screen value or vice versa), I missed that one

clewis7:
> Tested that the window layouts subplots move and resize properly with these changes? We can make automated tests for that later.

I tested locally that the window layout examples are still working (resizing and moving subplots around)

clewis7:
image

Since I didn't post what it looked like after the changes

kushalkolar:
The example png file got committed directly to git instead of git lfs, no idea why 🤦‍♂️ . I wonder if there's CI action we can use to catch this in the future. Since this was a recent commit to main I think we can undo it without too much trouble.

clewis7:
:facepalm: 

Wow that is totally my fault 

I can open a PR to fix today. I need to go in and fix the contrib docs so that they are a little clearer on the screenshot uploads.

kushalkolar:
It shouldn't be possible, seems git just ignored the .gitattributes for some reason

PR #1007, opened 2026-02-28
add AI statements

clewis7:
@kushalkolar feel free to suggest or edit!


kushalkolar:
I think another thing we should add is something like: before making a PR that introduces significant changes, or new features that are not in our roadmap, please open an issue describing the changes so we can assess whether the contribution would be accepted or not, and so we can provide guidance on how their proposed implementation can be tailored so it conforms with the rest of the codebase. 

We could also make a PR template that has this info and other key things.

PR #1020, opened 2026-03-20
add partial camera linking example

kushalkolar:
adds the example from here to the gallery https://github.com/fastplotlib/fastplotlib/pull/938


kushalkolar:
@clewis7 gtg

clewis7 (approved):

PR #1023, opened 2026-03-31
async NDProcessor

kushalkolar:
just wanted to work on this in a different branch, will merge into `ndwidget` branch once this works.


kushalkolar:
async isn't the bottleneck in masknmf, see: https://github.com/apasarkar/masknmf-toolbox/pull/207, can think about this later if it comes up again.

PR #1025, opened 2026-04-08
Includes code for doing batched transforms in both directions

apasarkar:
This PR allows you to do batched coordinate transforms. You have a graphic, the graphic has data points stored in model space. You want to transform all of those model points to world space (or go from world space to model space). Easy to batch/vectorize this via pylinalg but the fpl functions didn't allow for this. 

You can do this now: 

```
my_data = graphic.data[:] # Let's say this is shape (100, 3)
world_coords = graphic.map_model_to_world(my_data) 
```

graphic.map_world_to_model also supports this now

### DIFF
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -549,7 +549,7 @@ class Graphic:
         Parameters
         ----------
         position: tuple of (x, y, z) or np.ndarray of shape (num_points, 3)
-            The xyz positions we wish to map to model space
+            The xyz positions we wish to map to world space
 
         Returns
         -------
--- a/fastplotlib/graphics/_base.py
+++ b/fastplotlib/graphics/_base.py
@@ -529,35 +529,28 @@ class Graphic:
         """
 
         if isinstance(position, tuple):
-            if len(position) == 2:
-                # use z of the graphic
-                position = [*position, self.offset[-1]]
+            position = np.array(position)[None, :] #Shape (1, 3)
 
-            if len(position) != 3:
+        if isinstance(position, np.ndarray):
+            if not 0 < position.ndim < 3:
                 raise ValueError(
-                    f"position must be tuple or array indicating (x, y, z) position in *model space*"
+                    f"position can either be shape (num_points, 3) or (3,)"
                 )
-        elif isinstance(position, np.ndarray):
-            if position.ndim > 2:
-                raise ValueError(
-                    f"position can either be shape (num_points, [2, 3]) or ([2, 3],)"
-                )
-            if position.ndim == 1:
+
+            elif position.ndim == 1:
                 position = position[None, :]
 
-            if position.shape[-1] == 2:
-                z_data = np.zeros((position.shape[0], 1))
-                z_data[:] = self.offset[-1]
-                position = np.concatenate(
-                    [position, z_data], axis=1
-                )  # Shape (num_points, 3)
+            if position.shape[-1] != 3:
+                raise ValueError(f"position must have shape (num_points, 3) or (3,), provided shape of {position.shape}")
+
+
             elif position.shape[-1] != 3:
                 raise ValueError(
-                    "position must be a tuple or array indicating (x, y, z). The last dimension should be either 2 or 3"
+                    "position must be a tuple indicating (x, y, z) or an array of shape (num_points, 3)"
                 )
 
         else:
-            raise ValueError("positions must be either a tuple or np.ndarray")
+            raise ValueError("position must be either a tuple or np.ndarray")
 
         return position
 
@@ -569,8 +562,8 @@ class Graphic:
 
         Parameters
         ----------
-        position: (float, float, float) or (float, float) or np.ndarray of shape (num_points, 3), (num_points, 2), (3,), or (2,)
-            The xyz or xy positions we wish to map to model space. If z is not provided then 0 is used
+        position: tuple of (x, y, z) or np.ndarray of shape (num_points, 3)
+            The xyz positions we wish to map to model space
 
         Returns
         -------
@@ -590,8 +583,8 @@ class Graphic:
 
         Parameters
         ----------
-        position: (float, float, float) or (float, float) or np.ndarray of shape (num_points, 3), (num_points, 2), (3,), or (2,)
-            The xyz or xy positions we wish to map to model space. If z is not provided then 0 is used
+        position: tuple of (x, y, z) or np.ndarray of shape (num_points, 3)
+            The xyz positions we wish to map to model space
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,44 @@ def my_handler(event):
...
 
+    def _parse_positions(self, position: tuple | np.ndarray) -> np.ndarray:
+        """
+        Converts position data (in the form of tuple or np.ndarray) into a (num_points, 3)-shaped np.ndarray for processing
+        """
+
+        if isinstance(position, tuple):
+            if len(position) == 2:
+                # use z of the graphic
+                position = [*position, self.offset[-1]]
```
kushalkolar: if this is only used for model <-> world, we should enforce that z should always be provided. 

if they want to do model -> world then using the world z (which is the offset) will give the wrong result

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,44 @@ def my_handler(event):
...
+
+            if len(position) != 3:
+                raise ValueError(
+                    f"position must be tuple or array indicating (x, y, z) position in *model space*"
+                )
+        elif isinstance(position, np.ndarray):
+            if position.ndim > 2:
+                raise ValueError(
+                    f"position can either be shape (num_points, [2, 3]) or ([2, 3],)"
+                )
```
kushalkolar: "position must be of shape [3] or [n, 3]"

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,44 @@ def my_handler(event):
...
+            if position.ndim == 1:
+                position = position[None, :]
+
+            if position.shape[-1] == 2:
+                z_data = np.zeros((position.shape[0], 1))
+                z_data[:] = self.offset[-1]
+                position = np.concatenate(
+                    [position, z_data], axis=1
+                )  # Shape (num_points, 3)
```
kushalkolar: we shouldn't guess a z, user should provide it always

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,44 @@ def my_handler(event):
                 feature = getattr(self, f"_{t}")
                 feature.remove_event_handler(wrapper)
 
+    def _parse_positions(self, position: tuple | np.ndarray) -> np.ndarray:
+        """
+        Converts position data (in the form of tuple or np.ndarray) into a (num_points, 3)-shaped np.ndarray for processing
+        """
+
+        if isinstance(position, tuple):
```
kushalkolar: just convert it to array at the beginning, `np.asarray(position)` and use one code block.

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,44 @@ def my_handler(event):
...
+
+            if position.shape[-1] == 2:
+                z_data = np.zeros((position.shape[0], 1))
+                z_data[:] = self.offset[-1]
+                position = np.concatenate(
+                    [position, z_data], axis=1
+                )  # Shape (num_points, 3)
+            elif position.shape[-1] != 3:
+                raise ValueError(
+                    "position must be a tuple or array indicating (x, y, z). The last dimension should be either 2 or 3"
```
kushalkolar: [x, y, z] position or `[n, 3]`

kushalkolar (changes requested):

fastplotlib/graphics/_base.py
apasarkar: yep was thinking this

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,37 @@ def my_handler(event):
                 feature = getattr(self, f"_{t}")
                 feature.remove_event_handler(wrapper)
 
+    def _parse_positions(self, position: tuple | np.ndarray) -> np.ndarray:
+        """
+        Converts position data (in the form of tuple or np.ndarray) into a (num_points, 3)-shaped np.ndarray for processing
+        """
+
+        if isinstance(position, tuple):
```
kushalkolar: No if clause, always call np.asarray, it's free if the obj is already an array.

kushalkolar (changes requested):

fastplotlib/graphics/_base.py
apasarkar: @kushalkolar done

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,28 @@ def my_handler(event):
                 feature = getattr(self, f"_{t}")
                 feature.remove_event_handler(wrapper)
 
+    def _parse_positions(self, position: tuple | np.ndarray) -> np.ndarray:
+        """
+        Converts position data (in the form of tuple or np.ndarray) into a (num_points, 3)-shaped np.ndarray for processing
+        """
+        if not isinstance(position, (np.ndarray, tuple)):
```
kushalkolar: Just use np.asarray with no type checks, if numpy can't convert it to an array it'll raise, so it'll also handle a list etc.

kushalkolar (changes requested):

fastplotlib/graphics/_base.py
apasarkar: done @kushalkolar

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,23 @@ def my_handler(event):
                 feature = getattr(self, f"_{t}")
                 feature.remove_event_handler(wrapper)
 
+    def _parse_positions(self, position: tuple | np.ndarray) -> np.ndarray:
+        """
+        Converts position data (in the form of tuple or np.ndarray) into a (num_points, 3)-shaped np.ndarray for processing
+        """
+        position = np.asarray(position)
+
+        if not 0 < position.ndim < 3:
```
kushalkolar: This is clearer: `if position.ndim not in (1, 2):`

fastplotlib/graphics/_base.py
```diff
@@ -528,6 +523,23 @@ def my_handler(event):
...
+            raise ValueError(f"position must be of shape (num_points, 3) or (3,)")
+
+        elif position.ndim == 1:
```
kushalkolar: this can just be a `if`, not `elif`

kushalkolar (changes requested):

fastplotlib/graphics/_base.py
apasarkar: all done @kushalkolar

fastplotlib/graphics/_base.py
```diff
@@ -536,27 +548,18 @@ def map_model_to_world(
 
         Parameters
         ----------
-        position: (float, float, float) or (float, float)
-            (x, y, z) or (x, y) position. If z is not provided then the graphic's offset z is used.
+        position: tuple of (x, y, z) or np.ndarray of shape (num_points, 3)
+            The xyz positions we wish to map to model space
```
kushalkolar:
```suggestion
            The xyz positions we wish to map to world space
```

PR #1026, opened 2026-04-08
async NDProcessor

kushalkolar:
First time I'm doing async stuff, this is an adventure.


kushalkolar:
I just realized a few lines of code and GPU arrays can be fully async too, with the transfers to CPU only at the end for all of them :D 

cupy, torch and jax all implement `__cuda_array_interface__`

Check if `__cuda_array_interface__` is implemented, and if so treat that similar to a future. To resolve it just convert to cupy and send to CPU!

PR #1033, opened 2026-04-13
implement yuv and other colorspaces and "bufferless" `TextureArray`

kushalkolar:
- [x] implement `colorspaces`. Existing `ImageGraphic` uses rgb colorspaces.
- [x] `ImageYUVGraphic` for yuv colorspaces `yuv420p` and `yuv444p`
- [x] `ImageBase` that is inherited by`ImageGraphic` and `ImageYUVGraphic`
- [x] implement "bufferless" support in `TextureArray` using the `cpu_buffer=False` kwarg. Behaves exactly the same from the perspective of the `ImageGraphic` API! So `image_graphic.data = new_data` can be done just like before, but with bufferless it's much faster. `ImageYUVGraphic` is always bufferless.
- [x] update image graphic tooltip to handle bufferless, caveat is that it's no longer possible to get exact grayscale values so it does an inverse transform using the LUT. Tooltip has a note that the value is approximate, should document in the docstring too.
- [x] selector tools will not be able to grab the underlying data, document it and raise in the corresponding selector tools.
- [x] `NDImage` can handle both `ImageGraphic` and `ImageYUVGraphic`
- [x] `NDImageProcessor` subclass called `VideoProcessor` which can handle the tuple of arrays for `(y, u, v)`
- [x] `ndw_subplot.add_video()` which just calls `add_nd_image()` with with `processor_type=VideoProcessor`
- [x] docstrings for new functionality
- [x] example for yuv colorspaces
- [x] allow providing y, u, v explicitly, so it can be set directly from `pyav.Frame.planes` without creating an array which creates a copy.
  ```
  vh.current_frame
  Out[11]: <av.VideoFrame, pts=0 yuv420p 1600x1200 at 0x7f135828ea40>
  vh.current_frame.planes
  Out[12]: 
  (<av.VideoPlane 1996800 bytes; buffer_ptr=0x7f12441525c0; at 0x7f135828a6b0>,
   <av.VideoPlane 499200 bytes; buffer_ptr=0x7f1244341340; at 0x7f135828abb0>,
   <av.VideoPlane 499200 bytes; buffer_ptr=0x7f12443bcf00; at 0x7f135828b010>)
  ```

This gives us easy ~80fps on Gily's gerbil stuff which we were previously getting ~15fps on :partying_face: 

https://github.com/user-attachments/assets/423f89c9-7f06-43cf-89e9-1fb4179b2853

### DIFF
--- a/examples/image/image_yuv.py
+++ b/examples/image/image_yuv.py
@@ -1,0 +1,45 @@
+"""
+YUV Image
+=========
+
+Example that shows how to use YUV images. Most videos are stored in this colorspace.
+Y stores luma at full resolution, UV stores chroma values.
+In yuv420p UV channels are stored at half the resolution of Y. In yuv444p, UV channels are stored
+at full resolution.
+
+YUV is also called YCbCr for digital images.
+
+For more info: https://en.wikipedia.org/wiki/Y%E2%80%B2UV
+"""
+
+# test_example = true
+# sphinx_gallery_pygfx_docs = 'screenshot'
+
+import fastplotlib as fpl
+import numpy as np
+from skimage.color import rgb2ycbcr
+import imageio.v3 as iio
+
+# convert an rgb image to ycbcr for example purposes
+img = iio.imread("imageio:astronaut.png")
+img_yuv = rgb2ycbcr(img).astype(np.uint8)
+
+y = img_yuv[..., 0]
+u = img_yuv[::2, ::2, 1]
+v = img_yuv[::2, ::2, 2]
+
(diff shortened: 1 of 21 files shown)


kushalkolar:
ok I'm going to benchmark the unpacked YUV with larger data and more vids simultaneously and see if we really need it. If not I will just keep the packed since it's easier to deal with.

kushalkolar:
OK I'm going to add an entirely new `ImageYUVGraphic` for yuv images that uses a single texture specific for yuv, this is becoming a giant mess.

PR #1036, opened 2026-04-16
Adds kwargs so user can customize vector field estimates

apasarkar:


kushalkolar:
For the ndgraphic, have an explicitly kwarg "graphic_kwargs" and store that. See how the graphic kwargs are handled for ndimage and positions. When the graphic gets re-created if the data is re-set, it uses the original kwargs. 
In the add_nd_vectors helper method you can have size as a kwarg and put that in the dict along with any other graphic kwargs when you parse.

apasarkar:
Sweet thanks, made the fixes to mirror what ndpositions does, lmkwyt

fastplotlib/widgets/nd_widget/_nd_vectors.py
```diff
@@ -273,7 +279,7 @@ def graphic(
         return self._graphic
 
     @start_coroutine
-    def _create_graphic(self):
+    def _create_graphic(self, **kwargs):
```
kushalkolar: create_graphic shouldn't take kwargs

fastplotlib/widgets/nd_widget/_nd_vectors.py
```diff
@@ -294,7 +300,9 @@ def _create_graphic(self):
 
         # create the new graphic
         self._graphic = self._subplot.add_vectors(
```
kushalkolar: `add_nd_vectors` so it's like the others

kushalkolar (changes requested):

fastplotlib/widgets/nd_widget/_nd_vectors.py
apasarkar: i've updated the code to mirror what's being done in the other NDGraphic classes. something like: 

self._graphic = graphic_class(params)
self._subplot.add_graphic(self._graphic)

kushalkolar: oh I was looking at the wrong thing nvm :facepalm:

kushalkolar (approved):

PR #1038, opened 2026-04-17
`SelectionVector`, `HighlightSelector`, `VisibilitySelector`, `SelectorCollection`

kushalkolar:
- selection vector

$$
\begin{aligned}
\text{selection vector of size n: } \mathbf{s} \in \mathbb{U}^n \\
\text{mapping from selection vector to selectable: } \Phi(\mathbf{s}) \to \mathbf{s'} \in \\mathbb{R}^{n'} \\
linear: \\mathbb{R}^1 \\
LinearRegion: \\mathbb{R}^2 \\
rect: \\mathbb{R}^4 \\
point: \\mathbb{R}^2 \\
polygon, contour: \\mathbb{R}^{c\times2} \mbox{ , c: number of coordinates} \\
\end{aligned}
$$

Adds two new sets of selectors that directly modify the rendered fragments via shader injection. This was chosen so that the final pick_info isn't affected for highlighting objects, and also so that images (such as heatmaps) can be sorted in-place on the GPU without re-sending the buffers. 

Disclosure: I used claude to get a starting point with the shader injection stuff I'm not too familiar with, also used it for some boilerplate. Required extensive review and re-write though, since as per usual LLMs mostly produce hot garbage (even claude opus 4.7 latest and greatest blah blah version)


kushalkolar:
Can make the "AppearSelector" operate by directly setting the data property, so then it would work for both graphic collections and ND positions graphics. Could specify which dim it acts upon so it can also work for non collection graphics. For ND graphics it should only operate on a spatial dim.

kushalkolar:
For the image highlighter, allow it to have a property like "selection_options" which takes a dict of valid selections. These can be colored with a given color or lut to indicte the "selection options", such as light gray with low alpha. And then the user can set the selection as an int to choose from the already set selection options.

Not sure if we need a similar thing for lines and scatters.

kushalkolar:
Need to see if the highlight for images works across ImageTiles too.

kushalkolar:
For images, visibility selector and highlight seems to stop changing ~10 are selected, need to see what's up. Similar for line collection, but seems to break a bit later, maybe after ~20. will check tomorrow.

Also for ImageVisibilitySelector, need to show blank when selection is empty

kushalkolar:
I'm going to postpone reviewing the `PositionsHighlightSelector` and `CollectionHighlightSelector` until later, will merge into ndwidget and review there.

kushalkolar:
@almarklein I've been implementing a highlight selection and "visibility" selection for images by shader injection. I'm wondering if this should stay in fastplotlib or if it makes sense to add these properties upstream to pygfx.

The "highlight" alpha-blends a "highlight color" before writing the final output color. The advantage with this is that a separate `Image` object doesn't have to be rendered on top of the target image with alpha blending. So we can get the pick_info of the original image that we're interested in, rather than the mask image on top. We also get the actual data values instead of the highlight-blended data values in the pick-info.

The "visibility" option sets the rows or columns that are actually sampled, this only makes sense if the interpolation mode is nearest. It's very useful for heatmaps and for re-ordering large heatmaps in-place without sending new data to the GPU in the desired order.

Video illustrates both:

https://github.com/user-attachments/assets/d66d1deb-77f8-40cc-ae7c-f5a4a6379dd4

@hmaarrfk wondering if you've done similar stuff too

hmaarrfk:
I do the replacement trick a lot. 

I’m close to being over it and just forking some shaders. It is hard to maintain. But if you are on top of it, it works great. 

This highlighting is cool. 

Thanks for sharing

kushalkolar:
> I’m close to being over it and just forking some shaders. It is hard to maintain. But if you are on top of it, it works great. 

If enough people want some features we can add them to pygfx? 😁

hmaarrfk:
> If enough people want some features we can add them to pygfx? 😁

I think so, but generally speaking, there is always tradeoff between complexity, ease of use, functionality, and performance.

We should create a secondary repo to build these features in a standalone fashion and not bog down the development of PyGfx

kushalkolar:
> > If enough people want some features we can add them to pygfx? 😁
> 
> I think so, but generally speaking, there is always tradeoff between complexity, ease of use, functionality, and performance.
> 
> We should create a secondary repo to build these features in a standalone fashion and not bog down the development of PyGfx

True. I wonder if there's any standard way to customize shaders in a maintainable way, maybe specified hooks we could use via the existing templating? I'll think about this a bit and maybe post an issue there.

almarklein:
Cool effect that highlighting!

I think it's too specific and complex (as in needing additional lut etc) to be included in PyGfx.

But something like a repo for more specialized objects and shaders sounds like a good idea!

Also, we can think about the API for creating custom objects and see if we can make it easier to create custom objects, as it's a little involved now.

PR #1040, opened 2026-04-25
compute capabilities

kushalkolar:
Starting to prototype the combination of compute and viz, making required modifications to fastplotlib that are required for this. Going to start with creating the basic things we need for masknmf viz. Will put actual compute + viz examples and implementations in https://github.com/kushalkolar/wgpu-linalg

- [x] compute shader can target textures in a TextureArray so fastplotlib can display them

Thoughts:

I think a full basic numpy-like API with `.to("wgpu")` and `to_numpy()` calls would be useful. We need `TextureArray` and all other data-features in fastplotlib to behave identical regardless of where the data is. For `TextureArray` this means we need the `format` as well, so dtype etc.


kushalkolar:
I will do positional based graphic buffers later, merging for now.

PR #1050, opened 2026-05-09
use fully fledged async, `NDPositions` improvements

### LINKED ISSUE #1042: generator coroutine -> proper async in NDWidget
So I'm going to modify the current async to use fully-fledged python async. This will allow the render loop to not be blocked while waiting for data, it can also handle fast slider movements and better cancelling I think. This will also allow `window_funcs` and the `spatial_func` to also be async (provided they use numpy or something in them, which should cover most use cases). It takes ~8ms to calculate a mean average of a 33x1024x1024 array so this is definitely worth it.

Should just use threading for the `window_funcs` and `spatial_funcs async`, using a process adds impossible barriers/complication  ,ex: need to import all necessary modules in the other process etc.

kushalkolar:
This replaces the generator-as-a-couroutine with proper full async and uses rendercanvas's existing loop for scheduling. Throttling behavior is much more natural now and IO never blocks rendering even when waiting for Futures to resolve. The entire pipeline from getting the raw data --> `window_funcs` --> `spatial_func` is async and fully concurrent. 

closes #1042

@apasarkar this should allow concurrent running of `window_funcs` and `spatial_funcs` written using `torch` that stay on the device until the very end, ex: `torch.mean()` for rolling mean windows instead of `np.mean`.

- [x] need to benchmark and verify that the performance is just as good as before, especially with cuda arrays from masknmf.
- [x] `NDWSubplot.subplot` property @apasarkar 
- [x] `NDGraphic._set_indices_()` doesn't take any arguments, it retrieves the latest indices _directly_ from the ref_index instance. This is because since `_set_indices_()` is a _scheduled request_ AND the `indices` dict is produced at the time it's scheduled in `_render_indices()`, the `indices` argument will not reflect the latest indices at _execution time_. So `NDGraphic._set_indices_()` just always grabs the latest indices when it executes.
- [x] better `x_range_mode` logic in NDPostions, get rid of jittery and jigglyness when `x_range_mode="auto"`, fix other bugs. `_update_from_view_range()` uses `with block_indices_ctx()` to any async scheduling of `_set_indices_()` when `x_range_mode="auto"` but we allow async scheduling to update _all other_ `NDGraphics`. We then `run_sync(self._set_indices_())` immediately so that the `NDPositions` graphic immediately reflects the latest polled camera state. This gets rid of the jitter and jiggly effect! :partying_face: 
  - [x] no `x_range` fighting when setting indices via `_set_indices_`
  - [x] no `x_range` fighting with `_update_from_view_range()` - trying to figure this out
  - [x] fix bug when width was 0 due to figure auto_scale when using `VisibilitySelector` and no lines are visible (i.e. initial selection is empty, blank subplot).
  - [x] only one `NDPositions` `LinearSelector` per-subplot, so multiple timeseries can be overlaid in one subplot and they use the same linear selector.
- [x] better `PlotArea.x_range` and `y_range` setters and getters.
- [x] keep UI throttling for imgui slider, and also use `Future` cancellation in `ReferenceIndex` render request logic.
- [x] No throttling for other forms of viewing data, ex: play button, stepping through frames, using the LinearSelector, we want to see every single datapoint/frame of data when doing fine-grained inspection like this.
- [x] make sure new async works when slider is moved around very fast with multi-video decoding with `fava` - almost there
- [x] similarly, make sure new async works when viewing `masknmf.DemixingResults` and moving the slider very fast.
- [x] tested examples: masknmf single session, ephys, kcenia behavior, gerbils. WIP masknmf multi session, things wonky with timeseries.
- [x] cleanup comments and docstrings

@BalzaniEdoardo


kushalkolar:
No more jitter and jiggly effect in `x_range_mode="auto"`! Rock-solid.

https://github.com/user-attachments/assets/967cb12d-38a9-452e-97c4-7c94e259f0f2

IO such as video decoding in an independent process + new async lets us get 120+fps easily. We still get ~100fps+ rendering when we set the playback fps to 100. With the previous generator-as-a-coroutine simpler async without rendercanvas scheduling we got ~60fps, so we can get almost double the fps with this proper async that uses  rendercanvas scheduling! :partying_face: . The decoder does get stuck for a few seconds if a huge backlog piles up I think, but from inspecting the stack this lag comes from the decoder. Anyways the frame corresponding to the latest indices do appear within a few seconds and I think it's unlikely a user will be scrolling such that they request 100 different random frames/second for long periods of time, or playback at > 100fps.

https://github.com/user-attachments/assets/93549996-c2b6-49dc-b1af-b2212757f990

With masknmf demixing results with many subplots etc. ([this](https://github.com/fastplotlib/fastplotlib-neuro-examples/blob/main/examples/dmr.py)) I get ~28fps with the new propery async, 17 fps before.

PR #1052, opened 2026-05-09
local pyodide preview

Vipitis:
Hey,

as I getting wgpu-py and pygfx to run in the browser I also tried some of the fpl examples to find some more problems. 

thought I will just share the script which is really janky and adapted from rendercanvas, wgpu and pygfx... but mostly works. It includes a bunch of hacks to make examples work. It also pulls in the dependencies from the PR branch doc previews... so this might break when those are updated or merged etc.

I haven't looked into any fpl specific problems with pyodide, but there are a few I saw already (like the tool bar is missing?, windows overlapping?) - just clicking around to see any issues that aren't covered in the wgpu and pygfx examples and already found something that is now fixed.

to try this yourself, simply clone this branch (or copy the script really...)

```sh
> cd examples
> python serve_browser_examples.py
```

I only tested this on Win10 with Chrome but you should get this list of all the example files and most will run. You only run the script to patch in the local fpl wheel and serve the http. All execution is happening in the browser using pyodide. (so you could theoretically copy the html source and host it statically anywhere)

https://github.com/user-attachments/assets/281a30de-265a-4aa4-bc90-729a50e55bd5

load times and performance is really bad at the moment.
I haven't looked into how the docs work for fpl, there is a really janky iframe injection that puts embeds interactive examples in the pygfx docs for example which I might be able to port that to here to make it even more accessible. 

let me know if you have any questions or suggestions and I will see how much time I can find - just really excited for what the future will enable with pyodide support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,8 @@ dmypy.json
 # diffs from visual regression tests
 examples/desktop/diffs/*.png
 docs/source/_gallery/
+
+
+# not sure why this spammed my diff
+docs/source/api/
+docs/source/_static/*.whl
--- a/docs/source/_static/_pyodide_iframe.html
+++ b/docs/source/_static/_pyodide_iframe.html
@@ -1,0 +1,63 @@
+<!doctype html>
+<html>
+<!-- adapted from pygfx, adapted from wgpu-py, adapted from rendercanvas -->
+<head>
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <title>Fastplotlib examples in the browser using Pyodide</title>
+    <script src="https://cdn.jsdelivr.net/pyodide/v0.29.4/full/pyodide.js"></script>
+</head>
+
+<body>
+    <dialog id="loading" style='outline: none; border: none; background: transparent;'>
+        <h1>click the button to load pyodide</h1>
+        <button id="load_button" type="button" onclick="main()">load in browser!</button>
+    </dialog>
+    <canvas id='canvas' style='width:calc(100%); height:560px; background-color: #ddd;'></canvas>
+    <div id="output" style="white-space: per-line; overflow-y: auto; height:90px; background:#eee; border:1px solid #ccc;">
+        <p>Output:</p>
+    </div>
+    <script type="text/javascript">
+        let loading = document.getElementById('loading');
+        loading.showModal();
+        async function main() {
+            let button = document.getElementById("load_button");
+            button.disabled = true;
+            button.innerText = "loading...";
+            try {
+                let example_name = document.location.hash.slice(1);
+                pythonCode = await (await fetch(window.parent.location.pathname.replace(".html", ".py"))).text();
+                let pyodide = await loadPyodide();
+                pyodide.setStdout({
+                    batched: (s) => {
+                        el = document.getElementById("output");
+                        el.innerHTML += "<br>" + s.replace(/</g, "&lt;").replace(/>/g, "&gt;");
+                        el.scrollTop = el.scrollHeight;  // scroll to bottom
+                        console.log(s); // so we also have it formatted
+                    }
+                });
+                await pyodide.loadPackage("micropip");
+                const micropip = pyodide.pyimport("micropip");
+                await micropip.install([
+                    'https://wgpu-py--753.org.readthedocs.build/en/753/_static/wgpu-0.31.0-py3-none-any.whl',
+                    'imgui-bundle',
+                    "https://pygfx--1273.org.readthedocs.build/1273/_static/uharfbuzz-0.54.1-cp310-abi3-pyodide_2025_0_wasm32.whl",
+                    'hsluv', 'pylinalg', 'jinja2', 'httpx', 'trimesh', 'gltflib', 'imageio',
+                    "https://pygfx--1273.org.readthedocs.build/1273/_static/pygfx-0.16.0-py3-none-any.whl"
+                ]);
+                await micropip.install("fastplotlib") // this one should be replaced by the wheel during docs build.
+                await pyodide.loadPackagesFromImports(pythonCode); // additional non pypi packages from the examples... like imageio (not patched...)
+                pyodide.setDebug(false);
+                let ret = await pyodide.runPythonAsync(pythonCode);
+                // maybe we can trigger a resize event from the js side to get the canvas to fit the iframe when it's in the docs?
+                console.log("Example finished:", ret);
+                loading.close();
+            } catch (err) {
+                // TODO: this could be formatted better as this overlaps and is unreadable...
+                loading.innerHTML = "Failed to load: " + err;
+                console.error(err); // so we have it here too
+            }
+        }
+    </script>
+</body>
+
+</html>
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -1,0 +1,5 @@
+section[id^="interactive-example"] iframe {
+    width: 100%;
+    height: 520px;
+    border: none;
+}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,6 +44,94 @@ extensions = [
     "sphinx_gallery.gen_gallery",
 ]
 
+# note this is largely copied from the pygfx PR branch: https://github.com/pygfx/pygfx/pull/1273
+# -- Build wheel so Pyodide examples can use exactly this version of fpl -----------------------------------------------------
+import subprocess
+import shutil
+
+short_version = ".".join(str(i) for i in fastplotlib.version_info[:3])
+wheel_name = f"fastplotlib-{short_version}-py3-none-any.whl"
+
+# Build the wheel
+subprocess.run([sys.executable, "-m", "build", "-nw"], cwd=ROOT_DIR)
+wheel_filename = os.path.join(ROOT_DIR, "dist", wheel_name)
+assert os.path.isfile(wheel_filename), f"{wheel_name} does not exist"
+
+# Copy into static
+# TODO: you can use --outdir on the build command directly. also use the html_static_path in this namespace
+print("Copy wheel to _static dir")
+shutil.copy(
+    wheel_filename,
+    os.path.join(ROOT_DIR, "docs", "source", "_static", wheel_name),
+)
+
+# -- Sphix Gallery -----------------------------------------------------
+
+## pyodide demos, adapted from wgpu, adapted from rendercanvas... might make sense to put this in the scaper?
+iframe_placeholder_rst = """
+.. only:: html
+
+    Interactive example
+    -------------------
+
+    Try this example in your browser using Pyodide. Might not work with all examples and all devices. Check the output and your browser's console for details.
+
+    .. raw:: html
+
+        <iframe src="./../pyodide.html#example.py"></iframe>
+"""
+python_files = {}
+
+# I have a feeling this import might be really sketchy to have on CI... but hey - this is a hack for a hack
+from examples.server_browser_examples import patch_imageio_for_pyodide
+
+
+def add_pyodide_to_examples(app):
+    if app.builder.name != "html":
+        return
+
+    gallery_dir = ROOT_DIR / "docs" / "source" / "_gallery"
+    example_files = gallery_dir.glob("**/*.py")
+
+    for py_file in example_files:
+        fname = py_file.name
+        with open(py_file, "rb") as f:
+            py = f.read().decode()
+            py = patch_imageio_for_pyodide(py)
+        if fname:
+            print("Adding Pyodide example to", fname)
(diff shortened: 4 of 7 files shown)


kushalkolar:
This is really cool! The toolbar requires imgui, was easy to get it working by just adding `"imgui_bundle"` to the deps. Was able to run stuff in chrome after enabling the wgpu experimental features. Firefox gave `RuntimeError: WebAssembly stack switching not supported in this JavaScript runtime`. Some examples that use uint8 don't work but this is a great start :partying_face:

Vipitis:
> The toolbar requires imgui, was easy to get it working by just adding `"imgui_bundle"` to the deps.

The way micropip resolves dependencies is a little odd to me, especially when you use .whl instead of a wheel on the index. Like the order matters for version conflicts etc. I saw imgui examples working, but that's due to a [pyodide helper](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackagesFromImports). I will add imgui_bundle to the list of deps there. Which might not be needed for every single example anyway. 
I think it's also an option to require imgui when `sys.platform == "emscripten"` in the `pyproject.toml`


> Firefox gave `RuntimeError: WebAssembly stack switching not supported in this JavaScript runtime`.

JSPI is used by pyodide to await an async function, it's used for the wgpu-py mappings. Looks like it's still worked on for [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1897981). I don't think moving pygfx to all async would drop that requirement... not sure.

> Some examples that use uint8 don't work

Can you share which examples caused errors for you? I can look into if they are are an issue in my wgpu changes that needs fixes.

kushalkolar:
> > Firefox gave `RuntimeError: WebAssembly stack switching not supported in this JavaScript runtime`.
> 
> JSPI is used by pyodide to await an async function, it's used for the wgpu-py mappings. Looks like it's still worked on for [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1897981). I don't think moving pygfx to all async would drop that requirement... not sure.

This is the full traceback I get on firefox, maybe I can try updating.

```python
Failed to load: PythonError: Traceback (most recent call last): File "/lib/python313.zip/_pyodide/_base.py", line 597, in eval_code_async await CodeRunner( ...<9 lines>... .run_async(globals, locals) File "/lib/python313.zip/_pyodide/_base.py", line 411, in run_async coroutine = eval(self.code, globals, locals) File "", line 11, in File "/lib/python3.13/site-packages/fastplotlib/__init__.py", line 6, in from .utils import loop # noqa ^^^^^^^^^^^^^^^^^^^^^^^ File "/lib/python3.13/site-packages/fastplotlib/utils/__init__.py", line 5, in from .functions import * File "/lib/python3.13/site-packages/fastplotlib/utils/functions.py", line 7, in from pygfx import Texture, Color File "/lib/python3.13/site-packages/pygfx/__init__.py", line 9, in from .objects import * File "/lib/python3.13/site-packages/pygfx/objects/__init__.py", line 79, in from ._ruler import Ruler File "/lib/python3.13/site-packages/pygfx/objects/_ruler.py", line 8, in from ._text import MultiText File "/lib/python3.13/site-packages/pygfx/objects/_text.py", line 43, in from ..utils import text as textmodule File "/lib/python3.13/site-packages/pygfx/utils/text/__init__.py", line 17, in from ._fontmanager import FontProps, FontManager, font_manager # noqa: F401 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontmanager.py", line 369, in font_manager = FontManager() File "/lib/python3.13/site-packages/pygfx/utils/text/_fontmanager.py", line 130, in __init__ self._load_fonts() ~~~~~~~~~~~~~~~~^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontmanager.py", line 165, in _load_fonts for ff in get_all_fonts(): ~~~~~~~~~~~~~^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontfinder.py", line 214, in get_all_fonts return get_builtin_fonts() | get_system_fonts() ~~~~~~~~~~~~~~~~~^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontfinder.py", line 220, in get_builtin_fonts return {FontFile(p) for p in file_paths} ~~~~~~~~^^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontfinder.py", line 71, in __init__ self._get_face() # load it early so the direct access to freetype attributes is skipped... feels janky. ~~~~~~~~~~~~~~^^ File "/lib/python3.13/site-packages/pygfx/utils/text/_fontfinder.py", line 97, in _get_face run_sync(face.load().then(lambda _: document.fonts.add(face), lambda err: print(err))) ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ RuntimeError: WebAssembly stack switching not supported in this JavaScript runtime 
```

> > Some examples that use uint8 don't work
> 
> Can you share which examples caused errors for you? I can look into if they are are an issue in my wgpu changes that needs fixes.

I got it in the `ndwidget` branch, but that might be because we recently allowed direct uint8 textures there. But in `main` the colormap picker doesn't show up when you right-click the colorbar in the `image_widget/image_widget.py` example. I wonder if it's an issue with the wgpu texture upload. It's basically this wgpu-imgui example: https://github.com/pygfx/wgpu-py/blob/main/examples/imgui_cmap_picker.py

Vipitis:
> the colormap picker doesn't show up when you right-click the colorbar in the `image_widget/image_widget.py` example. I wonder if it's an issue with the wgpu texture upload. It's basically this wgpu-imgui example: https://github.com/pygfx/wgpu-py/blob/main/examples/imgui_cmap_picker.py

ah, that was one example I never tried in wgpu-py and it also had this issue (it spews your browser console with a sorta helpful message). I pushed a commit that fixes the behaviour. But I think there might be an actual problem with how we transform data from python to js (some dimension of the shape might be lost), as the default (no args) for `.create_view()` seems to behave as expected natively but not in the browser.

I will look into that later, thanks!

kushalkolar:
I wonder what performance we can get with compute shaders here, like what's the overhead. When I get a chance I want to try running this stuff on it: https://github.com/kushalkolar/wgpu-sparse-matvec

Vipitis:
I am currently seeing about 20% performance compared to native when you look at framerates. But there is a lot to that. Moving data is probably the worst right now since it's doing 3 or more copies. And scheduling as issues because of the waiting and being driven by the browser framerate too. 
There should be very little overhead once you run compute on the GPU. Because it's the same GPU you would use native. But might be fun to measure it. 

I think one limitation with wasm is 2GB of memory. But not sure if that applies to resources on the GPU

Vipitis:
Not a webdev so the styling is all awful - but this works locally when I built the docs (plenty of issues with duplication for example). And the imgui cmap picker works!


https://github.com/user-attachments/assets/10180140-44b7-4db0-b372-d680fb23ce2b


I didn't want to mess the ci job for docs preview but I think it could work. Let me know if that is something I should try

(also found additional issues for myself, something about unmapped buffers in the `surface_ripple` example).

pthom:
Hi! Dear ImGui Bundle author here (@vipitis and I have talked on the bundle discord). 

First, congrats on this effort! Having worked on a similar approach, I know that this is a very complex endeavor. 
I'm not a fastplotlib dev, but since this PR can also use imgui_bundle under Pyodide, I thought I'd share some of the lessons I learned while working on creating a dedicated pyodide runner for imgui bundle's own runners. 

**Main suggestion: reuse the Pyodide runtime between examples** instead of calling `loadPyodide()` per page. Most of the per-example wait is not downloading the wheels (those get cached) but re-installing them into the fresh in-memory filesystem and re-initializing the interpreter. In the imgui_bundle playground (https://imgui-bundle.pages.dev/playground/), all demos share one runtime (and one canvas in my case): switching demos is near-instant because we only stop the previous app's render loop, tear it down, and exec the new code in the same runtime.

How imgui_bundle supports this, in case it's a useful reference:

- Importing imgui_bundle under Pyodide automatically patches `hello_imgui.run` / `immapp.run` into a non-blocking renderer driven by `requestAnimationFrame`, with a setup / render / teardown lifecycle: [pyodide_patch_runners.py](https://github.com/pthom/imgui_bundle/blob/main/bindings/imgui_bundle/pyodide_patch_runners.py)
- Before exec'ing a new demo, the playground calls `stop_active_renderer()` from that module, so the previous loop stops cleanly before its globals get rebound (the comments in that file describe the race this avoids)
- Playground source: https://github.com/pthom/imgui_bundle/tree/main/pyodide_projects/projects/playground (demo switching is in `js/pyodide_loader.js`)

Since fastplotlib drives its own loop via rendercanvas, this is not drop-in for you. But the same pattern may apply: keep one runtime alive, add a "stop the current loop and tear down" hook on the rendercanvas side, then run the next example against the same runtime and canvas.

**Second suggestion: embrace the async model end to end.** Two things mattered when I adapted imgui_bundle's runners to the browser:

- Never block or sleep in Python while a GUI is up. Each `requestAnimationFrame` callback renders exactly one frame and returns; anything that needs waiting (downloads, etc.) is a real `await` outside the frame callback.
- In the browser, `run()` becomes fire-and-forget: it sets up the app, schedules the first `requestAnimationFrame` callback, and returns immediately. The GUI then lives entirely in browser callbacks, so the rest of the script (and other asyncio tasks) keeps running between frames. I also provide an awaitable `run_async()` for code that needs to know when the GUI exits.

A nice side effect of this fully callback-driven approach: it never needs `run_sync`, and therefore does not depend on JSPI ("WebAssembly stack switching", the experimental feature that lets `run_sync` suspend synchronous code while a promise resolves; currently Chrome-only). That is why the playground also works on Firefox, and it matches your hunch that moving fully async would remove the JSPI requirement.

For what it's worth, when I tried this PR on Firefox, the first failure I hit was the font loading in pygfx: `run_sync(face.load().then(...))` in `pygfx/utils/text/_fontfinder.py`. It would be an interesting candidate to convert to a plain `await`.

One last trick: a synchronous download is still possible without JSPI, because the legacy synchronous `XMLHttpRequest` blocks inside the browser itself, not in the WASM stack. imgui_bundle uses this in [`immapp.download_url_bytes`](https://github.com/pthom/imgui_bundle/blob/main/bindings/imgui_bundle/immapp/__init__.py) (binary data needs a small base64 detour, see the code; the tab is frozen while downloading, so it's for one-shot loading, not mid-frame use). It could be a Firefox-compatible replacement for (note that it only works for downloads though).

Finally, there are more recent wheels for imgui bundle here : https://imgui-bundle.pages.dev/local_wheels/ 
(they are lighter: 5MB instead of 9MB). I hope to deploy those will be available in the standard pyodide distrib soon. 

Happy to discuss details here or on discord.

Vipitis:
hey @pthom thanks for you advice. I can see the benefit in all of and really think the playground kinda approach would work out. 

We can gladly chat more on Discord as I have some additional ideas. I'd say that this PR is way ahead of the current blockers which are still in wgpu-py. Pygfx and fastplotlib *should* just work once that is ready and released. 

How interactive docs/gallery might look and work is more of a design decision not really up to me.

.gitignore
```diff
@@ -139,3 +139,8 @@ dmypy.json
 examples/desktop/diffs/*.png
 docs/source/_gallery/
 docs/source/_imgui_images/
+
+
+# not sure why this spammed my diff
+docs/source/api/
```
kushalkolar, line 145: yea, I should make a sphinx extension that for `generate_api.py`, PR incoming...

Vipitis: no worries. The whole approach for this iframe patching is awful, but I also don't know sphinx enough to (or could find quick answers) for how to just append a few lines. There also must be better ways to include the python than to include all the files, as it's like on the page already.

Also see the comments about about making something like an interactive playground instead. (Can you mount a filesystems with all the examples perhaps?)

I just rebased this because the ndwidget merge was like a large rewrite. And something new broke with map_sync that I need to look into with wgpu-py and maybe hand implement.

It doesn't look like anything in fpl needs to be changed for pyodide, so hopefully we can get the wgpu-py (and pygfx) wheels ready for all the downstream projects to play with.

kushalkolar:
> The whole approach for this iframe patching is awful, but I also don't know sphinx enough to (or could find quick answers) for how to just append a few lines. There also must be better ways to include the python than to include all the files, as it's like on the page already.

If it works fine for now! 

> Also see the comments about about making something like an interactive playground instead. (Can you mount a filesystems with all the examples perhaps?)

I did see it but I'm not familiar enough with JS and web stuff to really grasp it. Mounting a filesystem sounds like a nice idea.

PR #1061, opened 2026-06-11
Selection vector improve

### LINKED ISSUE #1047: SelectionVector improvements
- [x] implement stuff that has placeholder (def remove() etc.)
- [x] `_inverse_map`, reentrance block is probably supposed to be in selection setter, append, and remove

apasarkar:
closes #1047. 

Concrete changes: 
- Includes logic to block re-entrance when selections are changed
- Includes more compact code for adding selectors to make things more declarative. Now the user just provides the selector + a global index --> local index mapping. This mapping is given by a 1D array (or a dictionary). The inverse mapping and event handler logic is dealt with internally within SelectionVector.
- Fills in missing functionality: clear selectables, remove, remove_selector


fastplotlib/graphics/selectors/_highlight_selector.py
```diff
@@ -681,7 +681,7 @@ def selection(self) -> tuple[int | None, ...] | dict[str, tuple]:
         return {k: tuple(v) for k, v in self._selection.items()}
 
     @selection.setter
-    def selection(self, value: Iterable[int] | dict[Literal["rows", "cols", "pixels"], list]) -> None:
+    def selection(self, value: Iterable[int] | dict[Literal["rows", "cols", "pixels"], list] | None) -> None:
```
kushalkolar: value can also be `Iterable[int | None]`

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
 from collections.abc import Callable
 from functools import partial
 from typing import Any, Sequence
-
+from numbers import Integral
+import numpy as np
 from ._protocols import SelectorProtocol, MultiSelectorProtocol
-
+from fastplotlib.graphics.features._base import GraphicFeatureEvent
```
kushalkolar: relative import

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
+    For each selector in the selection vector, the user specifies how the global indices (shared across selectors)
+    maps to the local indices.
+
+    The SelectionVector manages everything else, including the coordinated updating of indices whenever a selection changes
+    """
     def __init__(self, max_size: int = None):
         # selector -> (map, map_inv)
         self._selectors: dict[
-            SelectorProtocol | MultiSelectorProtocol, tuple[Callable, Callable]
+            SelectorProtocol | MultiSelectorProtocol, tuple[Callable, Callable, list]
```
kushalkolar: `list[Callable]` for the handlers

kushalkolar: code comment above when type signature is large

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
-            indices_local = map_(new)
-            selector_local.selection = indices_local
+    def selection(self, new: Integral | Sequence[Any]):
+        if self._block_reentrance:
+            return
+        else:
+            self._block_reentrance = True
+            if isinstance(new, Integral):
+                new = [new]
+            self._selection = [i for i in new]
```
kushalkolar: this doesn't do anything, can just be `self._selection = new`

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
+            # iterate through each selector that operates in its own "local" space
+            for selector_local, (map_, map_inv, handler) in self._selectors.items():
+                cumulated_output = []
```
kushalkolar: call it `local_indices`

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
+                for value in new:
+                    curr_indices = map_(value)
+                    cumulated_output.append(curr_indices)
+                # indices_local = map_(new)
```
kushalkolar: get rid of stale code

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
+            self._block_reentrance = False
 
     def append(self, index):
         self._selection.append(index)
-        for selector, (map_, map_inv) in self._selectors.items():
+        for selector, (map_, map_inv, handler_list) in self._selectors.items():
             if not isinstance(selector, MultiSelectorProtocol):
                 continue
 
             index_local = map_([index])
```
kushalkolar: this should be a scalar now and not a vector

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,127 @@
...
         self,
         new: (
             SelectorProtocol
-            | tuple[SelectorProtocol, Callable]
-            | tuple[SelectorProtocol, Callable, Callable]
+            | tuple[SelectorProtocol, np.ndarray | dict[int, int]]
         ),
     ):
+        """
+        User specifies (1) the selector and (2) The master --> local index mapping. This
```
kushalkolar, line 100: they can also specify a mapping Callable directly

kushalkolar (changes requested):

fastplotlib/graphics/selectors/_selection_vector.py
apasarkar: deleting for now because SelectionVector just works with VisibilitySelector + HighlightSelector (+ subclasses)

fastplotlib/graphics/selectors/_selection_vector.py
apasarkar: @kushalkolar thought about this -- this does something. if the input is any iterable other than a list, it makes the input a list.

fastplotlib/graphics/selectors/_selection_vector.py
kushalkolar: standardizing event typing throughout the entire rendercanvas-pygfx-fastplotlib stack is something we've proposed in the grant so yea let's worry about it later.

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,26 +1,54 @@
...
-from typing import Any, Sequence
+from typing import Any, Sequence, TypeAlias
 from numbers import Integral
+
 import numpy as np
+
 from ._protocols import SelectorProtocol, MultiSelectorProtocol
-from fastplotlib.graphics.features._base import GraphicFeatureEvent
+
+Mapping: TypeAlias = np.ndarray | dict[int, int] | Callable
```
kushalkolar:
```suggestion
Mapping = np.ndarray | dict[int, int] | Callable
```

kushalkolar: in the rest of the codebase we just assign type aliases directly

fastplotlib/graphics/selectors/_selection_vector.py
kushalkolar: Right it needs to be a `list` so `.append` works. But just use `list(new)` instead of a comprehension, easier to understand whta's going on.

kushalkolar (changes requested):

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -99,6 +99,7 @@ def add_selector(
             - A 1D np.ndarray of integers. The array index is the global index, and the array value is the local index
             - A dictionary where keys (master indices) and values (local indices) are both integers
             - Two callables. The first callable defines the global index --> local index map, the second specifies the local index --> global index map.
+                All callables take as input nonnegative integers and output nonnegative integers.
```
kushalkolar, line 105: should probably have a check to only allow positive indices?

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -16,15 +16,15 @@ def array_map(arr: np.ndarray, index: Integral):
...
+    return None if np.isnan(arr[index]) else int(arr[index])
 
 def inv_array_map(arr: np.ndarray,
                 value: int) -> None | Integral:
     """
     arr[i] gives the global index
     """
     x = np.flatnonzero(arr == value)
-    return None if x.size == 0 else x[0]
+    return None if x.size == 0 else int(x[0])
```
kushalkolar, line 27: How would a non-integer (float) end up here? I would think we should expect the user to pass in integer arrays, unless I'm missing something where we have a numpy operation somewhere here which just gives out floats?

kushalkolar (changes requested):

fastplotlib/graphics/selectors/_selection_vector.py
```diff
@@ -1,89 +1,161 @@
...
-            indices_local = map_(new)
-            selector_local.selection = indices_local
+    def selection(self, new: Integral | Sequence[Any]):
+        if self._block_reentrance:
+            return
+        else:
+            self._block_reentrance = True
+            if isinstance(new, Integral):
+                new = [new]
+            self._selection = [i for i in new]
```
kushalkolar:
```suggestion
            self._selection = list(new)
```

fastplotlib/graphics/selectors/_selection_vector.py
apasarkar: @kushalkolar for a callable, the mapping is given implicitly, so this is not possible. 

We could wrap each input callable in another callable that adds this check, but in my view that's not desirable. If the user wants to do a callable, they should follow what we state in the docs I'd say.

fastplotlib/graphics/selectors/_selection_vector.py
apasarkar: @kushalkolar the intention here was to actually make sure the return type was guaranteed to be "int" (not a numpy scalar or anything else). I've updated the type signature in the latest commit.

fastplotlib/graphics/selectors/_selection_vector.py
kushalkolar: When the index value is retrieved, I can see situations where it's hard to debug because negative indices are coming in somewhere and numpy arrays are indexed with a negative

apasarkar: @kushalkolar i would say that if the user provides a callable this is on them to debug properly. there's no way to exhaustively scan the input space of the callable to confirm that both mappings do not generate negative values. 

the only other possibility I can think of is we wrap each callable in a new callable that explicitly checks for negative values (composition). in my opinion this is not good design since it's less transparent to the user. 

what do you think?

kushalkolar: We don't check the callable but we check the value before using it.

apasarkar: @kushalkolar see latest!

kushalkolar (approved):

PR #1067, opened 2026-07-30
Includes basic np dlpack function that is tested to work across jax a…

### LINKED ISSUE #1063: use torch dlpack instead of cupy to convert cuda array to numpy

apasarkar:
Fixes #1063

The numpy dlpack fully sidesteps the need to rely on torch, but it requires numpy > 2.1. I think this is ok, since numpy is now at 2.5, and only users who care about high-performance (i.e. computing on GPUs) will use this execution path anyways.

### DIFF
--- a/fastplotlib/utils/functions.py
+++ b/fastplotlib/utils/functions.py
@@ -409,7 +409,7 @@ def parse_cmap_values(
 
 def cuda_to_numpy(arr: CudaArrayProtocol) -> np.ndarray:
 
-    data = np.from_dlpack(arr, device='cpu') #This requires numpy >= 2.1
+    data = np.from_dlpack(arr, device='cpu')
     return data
 
 
--- a/fastplotlib/utils/functions.py
+++ b/fastplotlib/utils/functions.py
@@ -408,14 +408,10 @@ def parse_cmap_values(
 
 
 def cuda_to_numpy(arr: CudaArrayProtocol) -> np.ndarray:
-    try:
-        import cupy
-    except ImportError:
-        raise ImportError(
-            "`cupy` is required to work with GPU arrays\npip install cupy"
-        )
 
-    return cupy.asnumpy(arr)
+    data = np.from_dlpack(arr, device='cpu') #This requires numpy >= 2.1
+    return data
+
 
 
 def subsample_array(
--- a/fastplotlib/utils/functions.py
+++ b/fastplotlib/utils/functions.py
@@ -413,7 +413,6 @@ def cuda_to_numpy(arr: CudaArrayProtocol) -> np.ndarray:
     return data
 
 
-
 def subsample_array(
     arr: CudaArrayProtocol,
     max_size: int = 1e6,
--- a/fastplotlib/widgets/nd_widget/_base.py
+++ b/fastplotlib/widgets/nd_widget/_base.py
@@ -198,18 +198,12 @@ class NDProcessor:
 
         self._spatial_dims = tuple(sdims)
 
-        ## This is the ordered sequence of data indices that will be displayed
-        self._spatial_dims_indices = tuple(
-            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
-        )
-
     @property
     def spatial_dims_indices(self) -> tuple[int, ...]:
         """
-        The ordered sequence of data indices that will be displayed
+        The ordered spatial dim indices that correspond to the named spatial dims
         """
-        return self._spatial_dims_indices
-
+        return tuple(self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims)
 
     @property
     def tooltip(self) -> bool:
--- a/fastplotlib/widgets/nd_widget/_base.py
+++ b/fastplotlib/widgets/nd_widget/_base.py
@@ -198,6 +198,19 @@ class NDProcessor:
 
         self._spatial_dims = tuple(sdims)
 
+        ## This is the ordered sequence of data indices that will be displayed
+        self._spatial_dims_indices = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
+
+
     @property
     def tooltip(self) -> bool:
         """
--- a/fastplotlib/widgets/nd_widget/_base.py
+++ b/fastplotlib/widgets/nd_widget/_base.py
@@ -529,12 +529,7 @@ class NDProcessor:
                 f"windowed_slice.ndim != len(self.spatial_dims): {windowed_slice.ndim} != {len(self.spatial_dims)}"
             )
 
-        # transpose to spatial dims
-        spatial_dims_int = tuple(
-            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
-        )
-
-        return windowed_slice.transpose(*spatial_dims_int)
+        return windowed_slice
 
     async def _get_raw_data_slice(self, indices: dict[str, Any]) -> ArrayProtocol:
         """
--- a/fastplotlib/widgets/nd_widget/_nd_image.py
+++ b/fastplotlib/widgets/nd_widget/_nd_image.py
@@ -137,6 +137,11 @@ class NDImageProcessor(NDProcessor):
         self._compute_histogram = compute_histogram
         self._recompute_histogram()
 
+        # Axis order of the spatial dimensions to display
+        self._spatial_dims_int = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
     @property
     def data(self) -> ArrayProtocol | None:
         """
@@ -258,7 +263,8 @@ class NDImageProcessor(NDProcessor):
         if isinstance(window_output, CudaArrayProtocol):
             window_output = await run_in_thread_pool(self._executor, cuda_to_numpy, window_output)
 
-        return window_output
+
+        return window_output.transpose(*self._spatial_dims_int)
 
     def _recompute_histogram(self):
         """
--- a/fastplotlib/widgets/nd_widget/_nd_image.py
+++ b/fastplotlib/widgets/nd_widget/_nd_image.py
@@ -186,18 +186,6 @@ class NDImageProcessor(NDProcessor):
 
         self._spatial_dims = tuple(sdims)
 
-        ## This is the ordered sequence of data indices that will be displayed
-        self._spatial_dims_indices = tuple(
-            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
-        )
-
-    @property
-    def spatial_dims_indices(self) -> tuple[int, ...]:
-        """
-        The ordered sequence of data indices that will be displayed
-        """
-        return self._spatial_dims_indices
-
     @property
     def rgb_dim(self) -> str | None:
         """
@@ -269,7 +257,6 @@ class NDImageProcessor(NDProcessor):
         # final CUDA -> numpy conversion at the end of the pipeline
         if isinstance(window_output, CudaArrayProtocol):
             window_output = await run_in_thread_pool(self._executor, cuda_to_numpy, window_output)
-
 
         return window_output.transpose(*self.spatial_dims_indices)
 
--- a/fastplotlib/widgets/nd_widget/_nd_image.py
+++ b/fastplotlib/widgets/nd_widget/_nd_image.py
@@ -137,11 +137,6 @@ class NDImageProcessor(NDProcessor):
         self._compute_histogram = compute_histogram
         self._recompute_histogram()
 
-        # Axis order of the spatial dimensions to display
-        self._spatial_dims_int = tuple(
-            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
-        )
-
     @property
     def data(self) -> ArrayProtocol | None:
         """
@@ -190,6 +185,18 @@ class NDImageProcessor(NDProcessor):
             )
 
         self._spatial_dims = tuple(sdims)
+
+        ## This is the ordered sequence of data indices that will be displayed
+        self._spatial_dims_indices = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
 
     @property
     def rgb_dim(self) -> str | None:
@@ -264,7 +271,7 @@ class NDImageProcessor(NDProcessor):
             window_output = await run_in_thread_pool(self._executor, cuda_to_numpy, window_output)
 
 
-        return window_output.transpose(*self._spatial_dims_int)
+        return window_output.transpose(*self.spatial_dims_indices)
 
     def _recompute_histogram(self):
         """
--- a/fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
+++ b/fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
@@ -112,6 +112,11 @@ class NDPositionsProcessor(NDProcessor):
         self.cmap_transform_each = cmap_transform_each
         self.sizes = sizes
 
+        # Axis order of the spatial dimensions to display
+        self._spatial_dims_int = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
     def _check_shape_feature(
         self, prop: str, check_shape: tuple[int, int]
     ) -> tuple[int, int]:
@@ -554,6 +559,8 @@ class NDPositionsProcessor(NDProcessor):
         # final CUDA -> numpy conversion at the end of the pipeline
         if isinstance(data, CudaArrayProtocol):
             data = await run_in_thread_pool(self._executor, cuda_to_numpy, data)
+
+        data = data.transpose(*self._spatial_dims_int)
 
         return {
             "data": data,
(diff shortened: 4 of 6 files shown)


apasarkar:
@kushalkolar so one issue I've been seeing is that the cuda array interface execution path seems to produce spatially transposed data. See below for a reproducible example.

```

## Make a reproducible example:
data = np.random.rand(100, 50, 200)
data_torch = torch.as_tensor(data)

ref_ranges = {'time': (0, data.shape[0], 1)}
subplot_names = ['data_numpy', 'data_torch']
curr_ndw = fpl.NDWidget(ref_ranges,
                        names=subplot_names,
                        shape=(1, 2),
                        size=(800, 800))

dims = ('time', 'height', 'width')
spatial_dims = ('height', 'width')

ndw_vid_numpy = curr_ndw['data_numpy'].add_nd_image(data,
                                       dims,
                                       spatial_dims,
                                       name = 'data_numpy')

ndw_vid_torch = curr_ndw['data_torch'].add_nd_image(data_torch,
                                       dims,
                                       spatial_dims,
                                       name = 'data_torch')



curr_ndw.show()
```

The reason for this is because at the following line, the transpose does different things if the input is a tensor vs. if it is a numpy array. https://github.com/fastplotlib/fastplotlib/blob/91faf5b16244c7e44bd1d07828744a71716d3458/fastplotlib/widgets/nd_widget/_base.py#L537

For a tensor, my_tensor.transpose(0, 1) will swap the axes 0 and 1. Otoh, my_numpy_array.transpose(0, 1) will keep the desired shape. 

Is there any reason we are doing this final transpose step? For an individual ndprocessor, it does not seem necessary. I think there are two options: 
(1) If there's no reason, drop it. 
(2) If we need to do it, defer this to after we convert the dlpack-compatible array to numpy. This way the function call will be consistent.

fastplotlib/utils/functions.py
```diff
@@ -408,14 +408,10 @@ def parse_cmap_values(
...
-    except ImportError:
-        raise ImportError(
-            "`cupy` is required to work with GPU arrays\npip install cupy"
-        )
-
-    return cupy.asnumpy(arr)
+
+    data = np.from_dlpack(arr, device='cpu') #This requires numpy >= 2.1
+    return data
+
```
kushalkolar, line 414: I just tried this and it works, nice!

fastplotlib/utils/functions.py
```diff
@@ -408,14 +408,10 @@ def parse_cmap_values(
...
-    try:
-        import cupy
-    except ImportError:
-        raise ImportError(
-            "`cupy` is required to work with GPU arrays\npip install cupy"
-        )
-
-    return cupy.asnumpy(arr)
+
+    data = np.from_dlpack(arr, device='cpu') #This requires numpy >= 2.1
```
kushalkolar:
```suggestion
    data = np.from_dlpack(arr, device='cpu')
```

kushalkolar (commented):
Where does it say that the `from_dlpack` requires numpy >= 2.1? Anyways that's fine, v2.1 is 2 years old at this point. pygfx requires min 2.1 anyways too https://github.com/pygfx/pygfx/blob/main/pyproject.toml#L28C1-L28C55

kushalkolar:
> For a tensor, my_tensor.transpose(0, 1) will swap the axes 0 and 1. Otoh, my_numpy_array.transpose(0, 1) will keep the desired shape.
> 
> Is there any reason we are doing this final transpose step? For an individual ndprocessor, it does not seem necessary. I think there are two options: (1) If there's no reason, drop it. (2) If we need to do it, defer this to after we convert the dlpack-compatible array to numpy. This way the function call will be consistent.

Annoying that torch behavior is different from numpy. 

The transpose is necessary by definition, that's how the NDWidget spec is. The specified spatial dims order defines the dim order for the graphic data.

Doing the transpose after it's converted to numpy sounds like the way to go.

apasarkar:
@kushalkolar Updated this -- so I've delayed the permutation operation until the sliced data is guaranteed to be a numpy array. nd vectors processor, nd image processor, and nd positions processor all subclass ndprocessor, so those are the files where i made this change. 

One thing to note: with this setup, the spatial functions must be written so they are applied data that has not yet been permuted. for example, if i specify data like this: 

dims = ['time', 'a', 'b']
spatial_dims = ['b', 'a'] # This means we end up transposing the sliced data

The spatial window function received the spatial dimensions in order (a, b) with these changes. 

I actually think this is ok -- this forces the user to write all their window/spatial function logic using the dimension order specified by dims, but the final order is transposed based on what's in spatial_dims. So we would ofc need to update the specs. I think so long as the user can write their whole pipeline (raw data --> displayed graphic) with a fixed dimension ordering it will be clear. 

Let me know what you think!

kushalkolar:
> One thing to note: with this setup, the spatial functions must be written so they are applied data that has not yet been permuted. for example, if i specify data like this:
> 
> dims = ['time', 'a', 'b'] spatial_dims = ['b', 'a'] # This means we end up transposing the sliced data
> 
> The spatial window function received the spatial dimensions in order (a, b) with these changes.
> 
> I actually think this is ok -- this forces the user to write all their window/spatial function logic using the dimension order specified by dims, but the final order is transposed based on what's in spatial_dims. So we would ofc need to update the specs. I think so long as the user can write their whole pipeline (raw data --> displayed graphic) with a fixed dimension ordering it will be clear.
> 
> Let me know what you think!

I think that makes sense. The whole point is that window_funcs and spatial_funcs should be fast and async, and things are naturally async when they remain in torch as long as possible. The `spatial_dims` arg for an `NDGraphic` is really just supposed to be the _display order_ of the dimensions for rendering the graphic, the `[x, y, z]` dims for positional data, `[row, col]` for image data nad `[depth, row, col]` for volumetric image data. The user needs to know what their dims are to create their window_funcs -> spatial_funcs pipeline.

Maybe we should rename the arg to `display_order` or `display_dims` or `render_dims` or something?

apasarkar:
ordered_spatial_dims might be a good one, specifying the two roles of this parameter: (1) it specifies what the spatial dims are and (2) it gives an order to output the spatial dims. thoughts?

apasarkar:
@kushalkolar I'll also flag the following potential issue: 

https://github.com/fastplotlib/fastplotlib/blob/91faf5b16244c7e44bd1d07828744a71716d3458/fastplotlib/widgets/nd_widget/_base.py#L525

At this line, a function "squeeze" is used. On a first read, I expected it to fail with torch tensors, since the documented parameters are tensor.squeeze(dim = ...), whereas the code is doing .squeeze(axis =...). Turns out at some point in pytorch the developers allowed axis to make it interoperable with numpy. 

Anyways it's not clear with other libraries that are compatible with dlpack whether this functionality will break.

kushalkolar:
`cupy` is supposed to be just like numpy and accepts `axis`: https://docs.cupy.dev/en/stable/reference/generated/cupy.squeeze.html
`jax` likewise: https://docs.jax.dev/en/latest/_autosummary/jax.numpy.squeeze.html

where is it documented that torch also accepts `axis`? I don't see it here: https://docs.pytorch.org/docs/2.13/generated/torch.squeeze.html#torch.squeeze

apasarkar:
@kushalkolar  Example:

https://discuss.pytorch.org/t/confuse-about-how-torch-argmax-supporting-axis-parameter-inplace-of-dim/182592

Looks like under the hood there is some code that parses keywords.

kushalkolar:
> @kushalkolar Example:
> 
> https://discuss.pytorch.org/t/confuse-about-how-torch-argmax-supporting-axis-parameter-inplace-of-dim/182592
> 
> Looks like under the hood there is some code that parses keywords.

wow that's hidden deep

apasarkar:
@kushalkolar I'm thinking from the standpoint of documentation. We ideally want to be able to tell someone "your dlpack compliant array can work here". Maybe there is not a workaround with the axis squeeze thing though

kushalkolar:
If it works with torch, cupy and Jax that covers most use cases I'd think?

apasarkar:
@kushalkolar  Fair enough, we can just make sure to note this in the documentation. 

Ok so what's left here - just the rename of the spatial_dims param to something else? If you're ok with renaming is ordered_spatial_dims, we'll have to update many of the function signatures in ndwidget, so let's make sure we're both happy with the naming here and I can go ahead and do it.

kushalkolar:
I added it to the list of renames to do later in bulk on #971

kushalkolar:
what's left is the suggestion and updating pyproject.toml and merge latest `ndwidget` into this

fastplotlib/utils/functions.py
apasarkar: Done!

apasarkar:
@kushalkolar Added the suggested change. 

Re: pyproject.toml, we are now using np.from_dlpack exclusively to bring gpu arrays to cpu and from there do zero-cost conversion to numpy. So it looks like we don't need to change the pyproject.toml at all.

kushalkolar:
> @kushalkolar Added the suggested change. 
> 
> Re: pyproject.toml, we are now using np.from_dlpack exclusively to bring gpu arrays to cpu and from there do zero-cost conversion to numpy. So it looks like we don't need to change the pyproject.toml at all. 

For the min version pin?

apasarkar:
@kushalkolar sorry I thought I did that earlier haha. 
I think g2g in latest commit.

fastplotlib/widgets/nd_widget/_nd_image.py
```diff
@@ -137,6 +137,11 @@ def __init__(
         self._compute_histogram = compute_histogram
         self._recompute_histogram()
 
+        # Axis order of the spatial dimensions to display
+        self._spatial_dims_int = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
```
kushalkolar: spatial_dims is a mutable property, so this should be a read only property or set in the spatial_dims setter as a private attribute (probably preferable, it doesn't have to be public) and not created once in the constructor

kushalkolar (changes requested):

fastplotlib/widgets/nd_widget/_nd_image.py
apasarkar: @kushalkolar good point, updated in latest commit. I also renamed spatial_dims_int to spatial_dims_indices for clarity

fastplotlib/widgets/nd_widget/_base.py
```diff
@@ -198,6 +198,19 @@ def spatial_dims(self, sdims: Sequence[str]):
...
+        self._spatial_dims_indices = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
```
kushalkolar:
```suggestion
        return tuple(
            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
        )
```

fastplotlib/widgets/nd_widget/_base.py
```diff
@@ -198,6 +198,19 @@ def spatial_dims(self, sdims: Sequence[str]):
...
 
+        ## This is the ordered sequence of data indices that will be displayed
+        self._spatial_dims_indices = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
```
kushalkolar:
```suggestion
        ordered spatial dim indices that correspond to the named spatial dims
```

fastplotlib/widgets/nd_widget/_base.py
```diff
@@ -198,6 +198,19 @@ def spatial_dims(self, sdims: Sequence[str]):
 
         self._spatial_dims = tuple(sdims)
 
+        ## This is the ordered sequence of data indices that will be displayed
+        self._spatial_dims_indices = tuple(
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -112,6 +112,7 @@ def __init__(
         self.cmap_transform_each = cmap_transform_each
         self.sizes = sizes
 
+
```
kushalkolar, line 115:
```suggestion
```

fastplotlib/utils/functions.py
```diff
@@ -408,14 +408,10 @@ def parse_cmap_values(
...
 def cuda_to_numpy(arr: CudaArrayProtocol) -> np.ndarray:
-    try:
-        import cupy
-    except ImportError:
-        raise ImportError(
-            "`cupy` is required to work with GPU arrays\npip install cupy"
-        )
-
-    return cupy.asnumpy(arr)
+
```
kushalkolar, line 411:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -298,6 +299,18 @@ def spatial_dims(self, sdims: tuple[str, str, str]):
...
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_positions/_nd_positions.py
```diff
@@ -555,6 +568,8 @@ async def get(self, indices: dict[str, Any]) -> dict[str, ArrayProtocol]:
         if isinstance(data, CudaArrayProtocol):
             data = await run_in_thread_pool(self._executor, cuda_to_numpy, data)
 
+        data = data.transpose(*self._spatial_dims_int)
```
kushalkolar:
```suggestion
        data = data.transpose(*self.spatial_dims_indices)
```

you made it a public property but didn't use it?

fastplotlib/widgets/nd_widget/_base.py
```diff
@@ -198,6 +198,19 @@ def spatial_dims(self, sdims: Sequence[str]):
...
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
+
+
```
kushalkolar, line 207:
```suggestion
```

fastplotlib/widgets/nd_widget/_base.py
kushalkolar, line 531: I think you can use the public property and make sure the rest of these code blocks are valid now?

fastplotlib/widgets/nd_widget/_nd_image.py
```diff
@@ -186,6 +186,18 @@ def spatial_dims(self, sdims: tuple[str, str] | tuple[str, str, str]):
...
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_image.py
```diff
@@ -258,7 +270,8 @@ async def get(self, indices: dict[str, Any]) -> ArrayProtocol:
         if isinstance(window_output, CudaArrayProtocol):
             window_output = await run_in_thread_pool(self._executor, cuda_to_numpy, window_output)
 
-        return window_output
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_vectors.py
```diff
@@ -143,6 +143,18 @@ def spatial_dims(self, sdims: tuple[str, str, str]):
...
+            self.spatial_dims.index(d) for d in self.dims if d in self.spatial_dims
+        )
+
+    @property
+    def spatial_dims_indices(self) -> tuple[int, ...]:
+        """
+        The ordered sequence of data indices that will be displayed
+        """
+        return self._spatial_dims_indices
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_vectors.py
```diff
@@ -175,7 +187,8 @@ async def get(self, indices: dict[str, Any]) -> ArrayProtocol:
         if isinstance(window_output, CudaArrayProtocol):
             window_output = await run_in_thread_pool(self._executor, cuda_to_numpy, window_output)
 
-        return window_output
+
```
kushalkolar:
```suggestion
```

fastplotlib/widgets/nd_widget/_nd_vectors.py
```diff
@@ -175,7 +187,8 @@ async def get(self, indices: dict[str, Any]) -> ArrayProtocol:
...
+        return window_output.transpose(*self._spatial_dims_indices)
```
kushalkolar:
```suggestion
        return window_output.transpose(*self.spatial_dims_indices)
```

it's a public property use it

kushalkolar (changes requested):

apasarkar:
@kushalkolar thanks -- all makes sense. made the changes, spatial_dims_indices is computed and set in the public property of the base class now. i was hesitant to do it this way at the start because the computation to define spatial dims indices will now be run at every single getitem. but if the number of dimensions is small, the overhead should never be a problem.

kushalkolar:
It's probably a nanosecond scale operation.

fastplotlib/widgets/nd_widget/_base.py
kushalkolar: Oh wait this is different nevermind (but similar name lol), this is to remove all non-spatial dims from the window slice.

apasarkar: yep!

PR #1072, opened 2026-08-08
Flat graphic features 2nd edition

### LINKED ISSUE #629: Flat architecture for all graphic features

### LINKED ISSUE #1004: Simplify colormap handling
As Almar suggested [here](https://github.com/fastplotlib/fastplotlib/pull/953#issuecomment-3606137250) we can just migrate to `pygfx.cm.create_colormap()`. This will make things easier and let us use custom colormaps with whatever number of custom colors. This is necessary for @KceniaB usecase with the probability heatmaps.

Can create a `TextureMap` like this with using the cmap library to generate the cmap array as well:

```python
texture_map = gfx.cm.create_colormap(cmap.Colormap(["r", "cyan", "magenta"]).lut(3), n=3)
```

Can then just set this on `material.map` for images, and `texcoords` for geometries (lines, scatters, meshes).

### LINKED ISSUE #1071: refactor `cmap` and `cmap_transform`
We can use pygfx `Geometry.texcoords` and `Material.map` to handle cmaps and transforms more elegantly:

Function to create [0, 1] normalized values for the transform that we feed to pygfx. This takes into account vmin, vmax, and gamma.

```python
def normalize_min_max(a, vmin: float = None, vmax: float = None, gamma: float = 1.0):
    """normalize an array between 0 - 1, clipped to (vmin, vmax)"""
    vmin = np.min(a) if vmin is None else vmin
    vmax = np.max(a) if vmax is None else vmax

    if vmax <= vmin:
        return np.zeros(a.size)

    transform = np.clip((a - vmin) / (vmax - vmin), 0, 1)
    if gamma == 1.0:
        return transform
    return transform ** gamma
```

This is basically everything we need to implement in the refactored graphic features:

```python
xs = np.linspace(0, 4 * np.pi, 100)
ys = np.sin(xs)
data = np.column_stack([xs, ys])

ys_scaled = normalize_min_max(ys)
geometry.texcoords = pygfx.Buffer(ys_scaled.astype(np.float32))

material.color_mode = "vertex_map"
material.map = cmap_lib.Colormap("seismic").to_pygfx()
```

We then have `vmin`, `vmax`, `gamma` graphic features on positional graphics as well and the imgui colorbar can keep them in sync.

kushalkolar:
closes #629 closes #1004 closes #1071

# Summary

## Flat graphic features

All graphic features are now flat:

`cmap` just handles cmap, nothing else, `cmap.transform` is now an independent flat feature `cmap_transform`.

```python
# so instead of
line.cmap.transform = sine_data
# do this
line.cmap_transform = sine_data
```

A `Colormap` object from the `cmap` lib can also be passed and it's used directly, internally we actually just pass cmap handling to the cmap lib!

```python
line.cmap = cmap_lib.Colormap(["r", "green", "orange"])
```

`VertexCmapRange` handles the `cmap_transform` range, it's a new graphic feature. So that we don't have to normalize the `cmap_transform` in the range `[0, 1]`.

Made a `types` module with type helpers for colors and colormaps, contains type aliases like `ColorLike`, `MultiColorLike`, etc. so that we don't have to put all the array, tuple, list, etc. types everywhere.

## GraphicCollections

Full re-write so that they behave much more intuitively and array-like.

With the new API we can do things like:

```python
data[3]            # (n_datapoints, xyz) view of graphic 3
data[1:5]          # object array (4,), cells (n_datapoints, xyz) views
data[1:5, :, 1]    # object array (4,), cells (n_datapoints,) y-views
colors[..., 3]     # object array (n_graphics,), cells alpha-column views
thickness[1:5]     # (4,) float array

sizes[:2] = 5                          # fills
sizes[:2] = np.array([1, 2])           # (2,) into (2, n_datapoints) -> broadcast ERROR
sizes[:2] = np.array([1, 2])[:, None]  # (2,1) -> per-line constant: line 0 all = 1, line 1 all = 2
sizes[:2] = np.array([1, 2])[None]     # (1,2) -> ERROR, same as on a rectangular array
data[:3, :, 1] = np.cos(xs)            # one curve into each of 3 lines (needs matching n_datapoints)
data[:3] = np.zeros((3, n_datapoints, xyz))   # per-line, per-point (homogeneous)

colors[:3] = "r"               # parsed -> lines 0-2 all red
colors[:10, 30:50] = "r"       # those points red
colors[0] = per_point          # per_point is (line 0's n_datapoints, RGBA)
colors[:4, 30:50] = c20        # c20 is (20, RGBA); same per-point colors into each of lines 0-3
colors[:3] = ["r", "g", "b"]   # parses to (3, 4) -> the 3 hits the datapoint axis -> broadcast error
colors[:3] = parse(["r","g","b"])[:, None]   # (3, 1, 4) -> one color per line
colors[::2, 10:15, 2] = 0.7    # raw: blue channel of those points
colors[:, :, 3] = 0.5          # raw: alpha everywhere
colors[:3, :, :2] = 0          # raw: zero red+green on lines 0-2

cmap = "jet"                     # one colormap across the collection -> one color per line (by index)
cmap = ["jet", "viridis", ...]   # per-line colormaps (len n_graphics), each along its own datapoints
cmap[i:j] = "jet"                # only if cmap is already per-line; on one shared colormap this raises

cmap_transform = vals[:, None]   # (n_graphics, 1): varies along graphics -> one value per line
cmap_transform = vals[None]      # (1, n_datapoints): same gradient on every line
cmap_transform = full            # (n_graphics, n_datapoints): per-line, per-point
cmap_transform = o[:, None]      # jagged per-line gradients (bare o collides)
```

Basically, when we have `collection.feature`, the `feature` property behaves like a numpy array, the first dim is always the `n_graphics` dim. Broadcasting etc. is handled just like how numpy does it.

### Implementation

Each `GraphicCollection.property` that pertains to graphic features of the graphics in the collection is wrapped by an Accessor, which is also has some subclasses.

#### Accessor

Basic class that parses a user's feature value and broadcasts it across graphics, used for the simple features that don't have a per-vertex/per-datapoint dimension such as line thickness. A user can do `lines.thickness = 5` or `lines.thickness = [2, 4, 3, 2, 5, ...]` or `lines.thickness[2:5]` or any other numpy-like slicing and it handles the broadcasting accordingly.

#### JaggedArray

Accessor subclass which accounts for features that have a `n_datapoints` dim, such as `VertexData` or `VertexColors`. It allows collection-features to behave as if they were a rectangular array and accounts for the fact that `n_datapoints` might vary across graphics in the collection.
This lets us do things like:

```python
lines.data = <new array of shape [n_lines, n_datapoints, xyz]
# or even this
lines.data = <list of array which each vary in the number of datapoints, [n_datapoints_i, xyz]>
# all forms of numpy-like slicing and indexing are allowed
lines.data[<slice or fancy index>] = <list or array>
```

#### ColorArray

Just subclasses `JaggedArray` and parses color args to RGBA

#### Cmap

Manages cmaps across-graphics in a collection or per-graphic cmaps.

Comparison operators are allowed so stuff like this is possible:

### `GraphicCollection`

The `GraphicCollection` base class takes care of automatically creator the `Accessor`/`JaggedArray` feature instance properties for the collection subclasses by just parsing the child graphic type's features, so essential `GraphicCollection` subclasses are auto-generated! It has some parsing to make sure cmap stuff is done correctly.

`graphic_collection[<slice or fancy index>]` returns an instance of the same graphic collection type with the sub-selection of graphics, this instance behaves like any other graphic collection of the same type.

```python
lines.thickness < 3
# returns bool numpy array like this
np.ndarray([True, True, False, False, True])
```

### `ImageCollection` and `ImageGrid` are new

Examples:

```python
# a collection of random 100 x 100 images
images = [rng.random((100, 100), dtype=np.float32) for _ in range(5)]

figure = fpl.Figure(size=(700, 560))

# stagger the images diagonally with one (x, y, z) offset per image
offsets = np.array([[i * 60, -i * 60, 0] for i in range(len(images))])

figure[0, 0].add_image_collection(images, offsets=offsets, cmap="viridis")
```

```python
# a collection of random 100 x 100 images
images = [rng.random((100, 100), dtype=np.float32) for _ in range(6)]

figure = fpl.Figure(size=(700, 560))

# lay the images out in a 2 x 3 grid with a gap between rows and columns
figure[0, 0].add_image_grid(images, shape=(2, 3), separation=(10, 10), cmap="plasma")
```

All the usual operators will work with all the graphic collection features!

## `PositionsGraphic`, `LineGraphic`, `ScatterGraphic` cleanup

* Lot of general cleanup, `color_mode` is gone, and all other bools like `uniform_markers`, `uniform_sizes`, etc. are gone! Everything is auto-determined.
* Automatic `GraphicFeature` inheritance in `__init_subclass__`, before it was easy to forget making sure these are all accounted for.

## NDWidget

All the garbage properties like `cmap_each` are gone, user just relies on usual numpy broadcasting :D . I'm still working on finishing stuff on the NDWidget side for this so it's not in the diff yet.

All features on an `NDGraphic` should work analogous to a collection, the `NDGraphic` mostly just passes the argument to the collection. So `NDGraphic.colors`, `cmap`, `cmap_transform`, `markers`, etc. should all behave just as if they were setting those values on a collection! Main thing to note is that if the new value has a `n_datapoints` dim, e.g. `cmap_transform` that maps to all the datapoints, then they should pass in an array that actually contains values for every datapoint and not just a display window (obvious I think?) and `NDPositions` handles setting it on the `PositionsCollection` object w.r.t. the current window!

`NDTimeseries` sets the y-max "steps" argument for the `LineStack`/`ScatterStack` so that they don't get mangled.

## Other

`ImageCmap` feature now just uses `cmap_lib.Colormap`, gets rid of ALL the ugly cmap parsing code we had :smile: 

# LLM Disclosure
Claude was used for parts of this, especially for docstring, but I babysat it through everything. With that said, this was _really_ pushing the limits of what claude can do on max effort.


kushalkolar:
More WIP, need to:

* ~vertex cmap vmin, vmax, gamma features~
* finish implementation for line, implement for scatter, can probably move some common kwargs making dict funcs to `PostionsGraphic`

kushalkolar:
Before I forgot I should check if the `cmap` lib has a builtin way to apply transforms, vmin, vmax so I can just use that an apply it instead of computing it on the fpl side. 

EDIT: `LineMaterial.maprange` is probably what I should be using.

Small changes I want to make before I forget:

- `PositionsGraphic.cmap_transform` keeps the original values and uses `maprange`, this way the user can intuitively keep the original values like cluster labels instead of seeing the interpolated normalized version
- Implement corresponding behavior in collections for `cmap_transform` across graphics.
- use cmap lib for images too.

kushalkolar:
Need to fix docs gen:

```
  File "/home/kushal/repos/fastplotlib/docs/source/./generate_api.py", line 524, in <module>
    main()
    ~~~~^^
  File "/home/kushal/repos/fastplotlib/docs/source/./generate_api.py", line 520, in main
    f.write(write_table(name, type_))
            ~~~~~~~~~~~^^^^^^^^^^^^^
  File "/home/kushal/repos/fastplotlib/docs/source/./generate_api.py", line 497, in write_table
    s += dict_to_rst_table(feature_cls.event_info_spec)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'JaggedArray' has no attribute 'event_info_spec'
```

fastplotlib/graphics/features/_image.py
```diff
@@ -519,18 +517,23 @@ class ImageCmap(GraphicFeature):
...
+    def set_value(self, graphic, value: ColormapLike | cmap_lib.Colormap):
+        _map = cmap_lib.Colormap(value).to_pygfx()
+        _map.min_filter = graphic._cmap_interpolation.value
+        _map.mag_filter = graphic._cmap_interpolation.value
+        _map.mipmap_filter = graphic._cmap_interpolation.value
+        _map.wrap_s = "clamp-to-edge"
+        _map.wrap_t = "clamp-to-edge"
+
+        graphic._material.map = _map
         graphic._material.map.texture.update_range((0, 0, 0), size=(256, 1, 1))
```
clewis7, line 543: code comment here

fastplotlib/graphics/features/_positions.py
```diff
@@ -339,138 +347,131 @@ def __len__(self):
...
-    def transform(
-        self,
-        values: np.ndarray | list[float | int],
-        indices: slice | list | np.ndarray = None,
-    ):
-        if self._cmap_name is None:
-            raise AttributeError(
-                "cmap name is not set, set the cmap name before setting the transform"
-            )
+    def _interpolate(self, value, n_datapoints):
```
clewis7, line 417: code comment/doc string here even though it is private method just to say why we need it

fastplotlib/graphics/_jagged_array.py
clewis7, line 1: the naming convention is a bit confusing 

Accessor -> JaggedArray, Cmap, ColorArray is not very intuitive 

I think to keep it uniform: Cmap -> CmapArray and seems like Accessor should be called something else SimpleArray or SingleArray

fastplotlib/graphics/_collections.py
```diff
@@ -0,0 +1,445 @@
...
+
+        offsets: array-like, optional
+            explicit (x, y, z) offset per image; when given, ``shape`` and ``separation`` are ignored
+
+        **kwargs
+            passed to :class:`.ImageCollection`, e.g. ``cmap``, ``vmin``, ``vmax``
+        """
+        super().__init__(data, **kwargs)
+        n = len(self)
+
```
clewis7, line 312: maybe just note that the default behavior is to organize the images in a grid with no offset

fastplotlib/graphics/_collections.py
clewis7, line 1: this is really cool! very clean

fastplotlib/graphics/_collection_base.py
clewis7, line 1: the collections base still feels a little soupy

clewis7: but I assume that is the only way to get the clean abstractions for everything above the base class

clewis7 (approved):
mostly LGTM, just had a view high-level comments 

overall, I think the design is good and I like how everything behaves uniformly now

fastplotlib/graphics/_jagged_array.py
kushalkolar: renamed to:

`CollectionFeatureAccessor`
`JaggedCollectionFeature`
`CollectionColors`
`CollectionCmap`

kushalkolar:
OK just need to finish tests and adjust the examples and then will merge!

PR #1080, opened 2026-09-13
config system

kushalkolar:
An actually maintainable config system.

- [x] screenshot test (should be simples, I did some manual testing already)
- [x] populate the default preset by converting all methods to dicts and store that dict
- [x] implement on collections and use .pyi file so IDEs provide useful info 
- [x] docs page or section in user guide
- [x] implement for the `add_<graphic>` methods by using fancie descriptor logic
- [x] Some axes label improvement stuff which got committed to this branch

## User API

```python
fpl.LineGraphic.config.init.colors = "b"
fpl.layouts.Subplot.config.init.background_color = "w"

fpl.layouts.Subplot.config.init.toolbar = False
```
Some presets:

```python
fpl.style.light()
fpl.style.compact()
fpl.style.flynn()
```

Presets can be stacked, they "merge" into each other, so `fpl.presets.light()` followed by `fpl.presets.compact()` will apply both. This of course has to be done sensibly, `fpl.presets.light()` followed by `fpl.presets.dark()` just reverts things. Each preset just sets config values, that's all.

## Registering a class to the config system

```python
@global_config.register
class Figure:
    config = global_config.descriptor
    @global_config.set(size=(900, 700)
    def __init__(self):
        pass

# Subclass doesn't have to declare @config.set for 
# any methods that are only defined in the parent class
# so this will work and it'll use the same configured `size` as Figure
@global_config.register
def ImguiFigure(Figure):
    ...


# all graphic subclasses get access since only the 
# base class has to declare the config class variable
class Graphic:
    config = global_config.descriptor


# still have to register all subclasses into the configuration though
@global_config.register
class LineGraphic(Graphic):
    @global_config.declare("colors", "thickness")
    def __init__(
        self,
        data,
        colors: str | tuple[float, float, float] = "w",
        thickness: float = 2.0,
        cmap: str | None = None,
    ):
        print(data, colors, thickness, cmap)
        print(LineGraphic.config)
```

### Rules:
1. A class must be registered with `@global_config.register`
2. A class must have the class attribute with the name `config` set to `global_config.descriptor`
3. A registered class must have at least one method with `@global_config.set` on them, or its parent classes must have them.
4. All subclasses of a registered class must also be explicitly registered with `@global_config.register`, however if this subclass is overriding methods in the parent class which have `@global_config.set` I don't _think_ it has to be registered but we don't have a test case for this in fastplotlib AFAIK.
6. The default value in the method signature is used as the default config value. 
7. Any method that is not called with an explicit value, e.g. `add_line(data=...)` will inject the colors from the `global_config`, if we do `add_line(data=..., colors="r")` the explicitly provided `colors="r"` will be used and the config value will be ignored.

All rules above are necessary for a class to be in the config system.

Other notes:
- A subclass that is registered but that does not explicitly mark a method with `@global_config.set` uses its parent's config values. For example `ImguiFigure` is registered, but the `ImguiFigure` constructor has no `@global_config.set` so it'll use values from the parent, i.e. the config object is shared.

## How this works

This is some fancie python. Relies on the fact that the class decorator executes after the class has been created, and therefore after all the method decorators have also been executed. So we populate a "_pending" list with method defaults, and then the class `@config.register` decorator triggers that class to be registered with those defaults for its methods. 

The rest is really just some parsing, and the actual "config objects", i.e. `LineGraphic.config` is a dataclass so that you can tab complete, and then `LineGraphic.config.init` itself is also a dataclass with each argument-value pair for that method, so users can just do `LineGraphic.config.init.colors = "r"`. This is much better than matplotlib rcParams where everything is just a string you have to remember: https://matplotlib.org/stable/users/explain/customizing.html#runtime-rc-settings

Also relies on proper namespacing of the methods using the qualifying name and making sure everything matches up at all levels of this process, relies on the same system that `pickle` uses for knowing the identity of classes it serializes (i.e. `__qualname__`), straight from the horse's mouth:

https://github.com/python/cpython/blob/978fba58aef347de4a1376e525df2dacc7b2fff3/Lib/pickle.py#L1050-L1062

### Better Mixin with descriptors!

The script to generate the mixin class with `add_<graphic>()` methods is gone! :partying_face: . It's replaced with elegant descriptors :smile: 

https://github.com/fastplotlib/fastplotlib/blob/09dfc8d7d51ad97fe700c91f543aa9bf5cc326e9/fastplotlib/layouts/_graphic_methods_mixin.py

## Video

https://github.com/user-attachments/assets/4ea492b0-095a-4a52-b710-efcd9450582f

Didn't show in the video but obviously if a value is provided for an argument it'll use that and not the config value:

image

This PR is fully organic written by a human. Claude was producing completely useless garbage for this task.


fastplotlib/utils/_config.py
```diff
@@ -0,0 +1,346 @@
...
+                    # configurable value
+                    if val is ConfigValue:
+                        # fill in from current config
+                        binding.arguments[arg] = getattr(method_config, arg)
+
+                return method(*binding.args, **binding.kwargs)
+
+            return injector
+
+        return wrapper
```
kushalkolar: @clewis7 fanciest decorator I've written 😂

kushalkolar:
@clewis7 r4r

This looks large but the main contribution is `utils/_config.py`: https://github.com/fastplotlib/fastplotlib/pull/1080/changes#diff-708bbac7e717325519dfb235a4f11989f227ac3f14a29f734280f06d0a8871b7

The rest is is just adding the decorators and stuff to register each class relevant method.

fastplotlib/utils/_style.py
```diff
@@ -0,0 +1,108 @@
...
+                title_kwargs=dict(font_size=0),
+            ),
+        )
+
+    @staticmethod
+    def flynn():
+        """preset that optical physiology like"""
+        presets.very_compact()
+
+        layouts.Subplot.config.auto_scale.zoom = 0.99
```
kushalkolar, line 108: @FlynnOConnell you can modify and we can keep this in here since you have lots of users, unless you want to store them yourself.

fastplotlib/graphics/_base.py
```diff
@@ -59,6 +60,8 @@
 
 
 class Graphic:
+    config = global_config.descriptor
```
kushalkolar, line 63: @clewis7 you might be wondering we have this whole descriptor stuff and it's because we can't do:

```python
class A:
    config = global_config._registry[A]
```

Because `A` doesn't exist yet when the python interpreter is creating the `config` variable! By using the descriptor you don't need `A` to exist yet, it only needs to exist when the user calls `A.config` 😆 

The interpreter parses everything in the class and creates all declared objects in the class (methods as well) and the class is created only after everything in the class has been created! It's like filling a cup of water but the cup exists after all the water has been poured into it 😂

fastplotlib/utils/_config.py
```diff
@@ -0,0 +1,346 @@
...
+        if instance is not None:
+            raise AttributeError("set config options on the class, not an instance")
+
+        if cls not in self.__classes.keys():
+            raise AttributeError("Class is not registered")
+
+        return self.__classes[cls]
+
+    def __set__(self, obj, value):
+        raise AttributeError("Cannot set")
```
kushalkolar, line 41: @clewis7 See comment on base `Graphic` if you're wondering why this exists

kushalkolar:
I think I can just get rid of ConfigValue and replace using this  https://stackoverflow.com/questions/17625695/is-it-possible-to-change-a-functions-default-parameters-in-python

Then I can just specify which arguments are configurable on the method decorator, and it populates the defaults in the wrapper, and inject the config values into `__defaults__ `

FlynnOConnell:
I absolutely love this 

image

fastplotlib/utils/_config.py
```diff
@@ -0,0 +1,371 @@
...
+                # this isn't actually necessary but is just a robust failsafe
+                # I think it should account for any weirdness with methods that have positional-only arguments
+                binding.apply_defaults()
+
+                # finally call method with updated binding from config
+                return method(*binding.args, **binding.kwargs)
+
+            return injector
+
+        return append_to_config
```
kushalkolar, line 335: @clewis7 fanciest decorator I've written 😂 (updated version which is even more elegant)

kushalkolar:
Graphic methods generated mixin is gone and replaced with descriptors :partying_face: 

image

kushalkolar:
Merging, @clewis7 has agreed to fast-track this PR.

PR #1086, opened 2026-09-21
generate api docs during sphinx build

kushalkolar:
This runs the `generate_api.py` right before the sphinx docs build so we don't have to remember to do it manually.


kushalkolar:
@clewis7 gtg!

clewis7 (approved):

